PR #87, opened 2019-11-07
Added unit tests for tECDSA client

nkuba:
This PR adds unit test coverage for tECDSA protocol and client initialization.  To achieve that we had to extract persistence handle mock code to a package that can be reusable across our tests. As well as improve the local ethereum chain implementation. We also added some tests to cover changes in the local chain code.

Depends on: https://github.com/keep-network/keep-tecdsa/pull/85


pdyraga:
I think this one is super out of date and there is no chance we can integrate this code as-is. Thoughts on closing it without merging and opening another one in the future? This time with a faster reviewer...

nkuba:
I'm afraid that if closed it will be forgotten 😢

nkuba:
Let's forget about this one 😉

PR #283, opened 2020-03-04
Slash KEEP tokens on signature fraud

nkuba:
In this PR we added slashing of KEEP tokens from keep members in case of signature fraud.

Closes: https://github.com/keep-network/keep-tecdsa/issues/254

### DIFF
--- a/solidity/contracts/KeepBonding.sol
+++ b/solidity/contracts/KeepBonding.sol
@@ -3,6 +3,8 @@ pragma solidity ^0.5.4;
 import "@keep-network/keep-core/contracts/Registry.sol";
 import "@keep-network/keep-core/contracts/TokenStaking.sol";
 
+import "openzeppelin-solidity/contracts/math/SafeMath.sol";
+
 // TODO: This contract is expected to implement functions defined by IBonding
 // interface defined in @keep-network/sortition-pools. After merging the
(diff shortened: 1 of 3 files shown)


solidity/contracts/BondedECDSAKeepFactory.sol
```diff
@@ -405,16 +417,39 @@ contract BondedECDSAKeepFactory is IBondedECDSAKeepFactory, CloneFactory {
...
+    function slashKeepMembers() public onlyKeep {
+        BondedECDSAKeep keep = BondedECDSAKeep(msg.sender);
+
+        tokenStaking.slash(minimumStake, keep.getMembers());
+    }
+
+    /// @notice Notifies the factory that a keep has been closed.
+    /// @dev Function should be called by a keep which is being closed.
+    function notifyKeepClosed() public onlyKeep {
+        keeps[msg.sender] = false;
```
Shadowfiend, line 443: Late hit here, but is this equivalent from a gas perspective to `delete keeps[msg.sender]`? If not, would it make sense to delete rather than set to `false`?

pdyraga: It's the same. `delete` basically sets to a default value.

PR #313, opened 2020-03-17
ETH-only bonding

eth-r:
Add a bonding contract which doesn't rely on staking KEEP, but instead permits bonding with ETH only. To reduce sortition pool manipulation, `EthBonding` implements the initialization period and inability to increase the staked amount from `TokenStaking`.


pdyraga:
Closing in favor of https://github.com/keep-network/keep-ecdsa/pull/483

PR #319, opened 2020-03-20
Full Rename: Rename ALL OF THE THINGS (remaining*) to keep-ecdsa

Shadowfiend:
Just infra renames here!

Loose order of operations:

- Double check renamed files for lingering tecdsa references.   Make changes where needed.

- `git mv` files from tecdsa -> ecdsa filenames

- If no keep-ecdsa (InitContainer and application) image published, rename the last keep-tecdsa image.  This will prevent deployments from failing without a rebuild.  We want to do this to test deployments from this branch without having to wait on builds.

- Apply `*-service` configs

- Take IPs from new services and set them in the appropriate `peers` configs for each ecdsa client.  In the case of `keep-test`, we'll need to update DNS entries with new addresses.

- Rotate keep-ecdsa StatefulSets one at a time.

- Kill tecdsa services

#### keep-dev

- [X] Double check renamed files for lingering tecdsa references. Make changes where needed.
- [X] `git mv` files from tecdsa -> ecdsa filenames
- [X] keep-ecdsa image and InitContainer available (docker pull / rename / docker push)
- [X]  Renamed services (0-4) applied to env
- [X] StatefulSet configs updated with new service addresses (0-4)
- [X] StatefulSet deployment rotation (destroy/create one at a time)
- [X] Remove old tecdsa Services from environment

#### keep-test

- [X] Double check renamed files for lingering tecdsa references. Make changes where needed.
- [X] `git mv` files from tecdsa -> ecdsa filenames
- [X] keep-ecdsa image and InitContainer available
- [X] Rename executable in keep-ecdsa application image locally and push to container registry.
- [X] Rename provisioner script for InitContainer locally and push to container registry.
- [X] Rename template file and config file write in provisioner script
- [X]  Renamed services (0-4) applied to env
- [X] New Service addresses added to DNS entries for ecdsa clients
- [X] StatefulSet deployment rotation (destroy/create one at a time)
- [X] Remove old tecdsa Services from environment

--------

Missing:
- [x] Merging open-season: #316 .


sthompson22:
@nkuba Ping, This is deployed, would prefer it doesn't linger.

.circleci/config.yml
```diff
@@ -205,8 +205,8 @@ jobs:
       - run:
           name: Upload contract data
           command: |
-            cd /tmp/keep-tecdsa/contracts
-            gsutil -m cp * gs://${CONTRACT_DATA_BUCKET}/keep-tecdsa
+            cd /tmp/keep-ecdsa/contracts
+            gsutil -m cp * gs://${CONTRACT_DATA_BUCKET}/keep-ecdsa
```
nkuba: Do we need to update any other repos using this bucket? `tbtc` for instance?

Shadowfiend: Good catch: https://github.com/keep-network/tbtc/blob/b87758e91960b0f2446e5355ef422a8bfcc46910/implementation/scripts/circleci-provision-external-contracts.sh#L10

This should go away once packaging is operating properly…

Issue #297, opened 2020-03-06
Initialize implementations for proxy contracts and protect initialization methods


pdyraga:
It should be avoided that the implementation of proxy contracts can be initialized by third parties. This can be the case if the `initialize` function is unprotected. Since the implementation contract is not meant to be used directly without a proxy delegate-calling it is recommended to protect the initialization method of the implementation by initializing on deployment.

When working on this issue, please take care of `BondedECDSAKeepVendorImplV1` initialization but also make sure there are no other similar problems in other places of the code.

nkuba:
@tintinweb's suggestions are handled in https://github.com/keep-network/keep-ecdsa/pull/323.

PR #323, opened 2020-03-25
Two-step proxy initialization call moved to the second step

nkuba:
This PR is a followup covering https://github.com/keep-network/keep-ecdsa/issues/297#issuecomment-601595942.

With the previous implementation of the two-step proxy upgradeability, we allowed the admin to interact with the implementation contract at the first step of the upgrade process, which gave some exploits possibilities. In this PR we moved the call to an initialize function of the implementation (logic contract) to a second step so it can only be executed after the required time delay.

To be able to pass the initialization call data between the two steps we were required to store dynamic length bytes array in the contract storage. Due to the risk of clashing data storage between proxy and implementation we previously decided to use Unstructured Storage pattern where we define fixed positions for each of the variables used in the proxy contract. Unfortunately, it didn't work for dynamic bytes array, so we decided to store this dynamic type data in a mapping which is [distributing storage based on the data's hash](https://solidity.readthedocs.io/en/v0.5.15/miscellaneous.html#mappings-and-dynamic-arrays).

To clean up the code a little bit and prepare it for reusing we extracted storage variables for the two-step upgradeability pattern to a separate contract. There are more parts that can be extracted to separate contracts but it does not make much sense now taking into consideration the timeframes and that we are willing to reorganize our repositories in the near future.

### DIFF
--- a/solidity/contracts/BondedECDSAKeepVendor.sol
+++ b/solidity/contracts/BondedECDSAKeepVendor.sol
@@ -36,8 +36,13 @@ contract BondedECDSAKeepVendor is Proxy, UpgradableProxyStorage {
     /// subtracted by 1, and is validated in the constructor.
     bytes32 internal constant UPGRADE_INIT_TIMESTAMP_SLOT = 0x0816e8d9eeb2554df0d0b7edc58e2d957e6ce18adf92c138b50dd78a420bebaf;
 
-    event UpgradeStarted(address implementation, uint256 timestamp);
-    event UpgradeCompleted(address implementation);
+    event UpgradeStarted(
+        string version,
+        address implementation,
+        bytes initialization,
+        uint256 timestamp
+    );
+    event UpgradeCompleted(string version, address implementation);
 
     constructor(
         string memory _version,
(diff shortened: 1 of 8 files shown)


solidity/contracts/BondedECDSAKeepVendor.sol
```diff
@@ -1,68 +1,42 @@
...
-            UPGRADE_INIT_TIMESTAMP_SLOT,
-            "network.keep.bondedecdsavendor.proxy.upgradeInitiatedTimestamp"
-        );
+    event UpgradeStarted(
+        string version,
+        address implementation,
+        bytes initialization,
+        uint256 timestamp
+    );
+    event UpgradeCompleted(string version, address implementation);
```
pdyraga: Should we emit address of the implementation again here? I think emitting just the version should be enough. We emitted it for `UpgradeStarted` event.

solidity/contracts/UpgradableProxyStorage.sol
```diff
@@ -0,0 +1,206 @@
...
+
+    function getImplementation(uint256 _version)
+        internal
+        view
+        returns (Implementation memory)
+    {
+        return implementations[_version];
+    }
+
+    function setImplementation(
```
pdyraga: Thoughts on naming it `addImplementation`? We are storing all implementations in the map, we are not overwriting anything.

EDIT: I think we shouldn't store implementations in this mapping. Everyone can track the history of upgrades from the events we are emitting. Storing just an initialization data we are committing to should be enough. And we can even delete the data from this mapping once the initialization is done.

solidity/contracts/BondedECDSAKeepVendor.sol
```diff
@@ -140,123 +154,31 @@ contract BondedECDSAKeepVendor is Proxy {
...
-        bytes32 position = UPGRADE_INIT_TIMESTAMP_SLOT;
-        /* solium-disable-next-line */
-        assembly {
-            _upgradeInitiatedTimestamp := sload(position)
-        }
-    }
+    function _implementation() internal view returns (address) {
+        Implementation memory implementation = getImplementation(
+            currentImplementationID()
+        );
```
pdyraga: Do we know what's the impact on the gas cost? It affects all the calls to the system. In the previous implementation we were doing a single `sload`, without reading the entire map key and calling other functions.

I think the most effective and simplest solution is to store the current implementation data as before with just
```
bytes32 slot = IMPLEMENTATION_SLOT;
/* solium-disable-next-line */
assembly {
  impl := sload(slot)
}
```
here and have a `mapping` between the new version ID and initialization call data. When the upgrade starts, we commit to initialization and add the data to the mapping. Then, when the upgrade completes we execute the initialization using the data we previously committed to. If one wants to go through the history of upgrades they can use events.

nkuba:
We replaced this implementation with a much simpler solution in https://github.com/keep-network/keep-ecdsa/pull/327

PR #327, opened 2020-03-26
Proxy initialization call moved to the upgrade's second step - simple approach

nkuba:
This PR is a followup covering https://github.com/keep-network/keep-ecdsa/issues/297#issuecomment-601595942.

With the previous implementation of the two-step proxy upgradeability, we allowed the admin to interact with the implementation contract at the first step of the upgrade process, which gave some exploits possibilities. In this PR we moved the call to an initialize function of the implementation (logic contract) to a second step so it can only be executed after the required time delay.

To be able to pass the initialization call data between the two steps we were required to store dynamic length bytes array in the contract storage. Due to the risk of clashing data storage between proxy and implementation we previously decided to use Unstructured Storage pattern where we define fixed positions for each of the variables used in the proxy contract. Unfortunately, it didn't work for dynamic bytes array, so we decided to store this dynamic type data in a mapping which is [distributing storage based on the data's hash](https://solidity.readthedocs.io/en/v0.5.15/miscellaneous.html#mappings-and-dynamic-arrays).

### DIFF
--- a/solidity/contracts/BondedECDSAKeepVendor.sol
+++ b/solidity/contracts/BondedECDSAKeepVendor.sol
@@ -33,6 +33,13 @@ contract BondedECDSAKeepVendor is Proxy {
     /// subtracted by 1, and is validated in the constructor.
     bytes32 internal constant UPGRADE_INIT_TIMESTAMP_SLOT = 0x0816e8d9eeb2554df0d0b7edc58e2d957e6ce18adf92c138b50dd78a420bebaf;
 
+    /// @notice Details of initialization data to be called on the second step
+    /// of upgrade.
+    /// @dev Mapping is stored at position calculated with keccak256 of the entry
+    /// details, hence it should be protected from clashing with implementation's
+    /// fields.
+    mapping(address => bytes) public initializationData;
+
     event UpgradeStarted(address implementation, uint256 timestamp);
     event UpgradeCompleted(address implementation);
 
@@ -88,9 +95,7 @@ contract BondedECDSAKeepVendor is Proxy {
             "Implementation address must be different from the current one."
         );
 
-        if (_data.length > 0) {
-            initializeImplementation(_newImplementation, _data);
-        }
+        initializationData[_newImplementation] = _data;
 
         setNewImplementation(_newImplementation);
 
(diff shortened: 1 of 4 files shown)


solidity/contracts/BondedECDSAKeepVendor.sol
```diff
@@ -106,7 +111,7 @@ contract BondedECDSAKeepVendor is Proxy {
     /// process. The function emits an event containing the new implementation
     /// address. It can be called after upgrade time delay period has passed since
     /// upgrade initiation.
-    function completeUpgrade() public {
+    function completeUpgrade() public onlyAdmin {
```
pdyraga: Why `onlyAdmin`? I think the previous version was better. We made a promise and we are publicly bound to that promise.

nkuba: I liked the previous one too, but right now when we are calling the initialize function we need to restrict the caller to be the admin in case the implementation wants to get `msg.sender` and set it as the contract owner.

pdyraga: I don't love it but indeed, there is a risk. Interested in @tintinweb thoughts here.

pdyraga: One option here could be: we let anyone cancel the upgrade if the admin did not complete it after some certain timeout. But this timeout should be quite long to avoid censorship. Still digesting it... 🤔

nkuba: Or explicitly state that implementation's initialization function cannot use `msg.sender` but if needed it should expect the address to be provided as a function parameter, so it could be defined by admin in the initialize data.

pdyraga: I was thinking about it but rejected this idea as the only thing that guards us against terrible failure is a comment. Not compiler, not unit test, just a comment.

I think the best option is to have `onlyAdmin` on this function and just change the perspective and make it clear in docs. The whole point is not to commit that _the upgrade will be complete in one day_ but that _we will not upgrade the system without a one day delay and announcing it first_.

PR #328, opened 2020-03-26
Solidity and JS linters

### LINKED ISSUE #42: Solidity and JS linters
Configure linters for Solidity and JS.

Implementer: @liamzebedee  
Feature buddy: @nkuba

nkuba:
In this PR we:
- configured linters for Solidity and JavaScript code,
- added a pre-commit hook to enforce correct code formatting,
- added CircleCI job to execute linters on git push
- fixed formating of JS and Solidity files.

Closes #42


.circleci/config.yml
```diff
@@ -28,6 +28,22 @@ jobs:
...
+  lint:
+    executor: docker-node
+    steps:
+      - checkout
+      - attach_workspace:
+          at: solidity
+      - run:
+          name: Install dependencies
+          working_directory: ~/project/solidity
+          command: npm install
```
Shadowfiend: `npm ci` is the jam here.

.circleci/config.yml
```diff
@@ -210,6 +226,13 @@ jobs:
 
 workflows:
   version: 2
+  lint:
+    jobs:
+      - setup_github_package_registry:
+          context: github-package-registry
```
Shadowfiend: You probably don't need this anymore since the repo is open, though not certain about that.

solidity/.prettierrc.yaml
```diff
@@ -0,0 +1,2 @@
+semi: false
+trailingComma: "all"
```
Shadowfiend: We shouldn't have a custom `.prettierrc`.

solidity/.soliumrc.json
```diff
@@ -1,19 +1,16 @@
...
-      ],
-      "security/no-block-members": ["off", ["timestamp"]],
-      "security/no-call-value": "off",
-      "security/no-inline-assembly": "off"
-    }
-}
\ No newline at end of file
+    "security/no-call-value": "off",
+    "security/no-inline-assembly": "off"
+  }
```
Shadowfiend: We should consider unifying these rules at the `keep` level, as a side note. Haven't peeked into this yet, so not a blocking note or anything.

solidity/contracts/BondedECDSAKeep.sol
```diff
@@ -1,3 +1,4 @@
+/* solium-disable function-order */
```
Shadowfiend: Any particular reason for this? I would kind of rather just go with the function order in our linter. Fewer toggles is better.

solidity/package.json
```diff
@@ -43,7 +52,14 @@
     "@truffle/hdwallet-provider": "^1.0.25",
     "bn-chai": "^1.0.1",
     "chai": "^4.2.0",
+    "eslint": "^6.8.0",
+    "eslint-config-keep": "github:keep-network/eslint-config-keep",
```
Shadowfiend: Let's anchor this to a version tag.

solidity/package.json
```diff
@@ -43,7 +52,14 @@
...
+    "eslint-config-prettier": "^6.10.1",
```
Shadowfiend: Burn it please :) `eslint-config-keep` should handle this. I realize it doesn't currently deal with the Solidity plugin, but we still need to explicitly include the prettier dependency here anyway.

My suggestion:
- Kill `eslint-config-prettier`. Let `eslint-config-keep#0.3.0` handle the JS side of prettier. You still need the prettier dependency below for this, but you lose the explicit `prettier` invocation for `lint:js` and `lint:fix:js`.
- Continue to pull in `prettier-plugin-solidity` and run prettier in the `:sol` scripts.

I also want to see if I can pull prettier into solium's config, but that seems unlikely. There may be room to set something up with our core eslint config and Solidity though, TBD.

.circleci/config.yml
nkuba: [Nope](https://circleci.com/gh/keep-network/keep-ecdsa/5969?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link),  `Unable to authenticate, need: Basic realm="GitHub Package Registry"`  😞

solidity/contracts/BondedECDSAKeep.sol
nkuba: We tried to place the functions in a logical order as they may be used in the flow. But yeah, sure I can rearrange the functions order to satisfy solium.

solidity/.prettierrc.yaml
nkuba: Nice! I just copy-pasted it from tbtc 😉

solidity/contracts/BondedECDSAKeep.sol
Shadowfiend: I actually get this and it tends to be my first preference during initial dev, but there are some strong arguments for function ordering for visibility. Diligence also flagged this possibility. My main thing is I'd rather either go all in on ignoring this, or not ignore it ever heh. I'm already twitchy over the other exceptions we have to solium.

solidity/.prettierrc.yaml
Shadowfiend: Yeah, it predates my adding prettier to eslint and I haven't migrated it yet :grimacing:

solidity/contracts/BondedECDSAKeep.sol
nkuba: As I like the idea to conform to the rules and order the functions accordingly, there is one place where IMO it would be better to disable the check. 
It's in `BondedECDSAKeepVendor` contract where we have a long list of getters/setters pairs where getters are public and setters internal:
https://github.com/keep-network/keep-ecdsa/blob/d7bf095e964580e40eac0c4bae5575232853ec9c/solidity/contracts/BondedECDSAKeepVendor.sol#L158-L260

solidity/contracts/BondedECDSAKeep.sol
```diff
@@ -593,6 +436,164 @@ contract BondedECDSAKeep is IBondedECDSAKeep {
...
+    /// doesn't have a constructor. We provide keep parameters for each instance
+    /// function after cloning instances from the master contract.
+    /// @param _owner Address of the keep owner.
+    /// @param _members Addresses of the keep members.
+    /// @param _honestThreshold Minimum number of honest keep members.
+    /// @param _tokenStaking Address of the TokenStaking contract.
+    /// @param _keepBonding Address of the KeepBonding contract.
+    /// @param _keepFactory Address of the BondedECDSAKeepFactory that created
+    /// this keep.
+    function initialize(
```
nkuba: Moving this function below `external` ones hurts me because this is like a constructor that we use in the factory pattern.

.circleci/config.yml
Shadowfiend: I think this happened for a couple of reasons:
- You need to update `package-lock` by running `npm i`.
- You need to drop `.npmrc`.
- You need to drop the `publishConfig`.

However, I'll do all of this stuff in one step when I handle deployment later today.

solidity/migrations/external-contracts.js
```diff
@@ -1,12 +1,12 @@
 // Addresses of externally deployed smart contracts
-const RegistryAddress = '0xAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA'
-const TokenStakingAddress = '0xBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBB'
-const RandomBeaconAddress = '0xCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC'
-const TBTCSystemAddress = '0xDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDD'
+const RegistryAddress = "0x9aF7d664E6d7DB318c6BdE54ED77A7D99361F539"
+const TokenStakingAddress = "0x602c5f7d11eb786E14cDfBafe34E70c0b6717a8c"
+const RandomBeaconAddress = "0xB901EDDE70EcB59c1243f404a238fd97B78f1F54"
+const TBTCSystemAddress = "0x065c3Ab1D2cb67c70dAC17Cf3BD4B8A165ffceFa"
```
Shadowfiend: Is this on purpose? <_<

nkuba: Oh boy, no! It's fixed in https://github.com/keep-network/keep-ecdsa/pull/338/commits/43bfb27ab472dd21a4725ea6f27503eb09496eba.
Thanks for noticing that!

PR #332, opened 2020-03-28
Updated tss-lib dependency to version 1.3.0

nkuba:
In this PR we switch to tss-lib library version 1.3.0. 
The new version contains improvements and a [security vulnerability fix](https://github.com/binance-chain/tss-lib/releases/tag/v1.2.0).

The new version required couple of changes in our code to cover new parameters structure.


pdyraga:
There is something wrong about `--- FAIL: TestGenerateKeyAndSign (60.46s)` - it keeps failing all over again here. I tried to re-run the job four times and I am still getting the same.

I've run tests locally multiple times everything is correct. Tested with the client, everything looks good as well.

Issue #271, opened 2020-03-03
ETH-only delegating contracts


pdyraga:
Requirements:
- ETH owner should be able to delegate their ETH, run client, and participate in keeps as a signer.
- One keep should not have mixed KEEP operators and ETH operators.
- For ETH-only option, we should define the same roles as for KEEP staking: owner, operator, authorizer. We should also expect the owner to authorize operator contract.

Implementation:
- We add a separate fully-backed sortition pool selecting signers based on their bond weight.
- We add a separate ETH-only ECDSA keep factory.
- We add a separate ETH-only bonding contract.
- tBTC decides on which factory to use.
- The ETH-only bonding contract defines owner-operator-authorizer roles and implements operator contract authorization mechanism.
- Ban malicious players

pdyraga:
Notes on the existing implementation based on our recent conversations:
- We do not require over-bonding from ETH-only operators. They will need to supply the same bond as KEEP operators.
- It would be nice if we could remove and ban operator from a sortition pool on a proven misbehavior.
- We should have some kind of lock/undelegation period for ETH-only bonds to elimate the possibility of griefing (e.g. miner sees a delegation to operator X, they put their own delegation transaction to X, and unbond tokens immediatelly)

nkuba:
> We should have some kind of lock/undelegation period for ETH-only bonds to elimate the possibility of griefing (e.g. miner sees a delegation to operator X, they put their own delegation transaction to X, and unbond tokens immediatelly)

We may require ETH to be passed along with the delegation. The ETH will be deposited as unbonded value in bonding contract. Additionally we can lock unbonded value withdraw for a specific period of time since the delegation. What values would be appropirate for minimum amount of ETH passed on delegation and the lock period before withdraw?

pdyraga:
We decided to not implement this natively in the protocol but instead introduce a KEEP-only pool. At the same time, there are efforts in Keep community to implement community staking pools.

PR #269, opened 2020-03-02
ECDSAKeep Rewards

NicholasDotSol:
Refs: #446
ECDSAKeep Rewards to distribute KEEP tokens to keep owners based on a predetermined curve separated by intervals. 

Draft until:
- [ ] fully tested
- [ ] documented
- [ ] cleaned up
- [ ] optimized

PR #333, opened 2020-03-28
Add a concept of terminated keep, track keep creation timestamps in the factory

pdyraga:
Refs: #271

Code extracted from https://github.com/keep-network/keep-ecdsa/pull/269

To properly implement stake drop rewards we need to track the creation timestamp for each keep as well as know which keeps have been properly closed.

Instead of using a simple `isActive` boolean allowing to represent only two states, we have an enum allowing to indicate the keep is either active, closed, or terminated. A keep is getting terminated when bonds are seized.

The factory stores the information about keeps it created as well as their timestamps. This allows extracting all the information needed by stakedrop rewards contract. Also, we no longer have to notify the factory when a keep is closed.


solidity/contracts/BondedECDSAKeep.sol
```diff
@@ -610,7 +653,19 @@ contract BondedECDSAKeep is IBondedECDSAKeep {
...
+
+    /// @notice Gets the owner of the keep.
+    /// @return Address of the keep owner.
+    function getOwner() external view returns (address){
+        return owner;
+    }
+
+    /// @notice Gets the timestamp the keep was opened at.
+    /// @return timestamp the keep was opened at.
+    function getTimestamp() external view returns (uint256){
```
nkuba: The function name doesn't specify which timestamp we want to get, maybe we should name it `getKeepCreationTimestamp` or `getKeepOpenedTimestamp`?

solidity/contracts/BondedECDSAKeepFactory.sol
```diff
@@ -426,6 +431,27 @@ contract BondedECDSAKeepFactory is IBondedECDSAKeepFactory, CloneFactory {
...
+    /// @return The address of the keep at the given index.
+    function getKeepAtIndex(uint256 index) external view returns (address){
+        require(index < keeps.length, "Out of bounds.");
+        return keeps[index];
+    }
+
+    /// @notice Gets the creation timestamp of the given keep.
+    /// @return Timestamp the given keep was created at or 0 if this keep
+    /// was not created by this factory.
+    function getCreationTime(address _keep) external view returns (uint256) {
```
nkuba: `getKeepCreationTimestamp`?

solidity/test/BondedECDSAKeepFactoryTest.js
```diff
@@ -1046,6 +1046,65 @@ contract("BondedECDSAKeepFactory", async accounts => {
...
+                keepOwner,
+                bond,
+                { from: application, value: feeEstimate }
+            )     
+
+            let keep = await BondedECDSAKeep.at(keepAddress)
+
+            assert.isTrue(await keep.isActive(), "keep should be active")
+        })
+
```
nkuba: Can we add test cases for `getKeepAtIndex` where `index = keeps.length` and `index > keeps.length` to check if it reverts?

nkuba: We should also add a test case for `slashKeepMembers` function to check if `BondedECDSAKeep(msg.sender).isActive()` check is enforced.

solidity/contracts/BondedECDSAKeep.sol
pdyraga: Renamed to `getOpenedTimestamp()` since this is a function executed on a keep; same case as with `getOwner()`. I'll rename factory's `getCreationTime` to `getKeepOpenedTimestamp()`

solidity/contracts/BondedECDSAKeepFactory.sol
pdyraga: Renamed to `getKeepOpenedTimestamp`

Shadowfiend:
Weird question here: might it make more sense to have a second boolean instead of an enum? It seems like the question “is this keep still active” is quite distinct from the question “is this keep in good standing”, and they have relatively orthogonal information.

Not throwing a block in this PR, just floating the suggestion.

solidity/test/BondedECDSAKeepFactoryTest.js
pdyraga: See d6b741e

pdyraga: See https://github.com/keep-network/keep-ecdsa/pull/333/commits/823124cbf42af77886bcc21cc676fa14b36d0243

pdyraga:
> Weird question here: might it make more sense to have a second boolean instead of an enum? It seems like the question “is this keep still active” is quite distinct from the question “is this keep in good standing”, and they have relatively orthogonal information.
> 
> Not throwing a block in this PR, just floating the suggestion.

I was thinking about it but enum seems to me just easier in this case and possibly less error-prone. A keep is either active or closed or terminated. Never two of them at the same time. Enum nicely represent the state. If we do `isActive` and e.g. `closedHappily` than we have to put a little more effort in expressing the final state:
```
bool isActive
bool closedHappily

isActive() = isActive
isClosed() = !isActive && closedHappily
isTerminated() = !isActive && !closedHappily
```

> I'm wondering if it would be worth setting just `status = Status.Closed;` instead of referring to `markAsClosed` logic. Same below for terminated.

I can do it but I think the current solution is just a little more elegant as we are just making some functions public. Having just `status = Status.Closed;` makes sense but then we are getting deeper into implementation details of `BondedECDSAKeep`.

pkg/chain/chain.go
```diff
@@ -88,13 +88,20 @@ type BondedECDSAKeep interface {
 		signature *ecdsa.Signature,
 	) error // TODO: Add promise *async.SignatureSubmissionPromise
 
-	// OnKeepClosed is a callback that is invoked when a notification is sent
-	// about the closed keep.
+	// OnKeepClosed installs a callback that will be called on closing the
```
nkuba: Should we also update docs for all other `On*` functions for consistency to replace `is` with `installs`?

pkg/chain/ethereum/ethereum.go
```diff
@@ -86,6 +84,25 @@ func (ec *EthereumChain) OnKeepClosed(
 	)
 }
 
+// OnKeepTerminated is a callback that is invoked on-chain when keep is terminated.
```
nkuba: This doc doesn't match the one in `chain.go`.

solidity/contracts/BondedECDSAKeep.sol
```diff
@@ -436,6 +448,18 @@ contract BondedECDSAKeep is IBondedECDSAKeep {
...
     }
 
+    /// @notice Gets the owner of the keep.
+    /// @return Address of the keep owner.
+    function getOwner() external view returns (address) {
+        return owner;
+    }
+
+    /// @notice Gets the timestamp the keep was opened at.
+    /// @return timestamp the keep was opened at.
```
nkuba:
```suggestion
    /// @return Timestamp the keep was opened at.
```

solidity/contracts/BondedECDSAKeepFactory.sol
```diff
@@ -271,9 +272,38 @@ contract BondedECDSAKeepFactory is IBondedECDSAKeepFactory, CloneFactory {
...
+    /// @notice Gets the opened timestamp of the given keep.
+    /// @return Timestamp the given keep was opened at or 0 if this keep
+    /// was not created by this factory.
+    function getKeepOpenedTimestamp(address _keep)
+        external
+        view
+        returns (uint256)
+    {
+        return keepOpenedTimestamp[_keep];
+    }
```
nkuba: I assume these functions are required by stake drop? Maybe they could also be defined in some interface so there is a reference why they are required in this contract?

pdyraga: I'd prefer to not do any decision about it in this PR. I think it's better to postpone it until https://github.com/keep-network/keep-ecdsa/pull/269 gets finalized.

pkg/chain/ethereum/ethereum.go
pdyraga: I'd prefer to not have any function docs in `ethereum.go` and `local.go`. Everything is already documented in `chain.go`. Don't want to delay this PR so I'll just update them but I think this should be addressed in a separate PR where we'll clear up whole `ethereum.go` and `local.go`.

pkg/chain/chain.go
pdyraga: Please see https://github.com/keep-network/keep-ecdsa/pull/333/commits/26fb21ef3cf006eb37e786737ba2cf92d117d53a

pkg/chain/ethereum/ethereum.go
pdyraga: Please see https://github.com/keep-network/keep-ecdsa/pull/333/commits/26fb21ef3cf006eb37e786737ba2cf92d117d53a

solidity/contracts/BondedECDSAKeep.sol
pdyraga: Please see https://github.com/keep-network/keep-ecdsa/pull/333/commits/705323bfd8596ff4a5cfa16bf12ddb257172ae43

Issue #334, opened 2020-03-28
Possible context leak around public key submission monitoring


nkuba:
`go vet` reports a possible context leak for context initialized to be passed to `monitorKeepPublicKeySubmission` function.

```
✗ go vet ./...
# github.com/keep-network/keep-ecdsa/pkg/node
pkg/node/node.go:136:2: the monitoringCancel function is not used on all paths (possible context leak)
pkg/node/node.go:157:2: this return statement may be reached without using the monitoringCancel var defined on line 140
```

https://github.com/keep-network/keep-ecdsa/blob/4cc39c1d48ed4269b0260ecf21184d63b2f677e6/pkg/node/node.go#L136-L140

Refs: https://github.com/keep-network/keep-ecdsa/pull/284#discussion_r388334726

PR #338, opened 2020-03-30
Fixed possible context leak for monitoring key publication context

nkuba:
There was a possible context leak around key submission monitoring reported by go vet.

We moved context initialization inside the function that is actually using the context.
We added a channel to be able to abort monitoring goroutine in case of an error occurred in the parent function.

Closes: https://github.com/keep-network/keep-ecdsa/issues/334

### DIFF
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -264,7 +264,7 @@ func (n *Node) monitorKeepPublicKeySubmission(
 		if monitoringCtx.Err() == context.DeadlineExceeded {
 			logger.Warningf(
 				"monitoring of public key submission for keep [%s] "+
-					"has been cancelled due to timeout exceeding",
+					"has been cancelled due to exceeded timeout",
 				keepAddress.String(),
 			)
 		}
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -137,18 +137,15 @@ func (n *Node) GenerateSignerForKeep(
 		return nil, fmt.Errorf("failed to serialize public key: [%v]", err)
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/node/node.go
```diff
@@ -256,13 +260,19 @@ func (n *Node) monitorKeepPublicKeySubmission(
 			event.SubmittingMember,
 			event.ConflictingPublicKey,
 		)
-	case <-ctx.Done():
-		if ctx.Err() == context.DeadlineExceeded {
+	case <-monitoringCtx.Done():
+		if monitoringCtx.Err() == context.DeadlineExceeded {
 			logger.Warningf(
 				"monitoring of public key submission for keep [%s] "+
 					"has been cancelled due to timeout exceeding",
```
pdyraga: I believe it should be `exceeding timeout` or `timeout exceeded`.

pkg/node/node.go
```diff
@@ -256,13 +260,19 @@ func (n *Node) monitorKeepPublicKeySubmission(
 			event.SubmittingMember,
 			event.ConflictingPublicKey,
 		)
-	case <-ctx.Done():
-		if ctx.Err() == context.DeadlineExceeded {
+	case <-monitoringCtx.Done():
+		if monitoringCtx.Err() == context.DeadlineExceeded {
```
pdyraga: Do we need this check? When the context gets cancelled we always have a timeout in this case as far as I understand.

nkuba: It can get canceled also by calling context cancel function.

PR #339, opened 2020-03-30
__beaconCallback(uint256) to update group selection seed

pdyraga:
Refs https://github.com/keep-network/keep-core/issues/1527

See https://github.com/keep-network/keep-core/pull/1532
See keep-network/keep-common#22

The random beacon does no longer allow to request relay entry passing signature of callback function to be called. Instead, it expects `__beaconCallback(uint256)` function to be declared by the contract receiving the random number.

Additionally:
- [Fixed tests](https://github.com/keep-network/keep-ecdsa/pull/339/commits/941f7aa82be68790aab06a94c249d7dce4374a6f) after bumping up `keep-core` reference
- [Reduced beacon callback gas estimate](https://github.com/keep-network/keep-ecdsa/pull/339/commits/7b04db60c466ec375cc60296ca987290b4e7f617)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,7 +18,7 @@ require (
 	github.com/ethereum/go-ethereum v1.9.10
 	github.com/gogo/protobuf v1.3.1
 	github.com/ipfs/go-log v0.0.1
-	github.com/keep-network/keep-common v0.1.1-0.20200228111656-16dd370ca897 // TODO: update to released version
+	github.com/keep-network/keep-common v0.1.1-0.20200330215727-8ca95ff9de56 // TODO: update to released version
 	github.com/keep-network/keep-core v0.10.0
 	github.com/olekukonko/tablewriter v0.0.4 // indirect
 	github.com/pkg/errors v0.9.1
--- a/solidity/contracts/BondedECDSAKeepFactory.sol
+++ b/solidity/contracts/BondedECDSAKeepFactory.sol
@@ -77,7 +77,7 @@ contract BondedECDSAKeepFactory is IBondedECDSAKeepFactory, CloneFactory {
     // gas required to call `__beaconCallback` function in the worst-case
     // scenario with all the checks and maximum allowed uint256 relay entry as
     // a callback parameter.
-    uint256 public constant callbackGas = 41830;
+    uint256 public constant callbackGas = 30000;
 
     // Random beacon sends back callback surplus to the requestor. It may also
     // decide to send additional request subsidy fee. What's more, it may happen
@@ -101,6 +101,10 @@ contract BondedECDSAKeepFactory is IBondedECDSAKeepFactory, CloneFactory {
         tokenStaking = TokenStaking(_tokenStaking);
         keepBonding = KeepBonding(_keepBonding);
         randomBeacon = IRandomBeacon(_randomBeacon);
+
+        // initial value before the random beacon updates the seed
+        // https://www.wolframalpha.com/input/?i=pi+to+78+digits
+        groupSelectionSeed = 31415926535897932384626433832795028841971693993751058209749445923078164062862;
     }
 
     /// @notice Adds any received funds to the factory reseed pool.
--- a/solidity/contracts/BondedECDSAKeepFactory.sol
+++ b/solidity/contracts/BondedECDSAKeepFactory.sol
@@ -308,10 +308,7 @@ contract BondedECDSAKeepFactory is IBondedECDSAKeepFactory, CloneFactory {
     /// @dev The function is expected to be called in a callback by the random
     /// beacon.
     /// @param _relayEntry Beacon output.
-    function __beaconCallback(uint256 _relayEntry)
-        external
-        onlyRandomBeacon
-    {
+    function __beaconCallback(uint256 _relayEntry) external onlyRandomBeacon {
         groupSelectionSeed = _relayEntry;
     }
 
--- a/solidity/contracts/BondedECDSAKeepFactory.sol
+++ b/solidity/contracts/BondedECDSAKeepFactory.sol
@@ -74,7 +74,7 @@ contract BondedECDSAKeepFactory is IBondedECDSAKeepFactory, CloneFactory {
     uint256 public constant minimumBond = 1;
 
     // Gas required for a callback from the random beacon. The value specifies
-    // gas required to call `setGroupSelectionSeed` function in the worst-case
+    // gas required to call `__beaconCallback` function in the worst-case
     // scenario with all the checks and maximum allowed uint256 relay entry as
     // a callback parameter.
     uint256 public constant callbackGas = 41830;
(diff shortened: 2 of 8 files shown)


solidity/test/contracts/RandomBeaconStub.sol
```diff
@@ -1,10 +1,10 @@
 pragma solidity ^0.5.4;
 
-import "@keep-network/keep-core/contracts/IRandomBeacon.sol";
+//import "@keep-network/keep-core/contracts/IRandomBeacon.sol";
 
 /// @title Random Beacon Service Stub
 /// @dev This contract is for testing purposes only.
-contract RandomBeaconStub is IRandomBeacon {
+contract RandomBeaconStub /*is IRandomBeacon*/ {
```
pdyraga: I hate what I did here but this can't compile until we release a new `keep-core` version or fix `-pre` deployments. I am seriously considering turning `is IRandomBeacon` into a TODO. Thoughts?

go.mod
```diff
@@ -18,7 +18,7 @@ require (
 	github.com/ethereum/go-ethereum v1.9.10
 	github.com/gogo/protobuf v1.3.1
 	github.com/ipfs/go-log v0.0.1
-	github.com/keep-network/keep-common v0.1.1-0.20200228111656-16dd370ca897 // TODO: update to released version
+	github.com/keep-network/keep-common v0.1.1-0.20200330215727-8ca95ff9de56 // TODO: update to released version
 	github.com/keep-network/keep-core v0.10.0
```
nkuba: We will need to update `keep-core` dependency version as soon as it's released.

solidity/contracts/BondedECDSAKeepFactory.sol
```diff
@@ -68,7 +68,7 @@ contract BondedECDSAKeepFactory is IBondedECDSAKeepFactory, CloneFactory {
     uint256 public constant minimumBond = 1;
 
     // Gas required for a callback from the random beacon. The value specifies
-    // gas required to call `setGroupSelectionSeed` function in the worst-case
+    // gas required to call `__beaconCallback` function in the worst-case
     // scenario with all the checks and maximum allowed uint256 relay entry as
     // a callback parameter.
     uint256 public constant callbackGas = 41830;
```
nkuba: Does the gas value remain the same after changes on the beacon side?

solidity/test/BondedECDSAKeepFactoryTest.js
```diff
@@ -1140,7 +1140,7 @@ contract("BondedECDSAKeepFactory", async (accounts) => {
     }
   })
 
-  describe("setGroupSelectionSeed", async () => {
+  describe("__beaconCallback", async () => {
     const newGroupSelectionSeed = new BN(2345675)
```
nkuba: `newGroupSelectionSeed` -> `newRelayEntry`? We updated naming in https://github.com/keep-network/keep-ecdsa/pull/339/files#diff-5325d7067cb217b70e5bed5c308051faR281 so it would be worth aligning it here as well.

solidity/contracts/BondedECDSAKeepFactory.sol
```diff
@@ -277,12 +277,12 @@ contract BondedECDSAKeepFactory is IBondedECDSAKeepFactory, CloneFactory {
     /// @notice Sets a new group selection seed value.
     /// @dev The function is expected to be called in a callback by the random
     /// beacon.
-    /// @param _groupSelectionSeed New value of group selection seed.
-    function setGroupSelectionSeed(uint256 _groupSelectionSeed)
+    /// @param _relayEntry Beacon output.
+    function __beaconCallback(uint256 _relayEntry)
```
nkuba: Thoughts on defining an interface on beacon side that would define `function __beaconCallback(uint256 _relayEntry)` function and be implemented by `BondedECDSAKeepFactory` contract?

solidity/test/contracts/RandomBeaconStub.sol
nkuba: We can't merge this PR without `keep-core` changes being released. The release will solve the problem you mentioned above.

pdyraga: No, this is not a good idea. The build result will depend on the currently cached `keep-core` version. Let's hold on with merging this PR until we release a new `keep-core`.

pdyraga: @nkuba We commented here at the same time :)

solidity/contracts/BondedECDSAKeepFactory.sol
pdyraga: It is the callback gas on our side, I don't see a reason why it could change but I'll double-check.

solidity/test/contracts/RandomBeaconStub.sol
pdyraga: It's back here: https://github.com/keep-network/keep-ecdsa/pull/339/commits/85fa8f073b8adb9761f526fc9997c65b5489b84b

go.mod
pdyraga: Hope to tag `keep-core` and `keep-common` end of this week.

solidity/contracts/BondedECDSAKeepFactory.sol
pdyraga: No thoughts on this at the moment, opened an issue on `keep-core` side to address it next week: https://github.com/keep-network/keep-core/issues/1537

solidity/test/BondedECDSAKeepFactoryTest.js
pdyraga: Yep! See https://github.com/keep-network/keep-ecdsa/pull/339/commits/c9407c6aa3931426669046f879051e5b7b591f08

go.mod
nkuba: Can this be updated already?

pdyraga: I'd prefer to leave it as-is for now and change to RC version end of this week.

solidity/contracts/BondedECDSAKeepFactory.sol
pdyraga: The gas estimate was correct but I realized we can get a discount of 10k, see https://github.com/keep-network/keep-ecdsa/pull/339/commits/7b04db60c466ec375cc60296ca987290b4e7f617

You can play with it yourself, just remember about using beacon as a sender so that the modifier in callback lets you go:
```
const BondedECDSAKeepFactory = artifacts.require("BondedECDSAKeepFactory.sol")
const BN = web3.utils.BN;

module.exports = async function () {
  try {
    let factory = await BondedECDSAKeepFactory.at("0x7741C7E704388bFf29f078bF95d04A86C534Bce9")

    let newEntry = new BN(2).pow(new BN(256)).subn(2)

    console.log('estimating...')
    let gas = await factory.__beaconCallback.estimateGas(newEntry)
    console.log("gas estimate: " + gas)

    console.log('calling...')
    let tx = await factory.__beaconCallback(newEntry)
    console.log(tx);
  } catch (error) {
    console.log(error)
  }
}
```

PR #340, opened 2020-03-31
Clearing up README: there is no developer-specific documentation directory

pdyraga:
We do not have any `docs/` in `keep-ecdsa`, just `README`. Removing a non-existing link here.


Shadowfiend:
Heh, this was a great catch in the wrong direction. I meant to move everything past the directory layout into `docs/README.adoc` but evidently forgot.

PR #344, opened 2020-03-31
Circle Migrations: Don't fail if Tenderly fails

sthompson22:
Tenderly has a handful of kinks we're still trying to work out for
reliable push.  Until then we're not going to fail the job if it fails.

This is a rip of https://github.com/keep-network/keep-core/commit/adae0ef53ebd1848cd89e7a4887d05fdc0580329#diff-d1b31ada11a15635d2d99d71b9579b7a

There's a bonus `TBTCSystem` address update for `keep-test` configs. Picks up the latest migration done [here](https://app.circleci.com/pipelines/github/keep-network/tbtc/2096/workflows/55bccba1-f854-4aff-af60-aa4442772e1b/jobs/18151)

### DIFF
--- a/solidity/scripts/circleci-migrate-contracts.sh
+++ b/solidity/scripts/circleci-migrate-contracts.sh
@@ -63,7 +63,8 @@ ssh utilitybox << EOF
   echo ">>>>>>FINISH Contract Migration FINISH>>>>>>"
   echo "<<<<<<START Tenderly Push START<<<<<<"
   tenderly login --authentication-method token --token $TENDERLY_TOKEN
-  tenderly push --networks $ETH_NETWORK_ID --tag keep-ecdsa --tag $GOOGLE_PROJECT_NAME --tag $BUILD_TAG
+  tenderly push --networks $ETH_NETWORK_ID --tag keep-ecdsa \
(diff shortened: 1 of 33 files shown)


solidity/scripts/circleci-migrate-contracts.sh
```diff
@@ -62,8 +62,12 @@ ssh utilitybox << EOF
   ./node_modules/.bin/truffle migrate --reset --network $TRUFFLE_NETWORK
   echo ">>>>>>FINISH Contract Migration FINISH>>>>>>"
   echo "<<<<<<START Tenderly Push START<<<<<<"
+  # Temporary fix for some odd push issues; OpenZeppelin contracts are
+  # referenced at two paths due to older dependencies, leading to problems
+  # resolving contracts during tenderly push.
```
Shadowfiend, line 67: This actually doesn't pull in the `ln -s` for `@openzeppelin`. Are we cool with that?

sthompson22: Frankly, I wasn't sure we had the same dependency issue here.

sthompson22: Updated to include the symlink

sthompson22:
Slammin' this.

PR #345, opened 2020-04-02
Configurable TSS pre-parameters generation timeout

nkuba:
We used to set a timeout for a pre-parameters generation to a fixed value of `120 seconds`. It felt like a way more than enough when running it on machines with 4+ CPU cores. Unfortunately, it may be to low for less powerful machines. Here we added a possibility to set the timeout to an arbitrary value based on the operator's observations and resources the server has.

The value can be configured in a toml config file under `PreParamsGenerationTimeout`. If not set the default `2 minutes` will be assumed.

We also remove `PreParamsGenerationConcurrency` parameters as it's not longer required after the concurrency fix in `tss-lib`.

### DIFF
--- a/configs/config.toml.SAMPLE
+++ b/configs/config.toml.SAMPLE
@@ -38,4 +38,4 @@
 # should be provided based on resources available on the machine running the client.
 # This is an optional parameter, if not provided timeout for TSS protocol
 # pre-parameters generation will be set to `2 minutes`.
-#  PreParamsGenerationTimeout = "2m"
+#  PreParamsGenerationTimeout = "2m30s"
--- a/configs/config.toml.SAMPLE
+++ b/configs/config.toml.SAMPLE
@@ -29,11 +29,6 @@
 #   # AnnouncedAddresses = ["/dns4/example.com/tcp/3919", "/ip4/80.70.60.50/tcp/3919"]
 
 [TSS]
-# Level of concurrency for TSS protocol pre-parameters generation. The value
-# should be provided based on resources available on the machine running the client.
-# This is an optional parameter, if not provided concurrency level for TSS protocol
-# will be set to `3` which is a minimum to run pre-parameters generation process.
-#  PreParamsGenerationConcurrency = 3
 # Timeout for TSS protocol pre-parameters generation. The value
 # should be provided based on resources available on the machine running the client.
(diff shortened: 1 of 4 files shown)


configs/config.toml.SAMPLE
```diff
@@ -34,3 +34,8 @@
 # This is an optional parameter, if not provided concurrency level for TSS protocol
 # will be set to `3` which is a minimum to run pre-parameters generation process.
```
pdyraga, line 35: Is it still valid explanation after https://github.com/keep-network/keep-ecdsa/pull/347

configs/config.toml.SAMPLE
```diff
@@ -34,3 +34,8 @@
...
 #  PreParamsGenerationConcurrency = 3
+# Timeout for TSS protocol pre-parameters generation. The value
+# should be provided based on resources available on the machine running the client.
+# This is an optional parameter, if not provided timeout for TSS protocol
+# pre-parameters generation will be set to `2 minutes`.
+#  PreParamsGenerationTimeout = "2m"
```
pdyraga, line 41: Maybe we should use "2m30s" as an example on how to format this parameter?

configs/config.toml.SAMPLE
nkuba: If the value is <3 it will be handled by tss-lib and set to `3` anyways.

pkg/node/params_pool.go
```diff
@@ -29,13 +45,25 @@ func (n *Node) InitializeTSSPreParamsPool() {
 
 func (t *tssPreParamsPool) pumpPool() {
 	for {
+		logger.Info("generate new tss pre parameters")
```
pdyraga, line 48: `s/generate/generating`

configs/config.toml.SAMPLE
pdyraga: For 3 how many concurrent generations are started? Worth explaining here what 3 means.

pdyraga: Or... we can remove it now with the fix in `tss-lib`.

nkuba: We no longer need this property at all as the bug on tss-lib side was fixed https://github.com/keep-network/keep-ecdsa/pull/347

PR #350, opened 2020-04-03
keep-dev: TBTCSystem contract update

sthompson22:
Two things happening here:

1. Update to the latest `TBTCSystem`
2. Enable `nocolor` IPFS format logging

2 is to aid with log parsing in Stackdriver, as the default logging gets sliced up in weird ways when shipped to Stackdriver.


sthompson22:
This is rolled out to `keep-dev`, `keep-test` will go out on the next RC deployment.

Issue #349, opened 2020-04-03
Signer generation chokes when there are not enough pre-params in the pool


nkuba:
When the pre-params pool gets empty and multiple concurrent requests to generate signers come, the clients are getting out of sync and can't handle announcements and readiness to start the protocol.

We need to investigate how can we improve retries to work seamlessly.

We should consider not consuming pre-params until the protocol is actually started. Right now pre-params are consumed for each retry even if we fail on a stage before they are actually needed (readiness announcement). 

To test that run 3 clients and request more keeps than the pre-params pool size is (e.g. 10+)

PR #353, opened 2020-04-03
Updated TSS protocol to reuse pre parameters between retries

nkuba:
Previously when the signer generation failed there was a retry mechanism invoking the process once again. Unfortunately, it wasn't efficient as pre-parameters had to be generated again, even though they were not used for actual protocol execution, as the failure could happen before the
messages were exchanged between parties.

Here we introduce a change that will reuse pre parameters for protocol execution up to the moment when the messages are actually exchanged between parties. Before running the phase where parties interact we flush the parameters so if the phase fails we will need to provide fresh
set of parameters.

TODO:
- [ ] Update and add unit tests

Closes: https://github.com/keep-network/keep-ecdsa/issues/349


pdyraga:
We replaced it with #357

PR #357, opened 2020-04-04
Pre-params generation and timeout handling improvements

pdyraga:
Closes: #349 

It could happen that under a high load, clients could not generate public keys for some keeps. 
The problem was a combination of multiple factors, and _(hopefully)_ each one is addressed here.

Pre-parameters for threshold key generation are expensive to generate. We keep them in a pool that is automatically re-filled in a background goroutine. When there are no parameters left in the pool, the client needs to wait until new ones are generated.

There is a global 2.5 hours timeout for generating threshold key and publishing it to the chain. To generate a key, the client needs to go through four steps:
- announce its presence in the broadcast channel
- generate threshold signer, set up unicast communication channels, and announce readiness
- execute threshold key generation
- publish the result to the chain

There is a 2 minutes timeout for announcing presence, 2 minutes timeout for announcing readiness and 6 minutes timeout for threshold key generation. If any of those steps times out, we return to the beginning and retry.

Here was the problem. When retrying, we were taking new pre-parameters from the pool, throwing the previous ones away. It was not a problem when pre-parameters were instantly available but it was a huge problem when they were not and the client had to generate them again. 
What's even worse, in this case, they were not generated before the client entered the announcement part but after the client passed the announcement and before it signalled readiness. If one of the clients participating in the keep was generating those parameters for more than 2 minutes, two other clients were timing out and going back to the announcement phase. That one client, with parameters eventually generated, was signalling readiness and timing out after 2 minutes because no one else was there. All three clients were then at the very beginning of the protocol retrying presence announcement and after that taking new parameters from the pool and trying to signal readiness. The old parameters were thrown away so the same problem could happen. The situation was even worse when the clients participated in multiple key generations at once because pre-parameters cannot be reused and are uniquely generated for the given protocol execution.

There are four changes that should solve this problem.

First of all, we introduce `params.Box`. Instead of passing raw pre-parameters to key generation, we pass this box. When the node shares its pre parameters with other nodes, box content is destroyed. Until then, it's fine to pass the box around and consume its content for any calculations needed between retried key-generation attempts. It means that if the presence announcement timed out, we don't need to calculate new pre-parameters. If readiness signalling timed out, we also don't need to calculate new pre-parameters. The only scenario when we generate new pre-parameters is when the threshold signing protocol failed/timed out.

Secondly, we move the moment when we fetch pre-parametrs from the pool to before we announce the presence. It simplifies timing out and should buy as a little more time for the protocol in case the parameters need to be generated.

Also, we increase the pool size to 20. It should let the client participate in 20 key generations with no delay. The previous value (5) was a little too low for production use. 

I have also revisited and refactored all retries in key generation and signing protocol. Everything happens now inside `Node` except key submission retries in `ethereum.go`.

Last but not least, implemented a simple stress test for local use opening 50 keeps: 5 keeps at once in 10 iterations and observing how public keys are generated. Executed it locally with no problems after the changes here.


pkg/ecdsa/params/params_box.go
```diff
@@ -0,0 +1,55 @@
...
+func NewBox(params *keygen.LocalPreParams) (ppb *Box) {
+	return &Box{
+		params: params,
+	}
+}
+
+// Content gets the box content or error if the content has been previously
+// destroyed.
+func (b *Box) Content() (*keygen.LocalPreParams, error) {
+	if b.params == nil {
```
nkuba, line 38: Can we use `isEmpty` here?

pkg/ecdsa/params/params_box.go
```diff
@@ -0,0 +1,55 @@
...
+		return nil, fmt.Errorf("box destroyed")
```
nkuba, line 39: `box is empty`?

pkg/ecdsa/params/params_box.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	return b.params, nil
+}
+
+// IsEmpty returns true if the box content has not been destroyed.
```
nkuba, line 45: `returns true if the box content has been destroyed`

pkg/ecdsa/params/params_box.go
```diff
@@ -0,0 +1,55 @@
...
+// box content should be destroyed. Until then, it's fine to pass the box
+// around and consume its content for any calculations needed between retried
+// key-generation attempts.
+type Box struct {
+	params *keygen.LocalPreParams
+}
+
+// NewBox creates a new PreParamsBox with the provided key generation pre-params
+// inside.
+func NewBox(params *keygen.LocalPreParams) (ppb *Box) {
```
nkuba, line 29: We don't use `ppb` defined for return.

pkg/ecdsa/params/params_box_test.go
```diff
@@ -0,0 +1,69 @@
+package params
```
nkuba, line 1: Does it have to be a separate package? Can't we hold it in `tss` package?

pkg/node/node.go
```diff
@@ -101,105 +103,261 @@ func createAddressFilter(
...
 
-	// Publish signer's public key on ethereum blockchain in a specific keep
-	// contract.
-	serializedPublicKey, err := eth.SerializePublicKey(signer.PublicKey())
-	if err != nil {
-		return nil, fmt.Errorf("failed to serialize public key: [%v]", err)
-	}
-
 	monitoringAbort := make(chan interface{})
 	go n.monitorKeepPublicKeySubmission(ctx, monitoringAbort, keepAddress)
```
nkuba, line 213: What will happen when one of the members fails on key submission?
Maybe we should block on the monitoring and wait for the successful key establishment notification from the keep and restart generation in case of timeout? Or just retry submission as we do for signatures?

pkg/node/node.go
```diff
@@ -101,105 +103,261 @@ func createAddressFilter(
...
+				return nil
+			}
+
+			// Our public key submission transaction failed. We are going to
+			// wait for some time and then retry from the beginning.
+			logger.Errorf(
+				"failed to submit signature: [%v]; will retry after 10 minutes",
+				err,
+			)
+			time.Sleep(10 * time.Minute)
```
nkuba, line 354: Why 10 minutes? Isn’t it too much? Maybe we could monitor for an event that signature was already submitted by other member and return?

pkg/ecdsa/params/params_box_test.go
pdyraga: I don't want to hold it in `tss` because then the members of `tss` can refer to the box content directly while I want them to go through `Content` function.

nkuba:
> I have also revisited and refactored all retries in key generation and signing protocol. Everything happens now inside Node except key submission retries in ethereum.go.

Should we be consistent and handle key and signature submissions in the same place?

pkg/ecdsa/params/params_box.go
pdyraga: We can, good point.

pdyraga:
> Should we be consistent and handle key and signature submissions in the same place?

Those are two separate cases IMO. 

For key generation, we expect every keep member to publish their key. We are not monitoring other members' keys because it does not change our behaviour. We need to submit our key and that's it. If there is a failure, this is an infrastructure-level failure and should be retried inside `ethereum.go`. One possible scenario that we observed on Infura is when a public key submission fails because of the new cloned contract not registered yet on one of Ethereum clients behind the load balancer. This is not a kind of a problem you want to deal with in `node.go`.

For public key submission, ideally, only one keep member could be the submitter. We don't have a protocol for that, so they all try to submit but they also monitor other clients' submission and they does change client's behaviour. If the submission failed it is possible it wasn't something wrong with the infrastructure. It could be a perfectly valid situation when one of the members was faster. This is why we handle retries in `node.go` in this case. And we can't retry inside `ethereum.go` in this case because not every failed transaction is the submission protocol failure in this case.

pkg/node/node.go
pdyraga: Inside `ethereum.go` we have a retry for a situation when the public key submission fails. If the public key submission fails although it was retried multiple times, there is really nothing we can do. Some other member could submit their key so we can't re-run the protocol one more time. Implementing an on-chain cleanup for failed key submission can be an attack vector: I failed to manipulate threshold key generation and I don't publish the result so that I have yet one chance.

pkg/node/node.go
pdyraga: I did not want to complicate this function even more. That's why I added this sleep and in the next iteration, we check if `IsAwaitingSignature`. 

Now when I look at it, I do agree 10 minutes is too much. On the other hand, anything below 30 seconds is probably too aggressive given the block mining times. What if I set it to 1 minute?

pkg/ecdsa/params/params_box_test.go
nkuba: So maybe we could keep it under `/pkg/ecdsa/tss/params`? This package is strictly related to tss protocol, so we should probably keep it there.

pkg/node/node.go
```diff
@@ -101,105 +103,261 @@ func createAddressFilter(
...
+		//
+		// If signer announcement fails, we retry from the beginning.
+		memberIDs, err := n.AnnounceSignerPresence(
+			ctx,
+			operatorPublicKey,
+			keepAddress,
+			members,
+		)
+		if err != nil {
+			logger.Errorf("failed to announce signer presence: [%v]", err)
```
nkuba, line 161: Since we're in a retry loop maybe we should log just a warning, not an error.

pkg/node/node.go
```diff
@@ -101,105 +103,261 @@ func createAddressFilter(
...
+			ctx,
+			keepAddress.Hex(),
+			memberID,
+			memberIDs,
+			uint(len(memberIDs)-1),
+			n.networkProvider,
+			preParamsBox,
+		)
+		if err != nil {
+			logger.Errorf("failed to generate threshold signer: [%v]", err)
```
nkuba, line 179: Since we're in a retry loop maybe we should log just a warning, not an error.

pkg/node/node.go
nkuba: > I did not want to complicate this function even more. That's why I added this sleep and in the next iteration, we check if IsAwaitingSignature.

I believe we should prefer web socket events monitoring vs contract calls to optimize infrastructure costs.

> What if we set it to 1 minute?
1 minute is a fair value.

pkg/node/node.go
nkuba: Since publishing the key is so important and we cannot retry the whole process maybe we could give it more chances to submit the tx? Right now we have just 4 retries every 250 ms:
https://github.com/keep-network/keep-ecdsa/blob/3f2d73fe07e9f897ff0fb5e138ac7047a590aea0/pkg/chain/ethereum/ethereum.go#L225-L241

While testing it with Infura I observed a couple of times that this time frame was not enough to submit the transaction successfully.

pdyraga: I'll increase it to 10 retries, every 12 seconds, cc @dimpar 
It gives a total of 2 minutes.

pdyraga: See https://github.com/keep-network/keep-ecdsa/pull/357/commits/f23eb95888d54ccaaab6576a078d8d6a96121f36

pkg/node/node.go
pdyraga: See https://github.com/keep-network/keep-ecdsa/pull/357/commits/c21d70d1e0b442b1eee83f02d058ca4fb3cfefae

pkg/ecdsa/params/params_box_test.go
pdyraga: Good point. Moved it there; see https://github.com/keep-network/keep-ecdsa/pull/357/commits/159264584c822e5f3aa3bf9f61eebbb7ccc3f59f

pkg/node/node.go
pdyraga: I think it should be an error. The log level should be about a severity, not about the fact whether are going to retry or not. As an example, in `keep-core` we log on warning level when one of the nodes provided invalid message and this message was dropped and that node marked as misbehaving. This is not an error because the protocol can continue without that single misbehaving member with no problem. We warn that someone is misbehaving, though. In this case in `node.go`, the protocol cannot continue with all members announcing their presence so we should use error level. This is a kind of a situation that should potentially light the red bulb in the operator's room.

pkg/node/node.go
pdyraga: Same explanation as for the error<>warning comment before.

pkg/node/node.go
pdyraga: > I believe we should prefer web socket events monitoring vs contract calls to optimize infrastructure costs.

Agreed though I am not that concerned about this one here. This action is not executed very often and if we are looking for savings we should look them in sortition pool monitoring or minimum stake watchtower. More importantly, this one is nothing compared to the two previously mentioned and I don't have the bandwidth to refactor it right now and this is how it's in `master` so 🤷‍♂ 

Taking #364 into account, I think we should even execute `IsAwaitingSignature` another time after some delay to double-check our block was accepted 😬

pkg/node/node.go
```diff
@@ -101,105 +103,261 @@ func createAddressFilter(
...
-		keepMembersIDs,
-		uint(len(keepMembersIDs)-1),
-		n.networkProvider,
-		n.tssParamsPool.get(),
-	)
-	if err != nil {
-		return nil, fmt.Errorf("failed to generate threshold signer: [%v]", err)
+	logger.Infof("Getting new key-generation pre-parameters from the pool")
+	preParamsBox := params.NewBox(n.tssParamsPool.get())
+	logger.Infof("Received new key-generation parameters from the pool")
```
nkuba, line 121: I hesitate whether at least one of these loggers should be changed to `debug` level.

pkg/node/node.go
```diff
@@ -101,105 +103,261 @@ func createAddressFilter(
...
+			attemptCounter,
+		)
+
+		// If we are re-attempting the key generation, pre-parameters in the box
+		// could be destroyed because they were shared with other members.
+		// In this case, we need to re-generate them.
+		if preParamsBox.IsEmpty() {
+			logger.Infof("Getting new key-generation pre-parameters from the pool")
+			preParamsBox = params.NewBox(n.tssParamsPool.get())
+			logger.Infof("Received new key-generation parameters from the pool")
```
nkuba, line 139: Same here as in the comment above.

pkg/node/node.go
```diff
@@ -101,105 +103,261 @@ func createAddressFilter(
...
+
+		// We have the signature so now we need to publish it.
+		// This function implements internal retries so we do not need to
+		// retry here.
+		return n.publishSignature(ctx, keepAddress, digest, signature)
+	}
+}
 
-	if err := n.ethereumChain.SubmitSignature(keepAddress, signature); err != nil {
+// publisSignature takes the provided signature and attempts to publish it to
```
nkuba, line 276:
```suggestion
// publishSignature takes the provided signature and attempts to publish it to
```

pkg/node/node.go
```diff
@@ -101,105 +103,261 @@ func createAddressFilter(
...
+			// Check if we failed because someone else submitted in the meantime
+			// or because something wrong happened with our transaction.
+			if !isAwaitingSignature {
+				logger.Infof(
+					"signature for keep [%s] already submitted: [%+x]",
+					keepAddress.String(),
+					digest,
+				)
+				return nil
+			}
```
nkuba, line 346: Maybe we don't need this part as we're going to retry from the beginning and we've got this check starting in L303?

pkg/node/node.go
```diff
@@ -101,105 +103,261 @@ func createAddressFilter(
...
+// in case of a failure.
+//
+// We do implement a retry in this function because the retry mechanism is much
+// more complex than in case of e.g. publishSignerPublicKey. Although all keep
+// members are supposed to try to publish the signature, only one transaction
+// succeeds. For each attempt, we need to check if the keep still awaits
+// a signature. Also, we need to implement some sane delay between attempts so
+// that we do not waste gas.
+func (n *Node) publishSignature(
+	ctx context.Context,
```
nkuba, line 287: We're not using `ctx` in this function, so can remove this parameter.
On the other hand maybe we want to use this to check in the retry loop that we reached the timeout?

pkg/node/node.go
```diff
@@ -101,105 +103,261 @@ func createAddressFilter(
...
+			logger.Errorf(
+				"failed to submit signature: [%v]; will retry after 10 minutes",
+				err,
+			)
+			time.Sleep(1 * time.Minute)
+			continue
+		}
 
-	return nil
+		logger.Infof("Signature submitted")
```
nkuba, line 358:
```suggestion
		logger.Infof("signature submitted")
```

solidity/integration/stress_test.js
```diff
@@ -0,0 +1,96 @@
...
+  const application = accounts[5]
+
+  let keepFactory
+
+  try {
+    const keepVendor = await BondedECDSAKeepVendorImplV1.at(
+      (await BondedECDSAKeepVendor.deployed()).address
+    )
+    const keepFactoryAddress = await keepVendor.selectFactory()
+    keepFactory = await BondedECDSAKeepFactory.at(keepFactoryAddress)
```
nkuba, line 32: Maybe in this test we don't need to call the vendor?
```suggestion
    keepFactory = await BondedECDSAKeepFactory.deployed()
```

solidity/integration/stress_test.js
```diff
@@ -0,0 +1,96 @@
...
+
+const groupSize = 3
+const threshold = 3
+const bond = 10
+
+module.exports = async function () {
+  const accounts = await web3.eth.getAccounts()
+
+  const keepOwner = accounts[4]
+  const application = accounts[5]
```
nkuba, line 23: We added `accounts[4]` to operators list, so maybe want to replace `keepOwner` here?
```suggestion
  const keepOwner = accounts[5]
  const application = accounts[6]
```

pkg/node/node.go
pdyraga: I am not sure, as an operator, I'd like to know how fast my node generates pre-parameters so that I can scale it up if needed. I probably don't want to enable debug level for that.

pkg/node/node.go
pdyraga: I was thinking about it and I also don't like the code duplication here but I think this part is what's going to be executed in the majority of cases and it's nice to do not wait one minute with logging it. If we have a one-second delay, I'd remove it but for one minute, I think it makes sense to have it here.

pkg/node/node.go
pdyraga: > On the other hand maybe we want to use this to check in the retry loop that we reached the timeout?

We do, good catch. Fixed.

pkg/node/node.go
nkuba: We already got some loggers for that:
https://github.com/keep-network/keep-ecdsa/blob/3e4a0790b3f92453fd10ca950b480860a398fe78/pkg/node/params_pool.go#L45
https://github.com/keep-network/keep-ecdsa/blob/3e4a0790b3f92453fd10ca950b480860a398fe78/pkg/node/params_pool.go#L59-L63

solidity/integration/stress_test.js
pdyraga: Yep, good point. It can simplify things. Please see https://github.com/keep-network/keep-ecdsa/pull/357/commits/2751f092af5509cec187a83331ec2ce8b6749d26

solidity/integration/stress_test.js
pdyraga: Right! Done in https://github.com/keep-network/keep-ecdsa/pull/357/commits/650a589fc64e9d984bfbac0eeeee1ca1f37fe1fb

pkg/node/node.go
pdyraga: Oh, good point. In this case, I think we should drop those `Info` logs from `node.go` altogether. What do you think?

nkuba: I couldn't agree more!

pdyraga: Done in https://github.com/keep-network/keep-ecdsa/pull/357/commits/e8b95aea03fd3e029a5ee5120110f90f9dfbdd5f

pkg/node/node.go
pdyraga: Done in https://github.com/keep-network/keep-ecdsa/pull/357/commits/e8b95aea03fd3e029a5ee5120110f90f9dfbdd5f

pdyraga:
Smoke test:
<details>
➜  solidity git:(all-for-one) ✗ truffle exec integration/smoke_test.js --network local
Using network 'local'.

Browserslist: caniuse-lite is outdated. Please run next command `npm update`
selecting a keep factory...
validate registered member candidates...
operator [0x65EA55c1f10491038425725dC00dFFEAb2A1e28A] is registered for application [0x4f7C771Ab173bEc2BbE980497111866383a21172]
operator [0x524f2E0176350d950fA630D9A5a59A0a190DAf48] is registered for application [0x4f7C771Ab173bEc2BbE980497111866383a21172]
operator [0x3365D0Ed0e526D3B1d8b417fc0fdE5b1cEF2f416] is registered for application [0x4f7C771Ab173bEc2BbE980497111866383a21172]
operator [0x7020A5556Ba1ce5f92c81063a13d33512cf1305c] is registered for application [0x4f7C771Ab173bEc2BbE980497111866383a21172]
opening a new keep for owner [0x146748a2b46B99ee1470b587bc9812Ea59b79597]...
open new keep fee: [11138830000000000]
new keep opened with address: [0xA343fAf4f55775BaC0C81E1e6983ba5468553fa2] and members: [0x7020A5556Ba1ce5f92c81063a13d33512cf1305c,0x524f2E0176350d950fA630D9A5a59A0a190DAf48,0x3365D0Ed0e526D3B1d8b417fc0fdE5b1cEF2f416]
get public key...
public key generated for keep: [0x5a8f0d4bcaa91d513ff59f438ec46243db0c3187e6bb15401a562917bbe745f913777a836b3434e4e88d8ea008ceba63147feb84f07afe53008e5ea2bc81c83c]
request signature...
received valid signature:
r: [0xe2d9a9f99cb663fcce7049af0d65f8ea416be3af27639999f788cd4c78151990]
s: [0x761f3a403609e0edecfdde7aa478acd9fc38eaa81bda9e5a7f901d4cc41645fa]
recoveryID: [0]

wait for new relay entry generation...
get current group selection seed...
group selection seed successfully updated by the random beacon; seed = [74040203143020520661433123298497285997885898113360094613573902861888878456333]
</details>

Stress test:
<details>
➜  solidity git:(all-for-one) ✗ truffle exec integration/stress_test.js --network local
Using network 'local'.

opening a new keep at [4/7/2020, 11:49:57 AM]...
opening a new keep at [4/7/2020, 11:49:57 AM]...
opening a new keep at [4/7/2020, 11:49:57 AM]...
opening a new keep at [4/7/2020, 11:49:57 AM]...
opening a new keep at [4/7/2020, 11:49:57 AM]...
new keep created: [0xd6960991d47DED023d3763229c9A0285F6307B54] at [4/7/2020, 11:50:05 AM]
new keep created: [0xC60f78791450F121428A3DF81bf48613214b1fD4] at [4/7/2020, 11:50:05 AM]
new keep created: [0x0fA306472a627FF71B1E66E649a53306d9F66Cf4] at [4/7/2020, 11:50:05 AM]
new keep created: [0x1379065CeC7156dE83b53073C22b7d0Ec64D0918] at [4/7/2020, 11:50:05 AM]
new keep created: [0xC0Daa772623C896692714761C92bBdbaDC145bDE] at [4/7/2020, 11:50:05 AM]
opening a new keep at [4/7/2020, 11:50:35 AM]...
opening a new keep at [4/7/2020, 11:50:35 AM]...
opening a new keep at [4/7/2020, 11:50:35 AM]...
opening a new keep at [4/7/2020, 11:50:35 AM]...
opening a new keep at [4/7/2020, 11:50:35 AM]...
public key generated for keep [0xC60f78791450F121428A3DF81bf48613214b1fD4] at [4/7/2020, 11:51:00 AM]: [0x13a5fd5aefab29e8e806bc2d9f53cc28c33a63d1019106be2634ed0bb6075a75edcdc8b2b48a0f245fe269d64086b0e898c65a415c2f89e5cd76166075a8a7e7]
generated [1] public keys so far
public key generated for keep [0x1379065CeC7156dE83b53073C22b7d0Ec64D0918] at [4/7/2020, 11:51:00 AM]: [0x8d6cc270d97f7dbb8ac8f94348691962d0135712cb65164582b1edf5ddeb65ccf7ec3ac6453630d600fe1829954729fee8c640694daedeb242056d377cc525da]
generated [2] public keys so far
public key generated for keep [0xC0Daa772623C896692714761C92bBdbaDC145bDE] at [4/7/2020, 11:51:00 AM]: [0xd69116875781eaff84eff7ca313491dfa2ddf49ac72aeeeadf91ad52ddc9ac0d58d1c5c8bb01e0b50528bb8d268c2d303115a8305bcaf50e6558bb5cb95270fc]
generated [3] public keys so far
public key generated for keep [0x0fA306472a627FF71B1E66E649a53306d9F66Cf4] at [4/7/2020, 11:51:00 AM]: [0x15b1a5e133838d35810b7c0164b8f1bbda92610e5292d4ff30ffcd01a12aa3356363264d93df629448ed71aab0f62aca8817b8ac606c6ecf7b387cf0b23e1c62]
generated [4] public keys so far
public key generated for keep [0xd6960991d47DED023d3763229c9A0285F6307B54] at [4/7/2020, 11:51:00 AM]: [0x6566fdc27faad709340afa534e4e705d9f987d98207dac6ee1780034a49ba88e703ec9404b4181a17fa1a2668f3bbe6e16f1d69977532fbc75949d72b1d3b11f]
generated [5] public keys so far
new keep created: [0x77313E4704359332897527D5FD28020D35c6c661] at [4/7/2020, 11:51:01 AM]
new keep created: [0x00a08aA7601218ed1cf8ac033489e794ccc395C3] at [4/7/2020, 11:51:01 AM]
new keep created: [0xb836436477d013B1cdAcC57E55ba3d3fA0FF2110] at [4/7/2020, 11:51:01 AM]
new keep created: [0x579841cc2CCc27D870e19FdeC2710C191c101769] at [4/7/2020, 11:51:01 AM]
new keep created: [0xd0ee8d9966F34FA33658796999f08480C3Ecb198] at [4/7/2020, 11:51:01 AM]
opening a new keep at [4/7/2020, 11:51:31 AM]...
opening a new keep at [4/7/2020, 11:51:31 AM]...
opening a new keep at [4/7/2020, 11:51:31 AM]...
opening a new keep at [4/7/2020, 11:51:31 AM]...
opening a new keep at [4/7/2020, 11:51:31 AM]...
public key generated for keep [0x77313E4704359332897527D5FD28020D35c6c661] at [4/7/2020, 11:51:45 AM]: [0x8d8fb459be439cad4a6c4039f7ae8bb8441cf6a27cfcf042a13b6d13a49d57ee9af5bdf8b2635707cca8e4213f9ddcfc7556e76ccc0190869fc6e0e87006968d]
generated [6] public keys so far
public key generated for keep [0xb836436477d013B1cdAcC57E55ba3d3fA0FF2110] at [4/7/2020, 11:51:45 AM]: [0x6309f2d4b9b333e2062c6ef930c92252e9dbc4a38cd200a68cdda8b58420c2c803f593a56b2c67456e07dcff9bf25f5695253ec136d9edc88b649e957f827575]
generated [7] public keys so far
public key generated for keep [0xd0ee8d9966F34FA33658796999f08480C3Ecb198] at [4/7/2020, 11:51:45 AM]: [0x52a4e55e2a35a2f7f43194978c6c328fcd37b60c5d7978c13a145f7b9125873e0d9a0c1f33ecd3f1e2f34c94e2fb3fbba4dc476527b41468b122f10a73e1bac7]
generated [8] public keys so far
public key generated for keep [0x00a08aA7601218ed1cf8ac033489e794ccc395C3] at [4/7/2020, 11:51:45 AM]: [0x97ddddeef6a7c33682cbe360a20c0725ec915607d47599fb9ed4355ce350f2ba138b0ed6eb7b55daa98d5ea7f041ca5b88adb096303dab84f86bfbb9117d80fc]
generated [9] public keys so far
public key generated for keep [0x579841cc2CCc27D870e19FdeC2710C191c101769] at [4/7/2020, 11:51:45 AM]: [0xfc3fc92a77bedb843e36bf6da35408f9b9b7592f140ac7f5948670f97a0d8637800ecc686543001a54598e688ad0dfe806a200f91753f44a83f87421b4156040]
generated [10] public keys so far
new keep created: [0xc63cab430eFF6ad001cdFdA24Ee31D75224C5e43] at [4/7/2020, 11:51:45 AM]
new keep created: [0x0c7f2B8D94bD910C38F59d4858615EE56317F097] at [4/7/2020, 11:51:45 AM]
new keep created: [0x1e28400f655BBcB7870AB5dBdAE8C06cd1d9238c] at [4/7/2020, 11:51:45 AM]
new keep created: [0xC09c135978E35e3Cd828b4d94a49b5e33AA06665] at [4/7/2020, 11:51:45 AM]
new keep created: [0x7Ad27FfDe04940DeD6f1dbbdAD350b9f87f28330] at [4/7/2020, 11:51:45 AM]
opening a new keep at [4/7/2020, 11:52:15 AM]...
opening a new keep at [4/7/2020, 11:52:15 AM]...
opening a new keep at [4/7/2020, 11:52:15 AM]...
opening a new keep at [4/7/2020, 11:52:15 AM]...
opening a new keep at [4/7/2020, 11:52:15 AM]...
public key generated for keep [0x7Ad27FfDe04940DeD6f1dbbdAD350b9f87f28330] at [4/7/2020, 11:52:30 AM]: [0x6dd5b16271f91ce6a66a5ccad1ec097aaa5810b15e2f43499e671e84a661ab2e92487edf82d8757dde82f527524b4198c86924679bdee3129dc819107a4bf478]
generated [11] public keys so far
public key generated for keep [0x0c7f2B8D94bD910C38F59d4858615EE56317F097] at [4/7/2020, 11:52:30 AM]: [0xfe69fe0fc106b7d9fe0089a56351bc3fc59d0d97e3de5ecac86850252eb9c0caa14da6d4d73f4aa300abe9a005b4964c80799dcbbd389e52cdaa7ba126513c90]
generated [12] public keys so far
public key generated for keep [0x1e28400f655BBcB7870AB5dBdAE8C06cd1d9238c] at [4/7/2020, 11:52:30 AM]: [0xc79c485ba007c6c8e9fa59b49b53560ecfd9db57c4b22109cdea44f04b4a96aec6b9c71dfc8fc7a07fee8f03f360b2b5e53f09bf43e3952b687d8502a44fff4f]
generated [13] public keys so far
public key generated for keep [0xC09c135978E35e3Cd828b4d94a49b5e33AA06665] at [4/7/2020, 11:52:30 AM]: [0x1e6b67d40e8bd856efeb4ee8191d10474ae1fe245a5db2d2d2fbfac7c649f7eaf12ba19dc40acb3654500d4ae55750c4ad7061e74d79790fd1881abdb2cd4433]
generated [14] public keys so far
public key generated for keep [0xc63cab430eFF6ad001cdFdA24Ee31D75224C5e43] at [4/7/2020, 11:52:30 AM]: [0x7f3a7d37d78889321e7bc647db15e7558752f1e63668c960e190bbc7d75c88b03453d52168ade94606433993ff8d1094dc242df694e2c6618eb93b02a69b5fc3]
generated [15] public keys so far
new keep created: [0x4Db295bF483e3c8023aCE475EC10dADa3061d0c9] at [4/7/2020, 11:52:30 AM]
new keep created: [0x83Cf8DA7319A903711Bf54F610DB4269209B29E3] at [4/7/2020, 11:52:30 AM]
new keep created: [0x55C69Ca0c773Ed0b397254189d8df27Db6BEeddb] at [4/7/2020, 11:52:30 AM]
new keep created: [0xF387a84e8886857345DA7E11ca719bD2a72762c6] at [4/7/2020, 11:52:30 AM]
new keep created: [0xa82b8780e0EF3396254Dc13DAdBd232Df5f86D2d] at [4/7/2020, 11:52:30 AM]
opening a new keep at [4/7/2020, 11:53:00 AM]...
opening a new keep at [4/7/2020, 11:53:00 AM]...
opening a new keep at [4/7/2020, 11:53:00 AM]...
opening a new keep at [4/7/2020, 11:53:00 AM]...
opening a new keep at [4/7/2020, 11:53:00 AM]...
public key generated for keep [0x4Db295bF483e3c8023aCE475EC10dADa3061d0c9] at [4/7/2020, 11:53:11 AM]: [0x64ebe79f19e2ef2f0a3229a9f572123132ed17e922ae4ee36b415af51e62cda5c8a91b04061e74c1c76bbb79a3917daf8ecc7afb1286b388662f445b3344c9a0]
generated [16] public keys so far
public key generated for keep [0x83Cf8DA7319A903711Bf54F610DB4269209B29E3] at [4/7/2020, 11:53:11 AM]: [0xed361a63326070751dbf945aab97efc7a02fa5800c489f819875f4a93fd4d10eb965a972890502e3402f12b87cb172c992c1d15f55f250fba97892ec28f3d536]
generated [17] public keys so far
public key generated for keep [0xa82b8780e0EF3396254Dc13DAdBd232Df5f86D2d] at [4/7/2020, 11:53:11 AM]: [0x2d86950a68bd3de526eef1e5faa43d0ef20fff7565f9dc0aa1fd7282975fa2a98ebc3bf6d34302a761b99f9ccefa0c7e8d1d5113b22b86754850ef01717d2604]
generated [18] public keys so far
public key generated for keep [0x55C69Ca0c773Ed0b397254189d8df27Db6BEeddb] at [4/7/2020, 11:53:11 AM]: [0x2e3d14ca433d047ddf61867546343083a42c4db1f2d50fa2ba87e38d721c9ace49def3bef5aac411d7d8ddbc5d35e9503bd0d35e42c0e15460c1d42312d96c2e]
generated [19] public keys so far
public key generated for keep [0xF387a84e8886857345DA7E11ca719bD2a72762c6] at [4/7/2020, 11:53:11 AM]: [0x8ab6d5748cd4a8e7166bc7b999d94de05b1d20f2648759c61a7a6ee74ef8e04bf5e88e45f0cef8ee467369ad6a9191a9fa1750b8e769cfcea3d55f280ac9f6dd]
generated [20] public keys so far
new keep created: [0x28C26F1d169550Fac9933B5fe882E30DC087243A] at [4/7/2020, 11:53:11 AM]
new keep created: [0x2D7675Fd63cA51630783e0Ab161E76445AE807e1] at [4/7/2020, 11:53:11 AM]
new keep created: [0x69388932C052753542eBe96E74cDCA70f0Fd1e8D] at [4/7/2020, 11:53:11 AM]
new keep created: [0xCF856aB80395c64513FFAd3e2D1324b55BC4fFf1] at [4/7/2020, 11:53:11 AM]
new keep created: [0x39b30B4ef64c92600B89643517F0aF2329b5d147] at [4/7/2020, 11:53:11 AM]
opening a new keep at [4/7/2020, 11:53:41 AM]...
opening a new keep at [4/7/2020, 11:53:41 AM]...
opening a new keep at [4/7/2020, 11:53:41 AM]...
opening a new keep at [4/7/2020, 11:53:41 AM]...
opening a new keep at [4/7/2020, 11:53:41 AM]...
public key generated for keep [0x69388932C052753542eBe96E74cDCA70f0Fd1e8D] at [4/7/2020, 11:53:45 AM]: [0xf5b19c2046b49b806f4249e2c561eb106723c3a8b07774a6f2327c9b0964a81eab855a75aec987ba7f7f4a259aa4ab330759cdd492eb55af55337d32799282a7]
generated [21] public keys so far
public key generated for keep [0xCF856aB80395c64513FFAd3e2D1324b55BC4fFf1] at [4/7/2020, 11:53:45 AM]: [0xc1ce5c2a0982592fbc3a6589e1c06f0d58cd56cff2ce37ce59076a40ae5f04dc83aa346715c7445ba477301309091360ec5ce049cf8f80038a197d6a66ca2d4c]
generated [22] public keys so far
public key generated for keep [0x39b30B4ef64c92600B89643517F0aF2329b5d147] at [4/7/2020, 11:53:45 AM]: [0xdf2a5746657893bb2f49262712c94a48bfc588470186383813f2861edf0b5df16d19cb438efc58979c07a72b295a7fca3c25c8b21a231e5068faa3d1f775c89d]
generated [23] public keys so far
public key generated for keep [0x2D7675Fd63cA51630783e0Ab161E76445AE807e1] at [4/7/2020, 11:53:45 AM]: [0x4700f5b3cdb2687bc8d6273e8d4a830ccd42470b9eeda7fd1749df9a78d5ddce9d1c1446102d37878442f6dca3ae24bc371bcd2c9d87410f76547637ba9ca547]
generated [24] public keys so far
public key generated for keep [0x28C26F1d169550Fac9933B5fe882E30DC087243A] at [4/7/2020, 11:53:45 AM]: [0xaa835e6a47ba870c6bedf81190387dfb91588ccb69ac6bbd43a5eaf9aea4796d02ff55a7c1a2baab5a91bad6d62cda0d6be99315db1c2239b7829e1c1faf7d2a]
generated [25] public keys so far
new keep created: [0x711d8fCf50953308D429076E3d4954f05E9496BF] at [4/7/2020, 11:53:45 AM]
new keep created: [0x477F86e19799890c9ec34A95aE166b4388bD1Fd3] at [4/7/2020, 11:53:45 AM]
new keep created: [0x281e7264602879f7d9Fff6801eC1dc2DD215E19F] at [4/7/2020, 11:53:45 AM]
new keep created: [0x71d189603E0d76cB1ebd8552Ea7558D6357d39A8] at [4/7/2020, 11:53:45 AM]
new keep created: [0xb06F3134AaccEfc589b25c7ABB0f9BC26c4ef472] at [4/7/2020, 11:53:45 AM]
opening a new keep at [4/7/2020, 11:54:15 AM]...
opening a new keep at [4/7/2020, 11:54:15 AM]...
opening a new keep at [4/7/2020, 11:54:15 AM]...
opening a new keep at [4/7/2020, 11:54:15 AM]...
opening a new keep at [4/7/2020, 11:54:15 AM]...
public key generated for keep [0x477F86e19799890c9ec34A95aE166b4388bD1Fd3] at [4/7/2020, 11:54:43 AM]: [0x1aa1a5658c5fd3222e70f2d505d7c994e5847917510a20b00f6325014d52622d0feb2cdedea1a13ab23c9aae2442d31167fac9ac9507171b7cfb88cc270135f8]
generated [26] public keys so far
new keep created: [0xc79af0b2357CC0039778e4E9e4277ec5DE3656Df] at [4/7/2020, 11:54:44 AM]
new keep created: [0xf6D619418D53804a2e8288f0e72c864F35Fd9144] at [4/7/2020, 11:54:44 AM]
new keep created: [0x670024A964bdD0A81b7d29d3D9fD27A287AE68c1] at [4/7/2020, 11:54:44 AM]
new keep created: [0x67C2b9B252EC3aDb6FfF402a0A0bF302Df360175] at [4/7/2020, 11:54:44 AM]
new keep created: [0x31ECe9d6BA2DffEca7957ddd73d42Cd167e371aa] at [4/7/2020, 11:54:44 AM]
public key generated for keep [0x71d189603E0d76cB1ebd8552Ea7558D6357d39A8] at [4/7/2020, 11:55:03 AM]: [0x6568d8ba255c1191bf0ac1ff25db525f0a5316a5db55ee29a540e4652fd7e489883415a17116fb6bc9964c6cec1201f2c2e2b16e2664b730c6ff5f1fee979269]
generated [27] public keys so far
opening a new keep at [4/7/2020, 11:55:14 AM]...
opening a new keep at [4/7/2020, 11:55:14 AM]...
opening a new keep at [4/7/2020, 11:55:14 AM]...
opening a new keep at [4/7/2020, 11:55:14 AM]...
opening a new keep at [4/7/2020, 11:55:14 AM]...
public key generated for keep [0x711d8fCf50953308D429076E3d4954f05E9496BF] at [4/7/2020, 11:55:26 AM]: [0xf4e1cba281e2cd24c48eaf86c835ac147d93e78d020a883626f7249e37b3b98d01617466bb48144de91445da64594a9ecaee95094c01c78a7c4d75c5307d6b22]
generated [28] public keys so far
new keep created: [0x64C2Cbf64ED5C6Be8C01f5055C18F24912960345] at [4/7/2020, 11:55:26 AM]
new keep created: [0x30f6939faCEdcdD232bc3A00B0EBD11dd2E9a6E0] at [4/7/2020, 11:55:26 AM]
new keep created: [0xEBafE69B118FfCDe3F18A503EAC352B185EB0cc6] at [4/7/2020, 11:55:26 AM]
new keep created: [0xA4f51Ed3fb981d2a4F81fFDB6447ef96FC5c4742] at [4/7/2020, 11:55:26 AM]
new keep created: [0x00FA59cBCf818dd882d3b85CE84723C730F06CfF] at [4/7/2020, 11:55:26 AM]
opening a new keep at [4/7/2020, 11:55:56 AM]...
opening a new keep at [4/7/2020, 11:55:56 AM]...
opening a new keep at [4/7/2020, 11:55:56 AM]...
opening a new keep at [4/7/2020, 11:55:56 AM]...
opening a new keep at [4/7/2020, 11:55:56 AM]...
new keep created: [0xccCA032019724D5cC6893FeB4CBA028639521fFB] at [4/7/2020, 11:56:34 AM]
new keep created: [0xEcA549766b521612Aeb95E7E579e4f5faE8ae863] at [4/7/2020, 11:56:34 AM]
new keep created: [0x6aff15B43D651ebC6882B83388EcDb6411e5Fa67] at [4/7/2020, 11:56:34 AM]
new keep created: [0x28ffAb1266741C0d5442A8B8B2b3754a313f236e] at [4/7/2020, 11:56:34 AM]
new keep created: [0xd5b996BBC5e994c48Ba06C182aCeC79543A0B91A] at [4/7/2020, 11:56:34 AM]
opening a new keep at [4/7/2020, 11:57:04 AM]...
opening a new keep at [4/7/2020, 11:57:04 AM]...
opening a new keep at [4/7/2020, 11:57:04 AM]...
opening a new keep at [4/7/2020, 11:57:04 AM]...
opening a new keep at [4/7/2020, 11:57:04 AM]...
public key generated for keep [0xb06F3134AaccEfc589b25c7ABB0f9BC26c4ef472] at [4/7/2020, 11:58:05 AM]: [0xb74bcc4516398a248dc2f4d3422940e59750e3f15b5fdc26ca67c12c9435961f83a20b60466eba8c20e9e175ed185eceaf1698d2fd96b371ce33096e04d62d56]
generated [29] public keys so far
new keep created: [0xbf2cA419C81FcF6fAd6C7466133E030222D54b2A] at [4/7/2020, 11:58:05 AM]
new keep created: [0xa01E882dF86050691F09c3D2F650E234729aceA6] at [4/7/2020, 11:58:05 AM]
new keep created: [0xEdbb1A63b38dBaFfEBD2106b89d3ab59ECB81fAA] at [4/7/2020, 11:58:05 AM]
new keep created: [0x1c71F0742753Bf58fE8F8333C03e2b1294F67660] at [4/7/2020, 11:58:05 AM]
new keep created: [0xae3B63128D71eb90613fb2D862415031B7E01D24] at [4/7/2020, 11:58:05 AM]
public key generated for keep [0x281e7264602879f7d9Fff6801eC1dc2DD215E19F] at [4/7/2020, 11:58:53 AM]: [0xe29b30d24bd840cecb65e964413e48295f2974d79bc2f60bdab2d65892502df93c30af399c32c27ca0961fab08ebdb2381ba07b3683ad53f3498bd2a43bda255]
generated [30] public keys so far
public key generated for keep [0xf6D619418D53804a2e8288f0e72c864F35Fd9144] at [4/7/2020, 12:00:48 PM]: [0xbb77d4d55b5e0305d3d04f73080e405ea3aab0bbbdc7dddff95494120366be283e9e9cbe9dd33bcd20aa34eb1ebf19ec783891e7e81eeb2822ca3b6dbb35e9bf]
generated [31] public keys so far
public key generated for keep [0xc79af0b2357CC0039778e4E9e4277ec5DE3656Df] at [4/7/2020, 12:01:15 PM]: [0x6264191ff2ebf45b57eca9b72adb8cc8a8ae46741c2773f94a99f4bcb5639c8650210e30cac74495ce16e865737ea22d908c7ea072f50cff0e8d204c938510a3]
generated [32] public keys so far
public key generated for keep [0x00FA59cBCf818dd882d3b85CE84723C730F06CfF] at [4/7/2020, 12:01:36 PM]: [0xa980f2698a98f128181a3c474f1120be57d97c62387d5e6fb7d2f18489c3a1c5687630b132b522bceb53b04dd9e7d7e78a2b1889102e94e8f55c06a1640ffd1c]
generated [33] public keys so far
public key generated for keep [0x67C2b9B252EC3aDb6FfF402a0A0bF302Df360175] at [4/7/2020, 12:02:30 PM]: [0x0da4024691f2b8ce7d85033231ce32187620dbd00dbf05696a78c327a148045a49ef8b297c496114db5037df9e1d0f69257b0d09eb9758f4615442969c9bd40e]
generated [34] public keys so far
public key generated for keep [0x670024A964bdD0A81b7d29d3D9fD27A287AE68c1] at [4/7/2020, 12:02:30 PM]: [0xadff04cf1edcb5ae455b7863261a66ffb81d9fdc407abfd6a394cc37c71d19d37f4dc48a8de771ce6912297e4d2b516156ec5a0541a25f3eddf496bc86c1da1c]
generated [35] public keys so far
public key generated for keep [0x31ECe9d6BA2DffEca7957ddd73d42Cd167e371aa] at [4/7/2020, 12:03:53 PM]: [0xd263c73cea931742d6c007a2f1163da27380efc6780738df4949e3bd4b83cf5b737eed8a48c6c2c33803fcd4e8997802bd668684b250aeb2831298516847f099]
generated [36] public keys so far
public key generated for keep [0x30f6939faCEdcdD232bc3A00B0EBD11dd2E9a6E0] at [4/7/2020, 12:04:37 PM]: [0x51875ad3d0bc105ee54d3a1bc57df5aa5f8a0c7fef8db496c2e8521e40b4d6609aa2c8384df5eaf72bde8c30d803a3fd652db390407885273db54cba2a373428]
generated [37] public keys so far
public key generated for keep [0xA4f51Ed3fb981d2a4F81fFDB6447ef96FC5c4742] at [4/7/2020, 12:05:10 PM]: [0xdbd2cdd266295402ec18461a117c009d3ee48b096ed98f1185566c6682b8d78b6cd00e6a89aeeaf756654d4c6200f16dfa3e092430860c1b472467bcc45faecd]
generated [38] public keys so far
public key generated for keep [0x64C2Cbf64ED5C6Be8C01f5055C18F24912960345] at [4/7/2020, 12:06:07 PM]: [0x567cf79d3ae53685d479df9a30ef38edda88b870ee0195a1c334b72b875b9c62c5cf40998b9ad8db1c59ff629f584fc79e3e7c6685171bc235ede42c867dc273]
generated [39] public keys so far
public key generated for keep [0xEBafE69B118FfCDe3F18A503EAC352B185EB0cc6] at [4/7/2020, 12:09:18 PM]: [0x8cbefb29e28776764c0ace9562b0c4c84b573d18326ddf6aacee26b5832b8946705d0920451c54a57695bd50f3728320c6b6ab2d0512325362a249f68ee3dbb2]
generated [40] public keys so far
public key generated for keep [0xccCA032019724D5cC6893FeB4CBA028639521fFB] at [4/7/2020, 12:10:54 PM]: [0x61676aca8071028356fb6173fbd4b4ff0fa0e627c1b2f9b2b670736d5d74b4577de50a4ff3f2d57debadccca39e95624b7f6f9e3349469549c45b0c57651c5c8]
generated [41] public keys so far
public key generated for keep [0xEcA549766b521612Aeb95E7E579e4f5faE8ae863] at [4/7/2020, 12:12:23 PM]: [0x81b01eb318d17acc4bdf42aa1237cf95ec390a8f11e6a0146d1122ead7232eb38cda9726fee4a2081f24a488335397e40edad7db584f530319dcd0e233254f4d]
generated [42] public keys so far
public key generated for keep [0x6aff15B43D651ebC6882B83388EcDb6411e5Fa67] at [4/7/2020, 12:12:23 PM]: [0xb12316a2c00fd659a0fa64d9fa579f24beae6a560eb6aebb612b21e2838c1f69d57c13b6652fd16d3709dd36e9059f02dbec402dafa700a197d24c45cacc5b69]
generated [43] public keys so far
public key generated for keep [0xd5b996BBC5e994c48Ba06C182aCeC79543A0B91A] at [4/7/2020, 12:12:45 PM]: [0x5b24a29104b84dffc49a0039abe376d074e7772fbc9be901124974de20f6db141b16a7a625f42d0b3f61429cbf6bcbcef5552ce36a0e6484fe7bd24a1d1f6bff]
generated [44] public keys so far
public key generated for keep [0x28ffAb1266741C0d5442A8B8B2b3754a313f236e] at [4/7/2020, 12:13:13 PM]: [0x86c184e11dcf1c2050c7b6217f97a8565c1bd8703e74ace8397f63c79ba9cbdb825db087ccd150761d52b2165b56fd449f51e32135df2a4ec17c49e1485fde26]
generated [45] public keys so far
public key generated for keep [0xa01E882dF86050691F09c3D2F650E234729aceA6] at [4/7/2020, 12:14:30 PM]: [0x9ddf06000fe8b6cd774ebb8f1b99dbca2988b82e612d2fc5db2fa0bed7a02b42a8f21e869e84cc220dd473c393ac0dc703aef1027f43d7029a1620a1ed022967]
generated [46] public keys so far
public key generated for keep [0xbf2cA419C81FcF6fAd6C7466133E030222D54b2A] at [4/7/2020, 12:15:09 PM]: [0xb0bccd3ea215b72c1b1e06677e54cbc8fb8a6f4dbdf9b1b8d646a24188444fc83702475cc053fd99ce97267694d9d6fb03137a035f18e9740f2c00b71b84be89]
generated [47] public keys so far
public key generated for keep [0x1c71F0742753Bf58fE8F8333C03e2b1294F67660] at [4/7/2020, 12:16:33 PM]: [0x4d2f745db768382767036819cad086e4f95f2638dcbd64f4385aff8edbeb025410bf50307a9fbf0beab299b39bcc1181c2c3137baffb6fdbfe6b9e88c67619ae]
generated [48] public keys so far
public key generated for keep [0xEdbb1A63b38dBaFfEBD2106b89d3ab59ECB81fAA] at [4/7/2020, 12:17:23 PM]: [0xcfe691fe86d6d98100752ea7b15b7f8fd2d8f7d55820fe33cb565279e1e0b0074170ef3858a6f8797ac876c205e8403fe868abb47cec9dfb0fec1ec22cd7f388]
generated [49] public keys so far
public key generated for keep [0xae3B63128D71eb90613fb2D862415031B7E01D24] at [4/7/2020, 12:17:34 PM]: [0x0f634ed6ad250870ec141315b5cf75f0c07fa2addad016bcb85e5141c7a47578f7b3a96b8e27d4ec99e9d3f807b11dba6dd13a5a67c77eb59bb9974d3a775854]
generated [50] public keys so far

</details>

PR #368, opened 2020-04-07
Lint Trap: Add Go linting, drop packaging Action, move Solidity Linting to Actions

### LINKED ISSUE #358: Configure Go linting in CI
We have precommit hooks that run the Go linter, but let's also do this in CI to catch folks who may not be running the precommit hooks.

Shadowfiend:
Still a drafty draft.

Closes #358.


Shadowfiend:
The Go side of this is a bit of a quagmire… Basically going to have to switch to doing it in docker the way Circle does it, which I was kind of hoping to avoid for a few reasons, in part because solc is monstrous and doesn't have a 0.5.17 package in its Ubuntu PPA.

Shadowfiend:
Closing this in favor of the work @michalinacienciala and @nkuba have been doing around solidifying our Actions and CI build.

PR #369, opened 2020-04-07
Check awaiting key generation on client start

### LINKED ISSUE #337: Check for in-progress key generation on client start
We need similar monitoring for a key generation as implemented for signing https://github.com/keep-network/keep-ecdsa/issues/268.

When the client got restarted or missed an event is should still be able to join ongoing key generation process.

lukasz-zimnoch:
Closes #337

It might happen the client did not receive an open keep event or the client could be restarted during the key generation process. We check on the client start if there is no key generation in progress occurring on several latest keeps (the ones on which key generation timeout is not exceeded yet)


pdyraga:
In #374 we missed adding an Info log on client start about a requested signature. As a result, we wait for confirmations but there is no information on what do we want to confirm. Can we please take the opportunity and add `logger.Info` in `checkAwaitingSignature` about the fact there is an awaiting signature?

This is how it currently looks like:
```
15:14:19.696  INFO keep-ecdsa: generating new tss pre parameters params_pool.go:45
15:14:19.696  INFO    tss-lib: generating the safe primes for the signing proofs, please wait... prepare.go:66
15:14:19.696  INFO    tss-lib: generating the Paillier modulus, please wait... prepare.go:51
15:14:19.704  INFO keep-regis: loaded [3] keeps from the local storage keeps.go:123
15:14:19.704  INFO   keep-cmd: client started start.go:162
15:14:19.717  INFO keep-ecdsa: waiting for block [10836] to confirm chain state client.go:562
```

pkg/chain/chain.go
```diff
@@ -50,6 +51,12 @@ type BondedECDSAKeepFactory interface {
 	// UpdateStatusForApplication updates the operator's status in the signers'
 	// pool for the given application.
 	UpdateStatusForApplication(application common.Address) error
+
+	// GetKeepCount returns number of keeps.
+	GetKeepCount() (*big.Int, error)
+
+	// GetKeepAtIndex returns keep at given index.
```
pdyraga, line 58: `...returns the address of the keep at the given index.` ?

pkg/chain/chain.go
```diff
@@ -112,4 +119,14 @@ type BondedECDSAKeep interface {
...
 	LatestDigest(keepAddress common.Address) ([32]byte, error)
+
+	// GetPublicKey returns keep's public key.
+	GetPublicKey(keepAddress common.Address) ([]uint8, error)
+
+	// GetPublicKey returns keep's members.
+	GetMembers(keepAddress common.Address) ([]common.Address, error)
+
+	// HasKeyGenerationTimedOut returns whether key generation
+	// has timed out for given keep.
```
pdyraga, line 130: `*the given`

pkg/chain/chain.go
```diff
@@ -112,4 +119,14 @@ type BondedECDSAKeep interface {
 
 	// LatestDigest returns the latest digest requested to be signed.
 	LatestDigest(keepAddress common.Address) ([32]byte, error)
+
+	// GetPublicKey returns keep's public key.
```
pdyraga, line 123: What is returned when there is no public key yet? Worth being clear about it here.

pkg/client/client.go
```diff
@@ -246,20 +207,212 @@ func Initialize(
...
+	}
+
+	zero := big.NewInt(0)
+	one := big.NewInt(1)
+
+	lastIndex := new(big.Int).Sub(keepCount, one)
+
+	// Iterate through keeps starting from the end.
+	for keepIndex := new(big.Int).Set(lastIndex); keepIndex.Cmp(zero) != -1; keepIndex.Sub(keepIndex, one) {
+
```
pdyraga, line 231: Can we add a debug log that we are checking keep at the given index for key generation?

pkg/client/client.go
```diff
@@ -246,20 +207,212 @@ func Initialize(
...
+					"for keep [%v]: [%v]",
+				keep.String(),
+				err,
+			)
+			continue
+		}
+
+		// If key generation timeout of current keep has been exceeded,
+		// there is no sense to continue because the next keep (created earlier)
+		// will have this timeout exceeded as well.
```
pdyraga, line 255: 👌 smart.

pkg/client/client.go
```diff
@@ -360,6 +513,12 @@ func checkAwaitingSignature(
 	}
 
 	if isAwaitingDigest {
+		logger.Infof(
+			"awaiting signature for digest [%+x] exists for keep [%s]",
```
pdyraga, line 517: Let's simplify it to:
```
"awaiting a signature from keep [%s] for digest [%+x] "
```

pdyraga:
Testing time... 

giphy

pdyraga:
- key generation for a keep starts immediately once the event is received ✅ 
- signature generation starts 12 blocks after signature generation event is received ✅ 
- when the client starts it picks ongoing single key generation request  ✅
- when the client starts it picks several ongoing key generation requests ✅ 
- when the client starts it picks single signature generation request and waits for 12 blocks before starting the protocol
- waits for 12 blocks before archiving closed keep after receiving on-chain event ✅ 
- waits for 12 blocks before archiving terminated keep after receiving on-chain event ✅ 
- detect after starting that keep is no longer active and waits for 12 blocks before archiving it ✅

Issue #364, opened 2020-04-06
Wait for enough confirmations before signing.


pdyraga:
We need to track adding a delay to wait for a few confirmations to build up on top of the request. If it gets reorged away, we'll end up with ECDSA fraud exposure. Additionally, we should track which events have been handled so that we don't handle the same event more than one time in case it gets delivered more than once (observed on Infura couple of times); see #355

Shadowfiend:
To do this right we'll want to make sure that after the given time has elapsed, the transaction still exists on the current chain.

pdyraga:
Yep, our plan is to receive the event, wait for N blocks, and then check again. We'll also need to register this event somewhere so that we don't handle more than once (#355)

nkuba:
Thoughts around the reasonable block confirmation number?

pdyraga:
I'd say `5` to be sure but this affects tBTC signing times. Maybe we need another step in the redemption flow?

Shadowfiend:
6-12 feels right… Signing window for tBTC is in hours, right? The 3-4 minutes delay here shouldn't have a huge impact?

pdyraga:
It may have an impact on user experience in dApp, that's the only one piece I am worried about.

Shadowfiend:
Ah. Let's focus on security here, and figure out how to handle this in the dApp separately I think. The right solution from a dApp perspective is probably that the signers submit the Bitcoin tx anyway, tbh.

PR #370, opened 2020-04-07
Event confirmations

nkuba:
After receiving a signing request, keep closing or keep terminating events we want to wait for 12 blocks and check that state of the contract is really as in the emitted event.

Closes: https://github.com/keep-network/keep-ecdsa/issues/364

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -445,7 +445,7 @@ func waitForEventConfirmation(
 
 	result, err := stateCheck()
 	if err != nil {
-		return false, fmt.Errorf("failed to get state confirmation: [%v]", err)
+		return false, fmt.Errorf("failed to get chain state confirmation: [%v]", err)
 	}
 
 	return result, nil
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -3,6 +3,7 @@ package client
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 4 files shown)


pkg/client/client.go
```diff
@@ -344,3 +424,29 @@ func monitorKeepTerminatedEvent(
...
+	logger.Infof("waiting for [%d] block", blockHeight)
+
+	err := ethereumChain.BlockCounter().WaitForBlockHeight(blockHeight)
+	if err != nil {
+		return false, fmt.Errorf("failed to wait for block height: [%v]", err)
+	}
+
+	result, err := stateCheck()
+	if err != nil {
+		return false, fmt.Errorf("failed to get state confirmation: [%v]", err)
```
pdyraga, line 448: `chain state confirmation` perhaps?

pkg/client/client.go
```diff
@@ -344,3 +424,29 @@ func monitorKeepTerminatedEvent(
...
+// predefined number of blocks since the start block number provided. After the
+// required block number is reached it performs a check of the chain state with
+// a provided function returning a boolean value.
+func waitForEventConfirmation(
+	ethereumChain eth.Handle,
+	startBlockNumber uint64,
+	stateCheck func() (bool, error),
+) (bool, error) {
+	blockHeight := startBlockNumber + blockConfirmations
+	logger.Infof("waiting for [%d] block", blockHeight)
```
pdyraga, line 439: It looks weird, especially that we say `... at block [x]` in other places:
`21:53:13.777  INFO keep-ecdsa: waiting for [5594] block client.go:439` 

Can we change it to `"waiting for block [%d]"`?

nkuba: https://github.com/keep-network/keep-ecdsa/pull/374/commits/a59937b1f525f62015d645cb9c746ad83441d93f

Issue #355, opened 2020-04-04
Duplicated events delivery handled as separate events


nkuba:
It happened that clients received multiplicated notifications for the same event. Clients were using Infura API:
```
2020-04-03 20:54:21.235490 INFO keep-ecdsa client.go:113: new keep [0x51D9A0847F7BCdC884Df4A0415Df848854a8E897] created with members: [[c13727ea236cf95215694eab8f9fdd7ea1eb36c3 cda72e609c99ee673fe865300dee96fc46895a25 a978fd525ff8f93c9ecae731578399ed2d443e1a]]
2020-04-03 20:54:28.098227 INFO keep-ecdsa client.go:113: new keep [0x51D9A0847F7BCdC884Df4A0415Df848854a8E897] created with members: [[c13727ea236cf95215694eab8f9fdd7ea1eb36c3 cda72e609c99ee673fe865300dee96fc46895a25 a978fd525ff8f93c9ecae731578399ed2d443e1a]]
2020-04-03 20:54:28.132570 INFO keep-ecdsa client.go:113: new keep [0x51D9A0847F7BCdC884Df4A0415Df848854a8E897] created with members: [[c13727ea236cf95215694eab8f9fdd7ea1eb36c3 cda72e609c99ee673fe865300dee96fc46895a25 a978fd525ff8f93c9ecae731578399ed2d443e1a]]
```

It caused three separate signer generation processes invoked for the same keep by each of the above clients.

It worked fine on other clients that were using Alchemy API. These logged just one entry.

nkuba:
Log from the problematic client shows that there were some single-delivered messages and some multi-delivered messages:
```
2020-04-03 20:54:11.291844 INFO keep-ecdsa client.go:113: new keep [0x413B02d553A30E92fC6707d77f4c666eBf294EB7] created with members: [[c13727ea236cf95215694eab8f9fdd7ea1eb36c3 a978fd525ff8f93c9ecae731578399ed2d443e1a 22aad749b44bf70ec34288fc16a0fc12a56287c9]]
2020-04-03 20:54:11.292169 INFO keep-ecdsa client.go:121: member [0x22aAd749B44Bf70eC34288Fc16a0fC12A56287c9] is starting signer generation for keep [0x413B02d553A30E92fC6707d77f4c666eBf294EB7]...
[...]

PR #372, opened 2020-04-07
Duplicated events handling

nkuba:
It may happen that we receive the same event multiple times (https://github.com/keep-network/keep-ecdsa/issues/355). We don't want to kick off duplicated processes for such events. 

In this PR we introduce a way to track requests and start only single processes for them.

We register an event in the struct after receiving it. If it's already there we will log an error and return from the process. If it's a unique event we will start the process. While we return from the process (no matter if with failure or success) we will remove the event from the tracking struct.

Closes: https://github.com/keep-network/keep-ecdsa/issues/355


pdyraga:
@nkuba Can you please merge `master` or rebase?

pkg/client/events_tracking.go
```diff
@@ -0,0 +1,92 @@
...
+	"fmt"
+	"sync"
+
+	"github.com/ethereum/go-ethereum/common"
+)
+
+// requestedSignersTrack is used to track signers generation started after keep
+// creation event is received. It is used to ensure that the process execution
+// is not duplicated, e.g. when the client receives the same event multiple times.
+// When event is received, it should be noted in in this struct. When the signer
```
pdyraga, line 14: `in in`

pkg/client/events_tracking.go
```diff
@@ -0,0 +1,92 @@
...
+	rst.mutex.Lock()
+	defer rst.mutex.Unlock()
+
+	delete(rst.data, keepAddress.String())
+}
+
+// requestedSignaturesTrack is used to track signature calculation started after
+// signature request event is received. It is used to ensure that the process execution
+// is not duplicated, e.g. when the client receives the same event multiple times.
+// When event is received, it should be noted in in this struct. When the signature
```
pdyraga, line 48: `in in`

pkg/client/events_tracking.go
```diff
@@ -0,0 +1,92 @@
...
+// is not duplicated, e.g. when the client receives the same event multiple times.
+// When event is received, it should be noted in in this struct. When the signer
+// generation process completes (no matter if it succeeded or failed), it should
+// be removed from this struct.
+type requestedSignersTrack struct {
+	data  map[string]bool // <keep, bool>
+	mutex *sync.Mutex
+}
+
+func (rst *requestedSignersTrack) add(keepAddress common.Address) error {
```
pdyraga, line 22: Instead of returning an `error`, what do you think about returning a `bool`?

Two advantages I can think of:
- `failed to register signature requested event` does not really bring any value; we'd be able to do just:
    ```
     if !requestedSigners.add(event.KaepAddress) {
       logger.Errorf("signatur generation already requested for keep ...")
     }
    ```

- tests could be simpler as we don't have to check the internal state; so instead of:

    ```
    if !rs.data[keepAddress1.String()] {
		t.Errorf(
			"unexpected value for keep [%s]\nexpected: [%v]\nactual:   [%v]",
			keepAddress1.String(),
			true,
			rs.data[keepAddress1.String()],
		)
	}
    ```

    we just assert `bool` value returned by this function.

nkuba: Sure, good idea. Error is not definitely not needed here.

pkg/client/events_tracking_test.go
```diff
@@ -0,0 +1,361 @@
...
+	keepAddress1 := common.BytesToAddress([]byte{1})
+	keepAddress2 := common.BytesToAddress([]byte{2})
+
+	rs := &requestedSignersTrack{
+		data:  make(map[string]bool),
+		mutex: &sync.Mutex{},
+	}
+
+	if !rs.add(keepAddress1) {
+		t.Error("unexpected failure")
```
pdyraga, line 22: Maybe `unexpected result` or even better `event wasn't emitted before` or something similar.

pkg/client/events_tracking_test.go
```diff
@@ -0,0 +1,361 @@
...
+	}
+
+	if ok := rs.data[keepAddress1.String()]; !ok {
+		t.Errorf(
+			"unexpected value for keep [%s]\nexpected: [%v]\nactual:   [%v]",
+			keepAddress1.String(),
+			true,
+			rs.data[keepAddress1.String()],
+		)
+	}
```
pdyraga, line 32: Do we want to check it? I think we are going to deep inside the implementation, we should limit ourselves to the result returned by `add`.

pkg/client/events_tracking_test.go
```diff
@@ -0,0 +1,361 @@
...
+	if ok := rs.add(keepAddress1); !ok {
+		t.Error("unexpected failure")
+	}
+
+	if ok := rs.add(keepAddress1); ok {
+		t.Errorf(
+			"unexpected result\nexpected: [%v]\nactual:   [%v]",
+			false,
+			ok,
+		)
```
pdyraga, line 65: In the previous assertion we don't print expected vs actual, I think we should use the same pattern here and say e.g. `signature already requested` or something similar.

pkg/client/events_tracking_test.go
nkuba: I'm 50-50. How do we ensure that `add` actually introduced changes that we expected? Maybe we could check expected map similiary as we do fo signatures?
```go
expectedData := map[string]bool{
	keepAddress1.String(): true,
	keepAddress2.String(): true,
}

if ok := !rs.add(keepAddress1); !ok {
	t.Error("unexpected failure")
}

if ok := rs.add(keepAddress2); !ok {
	t.Error("unexpected failure")
}

if !reflect.DeepEqual(expectedData, rs.data) {
	t.Errorf(
		"unexpected map content\nexpected: [%v]\nactual:   [%v]",
		expectedData,
		rs.data,
	)
}
```

pkg/client/events_tracking_test.go
nkuba: For consistency compare whole data map? (https://github.com/keep-network/keep-ecdsa/pull/372#discussion_r405451898)

pkg/client/events_tracking_test.go
pdyraga: I think we shouldn't care about what's happening inside in test but what are the outcomes and how the API behaves. All we care is what `add` returns. So for signing:
```
func TestRequestedSignersTrackAdd(t *testing.T) {
	keepAddress1 := common.BytesToAddress([]byte{1})
	keepAddress2 := common.BytesToAddress([]byte{2})

	rs := &requestedSignersTrack{
		data:  make(map[string]bool),
		mutex: &sync.Mutex{},
	}

	if !rs.add(keepAddress1) {
		t.Error("event wasn't emitted before; should be added successfully")
	}

	if !rs.add(keepAddress2) {
		t.Error("event wasn't emitted before; should be added successfully")
	}
}

func TestRequestedSignersTrackAdd_Duplicate(t *testing.T) {
	keepAddress := common.BytesToAddress([]byte{1})

	rs := &requestedSignersTrack{
		data:  make(map[string]bool),
		mutex: &sync.Mutex{},
	}

	if !rs.add(keepAddress) {
		t.Error("event wasn't emitted before; should be added successfully")
	}

	if rs.add(keepAddress) {
		t.Error("event was emitted before; it should not be added")
	}
}

func TestRequestedSignersTrackRemove(t *testing.T) {
	keepAddress := common.BytesToAddress([]byte{1})

	rs := &requestedSignersTrack{
		data:  make(map[string]bool),
		mutex: &sync.Mutex{},
	}

	if !rs.add(keepAddress) {
		t.Error("event wasn't emitted before; should be added successfully")
	}

	rs.remove(keepAddress)

	if rs.add(keepAddress) {
		t.Error("event was removed from tracking; should be added successfully")
	}
}

func TestRequestedSignersTrackRemove_WhenEmpty(t *testing.T) {
	keepAddress := common.BytesToAddress([]byte{1})

	rs := &requestedSignersTrack{
		data:  make(map[string]bool),
		mutex: &sync.Mutex{},
	}

	rs.remove(keepAddress)

	if !rs.add(keepAddress) {
		t.Error("event wasn't emitted before; should be added successfully")
	}
}
```

nkuba: Okey!

pdyraga: OK, let me update the rest of the tests as I am already 50% there.

pdyraga:
@nkuba @lukasz-zimnoch Tests updated as discussed, please take a look.

Issue #371, opened 2020-04-07
Wait for enough confirmations on client start before archiving and signing


pdyraga:
In #364 we introduced protection against signing data in a forked chain to avoid the risk of producing a signature that can be used later as a fraud-proof. A similar mechanism of waiting for N confirmations should be implemented when we start the client and decide to archive keep data or sign a digest.

PR #374, opened 2020-04-08
Wait for enough chain confirmations on client start for archiving and signing

nkuba:
In this PR we implement chain state confirmations checks on client start. We cover two cases when the client starts:
- keep is inactive and we archive its' data
- keep is awaiting signatures and we start signing process.

~~Depends on: https://github.com/keep-network/keep-ecdsa/pull/372~~
Closes: https://github.com/keep-network/keep-ecdsa/issues/371


pkg/client/client.go
```diff
@@ -279,6 +348,43 @@ func checkAwaitingSignature(
...
+		isStillAwaitingSignature, err := waitForChainConfirmation(
+			ethereumChain,
+			currentBlock,
+			func() (bool, error) {
+				return ethereumChain.IsAwaitingSignature(keepAddress, latestDigest)
+			},
+		)
+		if err != nil {
+			logger.Errorf(
+				"failed to confirm signing request for digest [%+x] and keep [%s]",
```
lukasz-zimnoch, line 372: Maybe we can log `err` as well?

pkg/client/client.go
```diff
@@ -279,6 +348,43 @@ func checkAwaitingSignature(
...
 	if isAwaitingDigest {
+		if err := requestedSignatures.add(keepAddress, latestDigest); err != nil {
+			logger.Errorf("failed to register signature request [%v]", err)
+			return
+		}
+		defer requestedSignatures.remove(keepAddress, latestDigest)
+
+		currentBlock, err := ethereumChain.BlockCounter().CurrentBlock()
+		if err != nil {
+			logger.Errorf("failed to get current block height [%v]", err)
```
lukasz-zimnoch, line 359: Same with the colon here.

pkg/client/client.go
```diff
@@ -279,6 +348,43 @@ func checkAwaitingSignature(
 	}
 
 	if isAwaitingDigest {
+		if err := requestedSignatures.add(keepAddress, latestDigest); err != nil {
+			logger.Errorf("failed to register signature request [%v]", err)
```
lukasz-zimnoch, line 352: If I'm not wrong we have a convention to add a `:` before we print `err`. So I think it should be: `logger.Errorf("failed to register signature request: [%v]", err)`

pkg/client/client.go
```diff
@@ -223,32 +283,40 @@ func monitorSigningRequests(
...
+					ethereumChain,
+					event.BlockNumber,
+					func() (bool, error) {
+						return ethereumChain.IsAwaitingSignature(keepAddress, event.Digest)
+					},
 				)
-				return
-			}
+				if err != nil {
+					logger.Errorf(
```
lukasz-zimnoch, line 301: Can we log the cause (`err`)?

pkg/client/client.go
```diff
@@ -45,28 +46,78 @@ func Initialize(
...
+					return
+				}
+				logger.Warningf("keep [%s] is still active", keepAddress.String())
+			}
+
 			signers, err := keepsRegistry.GetSigners(keepAddress)
 			if err != nil {
+				// If there are no signers for loaded keep that something is clearly
+				// wrong. We don't want to continue processing for this keep.
 				logger.Errorf("no signers for keep [%s]", keepAddress.String())
```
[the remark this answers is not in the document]
pdyraga: I second that comment.

pkg/client/client.go
pdyraga: Agree with the comment above but this line needs to be refactored anyway after #372 is merged and we'll be no longer handling error here.

pkg/client/client.go
pdyraga: ☝️

pkg/client/client.go
pdyraga: ☝️

pkg/client/client.go
```diff
@@ -425,18 +531,18 @@ func monitorKeepTerminatedEvent(
...
 // a provided function returning a boolean value.
-func waitForEventConfirmation(
+func waitForChainConfirmation(
 	ethereumChain eth.Handle,
 	startBlockNumber uint64,
 	stateCheck func() (bool, error),
 ) (bool, error) {
 	blockHeight := startBlockNumber + blockConfirmations
-	logger.Infof("waiting for [%d] block", blockHeight)
+	logger.Infof("waiting for block [%d]", blockHeight)
```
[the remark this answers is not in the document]
pdyraga: Agreed. And I think `to confirm chain state` is even better.

Issue #317, opened 2020-03-19
Client breaks on status monitoring after being removed from sortition pool


nkuba:
The client is failing after updating the operator's status causing its removal from the sortition pool.

I've been running a client that was registered in the `BondedECDSAKeepFactory`.
After that I withdrew unbonded value for the given operator and noticed failures logged by the client:

```
2020-03-19 21:38:03.500964 INFO keep-tecdsa registration.go:223: updating operator status for application [0xa73a93cA44b73ECeEAEF5CaB56C0bd8e60B2390F]
2020-03-19 21:38:03.503284 DEBUG keep-contract-BondedECDSAKeepFactory BondedECDSAKeepFactory.go:601: submitting transaction updateOperatorStatus
 <F3>Ŗg<E4><A3>^B<BB>ɱ#^@<84><E5><FF>|F><E6>G <A7>:<93><CA>D<B7>><CE><EA><EF>\<AB>V<C0><BD><8E>`<B2>9^O
2020-03-19 21:38:05.685605 DEBUG keep-contract-BondedECDSAKeepFactory BondedECDSAKeepFactory.go:639: submitted transaction updateOperatorStatus with id: [0x2fa2997b08e78a8fd7ef7f0a555d8aeb076d598bdbac4fa021829278fc05a01f]
2020-03-19 21:38:05.686903 DEBUG keep-chain-eth-ethereum ethereum.go:341: submitted UpdateOperatorStatus transaction with hash: [2fa2997b08e78a8fd7ef7f0a555d8aeb076d598bdbac4fa021829278fc05a01f]
2020-03-19 21:38:07.187628 DEBUG keep-ethutil error_resolver.go:66: packing parameters for method [isOperatorUpToDate] with ABI [function isOperatorUpToDate(address _operator, address _application) constant returns(bool)]: [[[243 197 150 103 228 163 2 187 201 177 35 0 132 229 255 124 70 62 230 71] [167 58 147 202 68 183 62 206 234 239 92 171 86 192 189 142 96 178 57 15]]]
[...]

PR #376, opened 2020-04-08
Retry operator registration on status monitoring failure

nkuba:
Status monitoring can fail if the operator gets removed from the signer candidates pool. In such a case, we want to restart the registration loop to check if the operator is already registered and if not register when they are eligible.

Closes: https://github.com/keep-network/keep-ecdsa/issues/317

### DIFF
--- a/pkg/client/registration.go
+++ b/pkg/client/registration.go
@@ -2,9 +2,10 @@ package client
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/ethereum/go-ethereum/common"
-	"github.com/keep-network/keep-ecdsa/pkg/chain"
+	eth "github.com/keep-network/keep-ecdsa/pkg/chain"
 )
 
 // checkStatusAndRegisterForApplication checks whether the operator is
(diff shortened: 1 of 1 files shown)


pkg/client/registration.go
```diff
@@ -14,33 +15,44 @@ import (
...
-	isRegistered, err := ethereumChain.IsRegisteredForApplication(application)
-	if err != nil {
-		logger.Errorf(
-			"failed to check if member is registered for application [%s]: [%v]",
-			application.String(),
-			err,
-		)
-		return
-	}
+Registration:
```
lukasz-zimnoch, line 25: `Registration` -> `RegistrationLoop`?

pkg/client/registration.go
```diff
@@ -14,33 +15,44 @@ import (
...
+			return
+		default:
+			isRegistered, err := ethereumChain.IsRegisteredForApplication(application)
+			if err != nil {
+				logger.Errorf(
+					"failed to check if member is registered for application [%s]: [%v]",
+					application.String(),
+					err,
+				)
+				return
```
lukasz-zimnoch, line 38: Should we return? Since we are in the registration loop maybe it makes sense to retry in the next iteration?

Issue #302, opened 2020-03-09
Minimum stake


pdyraga:
https://github.com/keep-network/keep-core/issues/1368 specifies minimum stake parametrization and schedule. We need to make sure we follow the same minimum stake as `keep-core`. One option is to do a small refactoring and keep minimum stake on the `TokenStaking` contract. Then, all operator contracts can just refer there.

PR #377, opened 2020-04-08
Minimum stake schedule

pdyraga:
Closes: keep-network/keep-ecdsa#302

See: https://github.com/keep-network/keep-core/pull/1499
See: https://github.com/keep-network/sortition-pools/pull/74
See: https://github.com/keep-network/sortition-pools/pull/78

Minimum stake is no longer a system-wide constant and it changes over time in `TokenStaking`. `BondedECDSAKeepFactory` needs to be updated to support the minimum stake schedule.

The factory does no longer have the minimum stake hardcoded. To check if the operator has a minimum stake, it queries `TokenStaking` contract.

Also, the sortition pool now expects the recent value of the minimum stake to be passed for each group selection. This is happening in `openKeep` function.

Factory declares the eligible stake divisor for sortition pool to be `1 KEEP`.

### DIFF
--- a/solidity/test/BondedECDSAKeepFactoryTest.js
+++ b/solidity/test/BondedECDSAKeepFactoryTest.js
@@ -334,7 +334,7 @@ contract("BondedECDSAKeepFactory", async (accounts) => {
       )
     })
 
-    it("returns false if the operator stake increased", async () => {
+    it("returns true if the operator stake increased", async () => {
       await keepFactory.registerMemberCandidate(application, {
         from: members[0],
       })
--- a/solidity/test/BondedECDSAKeepFactoryTest.js
+++ b/solidity/test/BondedECDSAKeepFactoryTest.js
@@ -294,6 +294,8 @@ contract("BondedECDSAKeepFactory", async (accounts) => {
   })
 
   describe("isOperatorUpToDate", async () => {
+    const precision = new BN("1000000000000000000") // 1 KEEP = 10^18
+
     before(async () => {
       await initializeNewFactory()
       await initializeMemberCandidates()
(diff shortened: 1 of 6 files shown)


solidity/test/BondedECDSAKeepFactoryTest.js
```diff
@@ -307,44 +309,39 @@ contract("BondedECDSAKeepFactory", async (accounts) => {
...
+      await stakeOperators(members, minimumStake.subn(1))
 
-      assert.isTrue(
+      assert.isFalse(
         await keepFactory.isOperatorUpToDate(members[0], application)
       )
     })
 
-    it("returns false if the operator stake is above minimum", async () => {
+    it("returns false if the operator stake increased", async () => {
```
nkuba, line 337:
```suggestion
    it("returns true if the operator stake increased", async () => {
```
?

solidity/test/BondedECDSAKeepFactoryTest.js
```diff
@@ -307,44 +309,39 @@ contract("BondedECDSAKeepFactory", async (accounts) => {
...
         from: members[0],
       })
 
-      // We multiply minimumStake as sortition pools expect multiplies of the
-      // minimum stake to calculate stakers weight for eligibility.
-      await stakeOperators(members, minimumStake.mul(new BN(2)))
+      await stakeOperators(members, minimumStake.addn(1))
 
-      assert.isFalse(
+      assert.isTrue(
```
nkuba, line 344: Could you please explain this change? Have some rules changed on the sortition pool side?

solidity/test/BondedECDSAKeepFactoryTest.js
pdyraga: https://github.com/keep-network/keep-ecdsa/pull/377/commits/7b311d767ff61f09f1853fec5dcfb012230efab7

solidity/test/BondedECDSAKeepFactoryTest.js
pdyraga: I rethought tests after your comment and pushed more precise cases in https://github.com/keep-network/keep-ecdsa/pull/377/commits/3a61bf47809ab3b8fa01c27fc79473fa80934464

From https://github.com/keep-network/sortition-pools/pull/74:

> Minimum stake is no longer a hardcoded value and can change over time. To accommodate potential changes in minimum stake, we now use a constant pool weight divisor for bonded and standard sortition pool just like in the fully backed pool.
> 
> The initial minimum stake value is passed as a pool constructor parameter and the current value is later passed as an argument to select group function.

We agreed with @eth-r to set the pool weight divisor to 1 KEEP. There is an additional factor here: EVM performs implicit floor for `uint256` division. As a result:
- if stake drops by `10^-(18) KEEP`, `isOperatorUpToDate` returns `false`; this is covered by `returns false if the operator stake dropped` case
- if stake drops by `1 KEEP`, `isOperatorUpToDate` returns `false`; this is not covered by a separate case but is obvious from the previous one
- if stake increases by `1 KEEP`, `isOperatorUpToDate` returns `false`; this is covered by `returns false if the operator stake increased` case
- if stake increases by `10^(-18) KEEP`, `isOperatorUpToDate` returns `true`; this is covered by `returns true if the operator stake increase is below weight precision` case

solidity/test/BondedECDSAKeepFactoryTest.js
pdyraga: Nope, reworked in https://github.com/keep-network/keep-ecdsa/pull/377/commits/3a61bf47809ab3b8fa01c27fc79473fa80934464

PR #378, opened 2020-04-08
Monitor operator status changes based on events

nkuba:
In this PR we update the client to monitor events that are changing the operator's status in the pool instead of checking if it updated on every block. This change introduces some optimization to ethereum API calls.

We also had to update `KeepBonding` contract to emit events on bond-related changes.

Depends on https://github.com/keep-network/keep-ecdsa/pull/376


nkuba:
Heads up @sthompson22. This PR updated provisioning script as we added new addresses to the config file: https://github.com/keep-network/keep-ecdsa/pull/378/files#diff-3321a04e223905960e86ff144a30153b

pkg/chain/event.go
```diff
@@ -41,6 +43,34 @@ type KeepTerminatedEvent struct {
...
 }
 
+// UnbondedValueWithdrawnEvent is an event emitted when unbonded value has been
+// withdrawn for an operator.
+type UnbondedValueWithdrawnEvent struct {
+	Operator common.Address
+	Amount   *big.Int
+}
+
+// BondCreatedEvent is an event emitted a bond has been created.
```
lukasz-zimnoch, line 53: `... when a bond has been created.`

pkg/chain/event.go
```diff
@@ -41,6 +43,34 @@ type KeepTerminatedEvent struct {
...
+type BondCreatedEvent struct {
+	Operator    common.Address
+	Holder      common.Address
+	SignerPool  common.Address
+	ReferenceID *big.Int
+	Amount      *big.Int
+}
+
+// TokensSlashedEvent is an event emitted a tokens has been slashed for an operator.
```
lukasz-zimnoch, line 62: `... when tokens has been slashed for an operator.`

pkg/chain/event.go
```diff
@@ -41,6 +43,34 @@ type KeepTerminatedEvent struct {
...
+type TokensSlashedEvent struct {
+	Operator common.Address
+	Amount   *big.Int
+}
+
+// TokensSeizedEvent is an event emitted a tokens has been seized for an operator.
```
lukasz-zimnoch, line 68: `... when tokens has been seized for an operator.`

pkg/client/registration.go
```diff
@@ -211,11 +211,60 @@ func monitorSignerPoolStatus(
...
+		return fmt.Errorf("failed on registering for tokens seizure: [%v]", err)
+	}
+	defer subscriptionTokenSeized.Unsubscribe()
 
 	for {
 		select {
-		case <-newBlockChan:
+		case <-updateTrigger:
+			logger.Infof(
+				"operator status update triggered for application: [%s]",
```
lukasz-zimnoch, line 264: Maybe it's worth to log which event triggered the update?

pkg/client/registration.go
```diff
@@ -211,11 +211,60 @@ func monitorSignerPoolStatus(
 		application.String(),
 	)
 
-	newBlockChan := ethereumChain.BlockCounter().WatchBlocks(ctx)
+	updateTrigger := make(chan interface{})
```
lukasz-zimnoch, line 214: What do you think about extracting all subscription logic below to a separate function for better readability:
```
func subscribeStatusUpdateEvents() (chan interface{}, func()) {
    updateTrigger := make(chan interface{})

    // ... create all subscriptions

    unsubscribeAll := func() {
        subscription1.Unsubscribe()
        subscription2.Unsubscribe()

        // ... unsubscribe from the rest
     }

    return updateTrigger, unsubscribeAll
}
```
and use it here:
```
updateTrigger, unsubscribeAll := subscribeStatusUpdateEvents()
defer unsubscribeAll()
```

nkuba: Thanks man! I was thinking about that but couldn't find an elegant way to handle unsubscribe calls.

pkg/client/registration.go
nkuba: Do you think we need event with it's content or just a name of the event? The whole event with all it's parameters may give too much noise.

lukasz-zimnoch: I thought about the event name only. May be helpful when debugging.

pdyraga:
I think we can simplify it. Some initial assumptions first:

In practice, stake delegated to operator [can not increase](https://github.com/keep-network/keep-core/blob/v0.12.0-rc/solidity/contracts/TokenStaking.sol#L130). It can only go down as a result of slashing/seizing.

Operator is interested in updating sortition pool weight when stake decreases. Since we never slash/seize less than minimum stake and pool weight precision is set to 1 KEEP, we know that we always need to update pool when we were punished for misbehaviour. We don't need additional checks.

Operator can be removed from pool when the available unbonded value drops below the required minimum. Operator needs to monitor bonding contract outside of the client. Operator is interested in updating sortition pool *only* when it has been removed from the pool because of not having enough unbounded value. It does not want to update pool earlier because the operator would lose a chance for updating bonding contract without paying additional pool join fee.

Operator can be removed from a pool because of the insufficient bond only if it was selected and available unbounded value is not enough. The simplest option to get notified about group selection without modifying anything on-chain is to listen for `BondedECDSAKeepCreated` from the factory.

What I think is the minimum and sufficient version:
- `TokensSlashed` emitted for the operator -> operator updates the pool immediately without additional checks; they don't make any sense
- `TokensSeized` emitted for the operator -> same as above
- `BondedECDSAKeepCreated` emitted -> operator executes a part of `checkStatusAndRegisterForApplication` (we probably need to extract the logic without `monitorSignerPoolStatus` to a separate function) to check if it's still in the pool and if not, join as soon as it becomes eligible; we also emit a warning about the fact it was removed from the pool.

Nothing less, nothing more.

pdyraga:
We can add a retry delay parameter for the function we'll extract from `checkStatusAndRegisterForApplication`. When the client starts, we check the status for every new block and log a warning if the operator is not eligible. When we were kicked off the pool, we do the same but with a greater delay, e.g. 10 minutes. The entire point is that it should be a post-factum check and the monitoring of unbounded value should be implemented out of the client.

pdyraga:
After thinking about it a little more, we see that even with the strategy proposed above it is getting way too complex. For example, when we receive stake slash/seized event, we should:
1. Update operator status
2. Wait until transaction is confirmed
3. Check if the operator is still registered in sortition pool.
4. If the operator is not registered, we should check the minimum stake; if minimum stake is below the required minimum, we should give up
5. If the operator is not registered, it means there is not enough unbounded value and we should start monitoring operator's eligibility to join the pool again.

Instead of using events, we'll just increase the monitoring delay to 100 blocks (~15 minutes) when the operator is registered in the pool. Slashing/seizing should be so rare case that it does not make sense to invest time in developing solution updating pool immediately. Bond should be monitored by operator anyway and if it goes below the minimum, operator/owner needs some time to increase this value; ~15 minutes delay seems sane in this case.

Issue #336, opened 2020-03-30
Throw critical error and kill client when can't get to storage dir


nkuba:
During the client start process when it cannot get to the path configured as the storage data-dir a critical error should be reported and client process should be exited.

We should also validate if the process is able to both read and write to the directory.

If the client is not ensuring above it can run the protocol and fail on saving generated signer data.

PR #379, opened 2020-04-08
Handling errors for creating a new storage disk handler

dimpar:
Refs: https://github.com/keep-network/keep-ecdsa/issues/336

In this PR we are pre-checking if there are any potential errors with reading/writing to the disk storage.

~~depends on `keep-commons` https://github.com/keep-network/keep-common/pull/25, will need to update the version in `mod.go`~~

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,16 +12,14 @@ replace (
 
 require (
 	github.com/BurntSushi/toml v0.3.1
-	github.com/aristanetworks/goarista v0.0.0-20200211191935-58c705f5cf52 // indirect
 	github.com/binance-chain/tss-lib v1.3.1
 	github.com/btcsuite/btcd v0.20.1-beta
 	github.com/ethereum/go-ethereum v1.9.10
 	github.com/gogo/protobuf v1.3.1
 	github.com/ipfs/go-log v0.0.1
-	github.com/keep-network/keep-common v0.1.1-0.20200330215727-8ca95ff9de56 // TODO: update to released version
+	github.com/keep-network/keep-common v0.1.1-0.20200409131849-1363a51644b7 // TODO: update to released version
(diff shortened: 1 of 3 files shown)


lukasz-zimnoch:
@dimpar could you set the right version of `keep-common`?

go.mod
```diff
@@ -12,16 +12,14 @@ replace (
...
 	github.com/binance-chain/tss-lib v1.3.1
 	github.com/btcsuite/btcd v0.20.1-beta
 	github.com/ethereum/go-ethereum v1.9.10
 	github.com/gogo/protobuf v1.3.1
 	github.com/ipfs/go-log v0.0.1
-	github.com/keep-network/keep-common v0.1.1-0.20200330215727-8ca95ff9de56 // TODO: update to released version
+	github.com/keep-network/keep-common v0.1.1-0.20200409131849-1363a51644b7 // TODO: update to released version
 	github.com/keep-network/keep-core v0.10.0
-	github.com/olekukonko/tablewriter v0.0.4 // indirect
+	github.com/keep-network/keep-tecdsa v0.9.0
```
lukasz-zimnoch, line 22: It seems a little weird. @nkuba any thoughts?

dimpar: looked a bit weird for me too, and I started this conversation on [fd](https://www.flowdock.com/app/cardforcoin/tech/threads/mA-ShJwBV6TzagPN0PSLIWmurtK)

but yea, let's confirm it

nkuba: No idea how that happened. Run `go mod tidy`.

dimpar: I think I know why. My local path was `keep-network/keep-tecdsa` instead of `/keep-ecdsa` and it added additional dependency. Anyway, removed now

Issue #391, opened 2020-04-10
Where are the docs? 


creckhow:
keep-core has a great set of docs on the random beacon -- a glossary, details on misbehavior and slashing, how to run and operate the client. 

Am I correct, these don't exist yet for ecdsa/tBTC? Of particularly urgency for our business partners is information about running and operating the client, and ETH slashing/liquidation parameters. 

@Shadowfiend @sthompson22 @pdyraga let me know how I can help. I will just start forking Promethea's beacon docs and filling them with vague guesses over here in this repo at some point. :):):)

pdyraga:
We don't have anything for ECDSA keeps yet. The best source of information here is tBTC spec for now.

Shadowfiend:
> I will just start forking Promethea's beacon docs and filling them with vague guesses over here in this repo at some point. 

This is probably not the right thing to do. The beacon “docs” are a full spec, not documentation. The keep-ecdsa docs will *most likely* start as something less complete for folks to get oriented, and will be referring to the “beacon” docs for a lot of things, as what is currently the beacon spec also defines a lot of the root of the Keep system (e.g., authorizations and staking).

It would be better if you wanted to compile a list of questions you need/would like the docs to answer, as well as some initial takes at answers if you have them. If you want to base that on what's in the beacon docs, that's fine. We can gather that in this issue and start turning it into either checkboxes or multiple issues as we start organizing the written work.

mhluongo:
> ETH slashing/liquidation parameters.

This part is 100% in [`tBTC`](http://github.com/keep-network/tbtc). The spec covers all things that can happen with your ETH, *accept* what happens with it after a deposit is successfully redeemed IIRC.

Some stuff that might be appropriate in this issue though

- [ ] How does `keep-ecdsa` relate to `keep-core`? Is it a separate client?
- [ ] How do I opt in to using the `BondedECDSAKeep`?
- [ ] What other contracts do I need to authorize?
- [ ] What are the system requirements for running `keep-ecdsa`?
- [ ] Where are key shares stored? How should I ensure they aren't corrupted?
- [ ] How does ETH custody work? Can I run `keep-ecdsa` non-custodially? (NB - My understanding here is that anyone can bond for an operator, and that we're working to make sure beneficiaries work the same way as with KEEP. This is a high priority question for professional stakers)

PR #396, opened 2020-04-13
docs: Run Keep ECDSA

sthompson22:
Drafting for visibility and other contributors to add info

=======

The intended audience for this document is anyone who is interested in running their own Keep ECDSA client.

Here we provide a semi-detailed overview of mainnet/testnet context, operational considerations, and app/logging/network configurations for running a Keep ECDSA client.

This document is meant to be living, especially as we work through helping folks onboard in Discord. It's important that we're identifying gaps through those Discord interactions and updating this doc accordingly.

The information here certainly isn't exhaustive, on this pass we're aiming to cover the most critical considerations for running a client.

ref: https://github.com/keep-network/keep-ecdsa/issues/391


docs/run-keep-ecdsa.adoc
```diff
@@ -0,0 +1,419 @@
...
+[%header,cols=2*]
+|===
+|Egress
+|Port
+
+|Ethereum Network
+|`8545` / `8546`
+
+|Keep Network
+|`3919`
```
afmsavage, line 56: Note that these are TCP ports.  Will add some clarity when people set up their security groups (AWS) or firewall rules.
```suggestion
| TCP: `8545` / `8546`

|Keep Network
| TCP: `3919`
```

docs/run-keep-ecdsa.adoc
```diff
@@ -0,0 +1,419 @@
...
+
+=== Run Image
+This is a sample run command for illustration purposes only.
+
+```
+export KEEP_CLIENT_ETHEREUM_PASSWORD=$(cat .secrets/eth-account-password.txt)
+export KEEP_CLIENT_CONFIG_DIR=$(pwd)/config
+export KEEP_CLIENT_PERSISTENCE_DIR=$(pwd)/persistence
+
+docker run -dit \
```
afmsavage, line 247: There is no reason to use the `-it`  options here if using `-d`.  You are essentially telling Docker to open an interactive terminal but detach from it.  
```suggestion
docker run -d \
```

docs/run-keep-ecdsa.adoc
```diff
@@ -0,0 +1,419 @@
...
+== Docker
+
+=== Get Image
+
+Doesn't exist for the public yet.
+
+=== Run Image
+This is a sample run command for illustration purposes only.
+
+```
```
afmsavage, line 242: Not sure if this works on .adoc files, but you might be able to add bash after the gate open to get syntax highlighting on the commands within this code block.
```suggestion
``` bash
```

docs/run-keep-ecdsa.adoc
```diff
@@ -0,0 +1,419 @@
...
+non-default ports configured, firewall rules will need to be adjusted accordingly.
+
+=== Application
+
+Application configurations are stored in a `.toml` file and passed to the application run command
+ with the `--config` flag.
+
+==== Sample
+
+```
```
afmsavage, line 78: Add toml after start of code fence for syntax highlighting.
```suggestion
``` toml
```

docs/run-keep-ecdsa.adoc
```diff
@@ -0,0 +1,419 @@
...
+[%header,cols=2*]
+|===
+|Cloud Provider
+|Machine Type
+
+|Google Cloud
+|n1-standard-4
+
+|AWS
+|m5.xlarge
```
afmsavage, line 32: `c5.xlarge` is a 4x8 machine and costs about $18 less per month than the `m5.xlarge` as a suggestion to keep it cheaper on users.  I do not know the specs you are going off of, I am just using the specs of a self-hosted being 4x4 so my guess is that CPU is more important than memory.

docs/run-keep-ecdsa.adoc
```diff
@@ -0,0 +1,419 @@
...
+[TSS]
+# Timeout for TSS protocol pre-parameters generation. The value
+# should be provided based on resources available on the machine running the client.
+# This is an optional parameter, if not provided timeout for TSS protocol
+# pre-parameters generation will be set to `2 minutes`.
+  PreParamsGenerationTimeout = "2m30s"
+
+```
+
+==== Parameters
```
afmsavage, line 118: I tried to think about a better table layout for these parameters.  The non-uniform size of each of them is hard to read.  Might want to make it 5 columns wide and just include the left-most column header currently an entire new column all the way down.  Then you'd be uniform.  

Something like this? 

[%header,cols=5*]
|===
|Type
|Description
|Default
|Required

|`ethereum`
|`URL`
|The Ethereum host your keep-ecdsa will connect to.  Websocket protocol/port.
|""
|Yes

docs/run-keep-ecdsa.adoc
Shadowfiend: It should, though I don’t know if the interstitial space will mess it up.

docs/run-keep-ecdsa.adoc
sthompson22: > There is no reason to use the `-it` options here if using `-d`.

This is a weird one.  Even in detached mode if you set the `entrypoint` to `bash` or `sh` you need the pseudo-terminal to keep the container running in detached mode, hence the `-dit`.  To your point this is kind of a holdover from `docker run` scripts I use for local testing, where I frequently `--entrypoint /bin/sh`.  I should be able to remove this here.

docs/run-keep-ecdsa.adoc
sthompson22: Will give it a shot!

docs/run-keep-ecdsa.adoc
sthompson22: Will try!

docs/run-keep-ecdsa.adoc
sthompson22: You are right that CPU is more important here.  My specs are based off more general application such as being a Kube Node, where having the available memory for other components is relevant.  We should probably capture both cases here, if you're running on a VM directly or as part of a multi-tenant setup such as Kube.

docs/run-keep-ecdsa.adoc
sthompson22: Sure, can see what it looks like.

docs/run-keep-ecdsa.adoc
```diff
@@ -0,0 +1,419 @@
...
+toc::[]
+
+== System Considerations
+
+The Keep Network expects certain capabilites for each node running on the network.  To help attain
+these capabilities consider the following criteria:
+
+- It is paramount that Keep nodes remain available to the Keep Network. We strongly encourage a
+  stable and redundant internet connection.
+- A connection to a production grade self-hosted or third party Ethereum node deployment.
```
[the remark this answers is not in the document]
sthompson22: I've purposely tried to be a bit obtuse here so folks don't just pick what we have listed here.  Will think on this a bit.

docs/run-keep-ecdsa.adoc
sthompson22: Nice one!

sthompson22:
@afmsavage Thanks a lot for the feedback.  Going through these now.

docs/run-keep-ecdsa.adoc
sthompson22: This didn't render but 

```
[source,bash]
----
code
----
```

Seems to work.

docs/run-keep-ecdsa.adoc
sthompson22: I had a look at how this renders on our docs page: https://docs.keep.network/run-random-beacon.html#_application, and they line up.  I guess this is a product of how GH renders adoc files.

sthompson22: Noting ^ is a different document, but formatted the same way.

sthompson22:
Ready for another pass.

docs/run-keep-ecdsa.adoc
```diff
@@ -0,0 +1,431 @@
...
+and authorizer. Owner may delegate owned tokens or tokens from a grant. Owner may decide to delegate just a portion
+of owned tokens or just a part of tokens from a grant. Owner may delegate multiple times to different operators.
+Tokens can be delegated using Tokens page in https://dashboard.test.keep.network[KEEP token dashboard] and a certain minimum stake defined by the system is required to be provided in the delegation. The more stake is delegated, the higher chance to be selected to relay group.
+
+Delegation takes immediate effect but can be cancelled within one week without additional delay. After one week
+operator appointed during the delegation becomes eligible for work selection.
+
+=== Authorizations
+Before operator is considered as eligible for work selection, authorizer appointed during the delegation needs to review
+and authorize TBTCSystem smart contract. Smart contracts can be authorized using KEEP token dashboard. Authorized operator contracts may slash or seize tokens in case of operator's misbehavior.  Authorized operator contracts are set in `SanctionedApplications.Addresses` in your Keep ECDSA config.
```
pdyraga, line 429: There are two authorizations that have to be done:
- authorizer has to authorize `BondedECDSAKeepFactory` operator contract in `TokenStaking` contract - it lets the factory to slash tokens on misbehaviour and makes the operator eligible for work selection
- authorizer has to provide a secondary authorization for the sortition pool created for the particular application to operate on the bonds stored in `KeepBonding` contract.

If we do not upgrade the system, there will be only one `Tokenstaking` contract, `KeepBonding` contract, and only one `BondedECDSAKeepFactory`. The factory creates a sortition pool per application. So authorizer is de facto authorizing the application to operate on bonds, just indirectly.

Both actions will be available in the dashboard.

sthompson22: Thanks @pdyraga tried to distill that into a couple bullets.  Let me know what needs tweaking.

docs/run-keep-ecdsa.adoc
afmsavage: Gotcha, that makes sense.  I'm just being picky here really.

docs/run-keep-ecdsa.adoc
afmsavage: I'm cheap mostly is why I brought this up.  I know a lot of people are cheap too so mostly trying to make it so that someone doesn't run a node because of the cost.

PR #399, opened 2020-04-15
BondedECDSAKeepFactory implements IRandombeaconConsumer

pdyraga:
Closes https://github.com/keep-network/keep-core/issues/1537
Depends on https://github.com/keep-network/keep-core/pull/1609

It is to ensure the beacon callback function has a correct signature. If it has not, the contract cannot be deployed:

```
Error:  *** Deployment Failed ***

"BondedECDSAKeepFactory" is an abstract contract or an interface and cannot be deployed.
   * Import abstractions into the '.sol' file that uses them instead of deploying them separately.
   * Contracts that inherit an abstraction must implement all its method signatures exactly.
   * A contract that only implements part of an inherited abstraction is
   also considered abstract.
```

### DIFF
--- a/solidity/contracts/BondedECDSAKeepFactory.sol
+++ b/solidity/contracts/BondedECDSAKeepFactory.sol
@@ -29,7 +29,8 @@ import "openzeppelin-solidity/contracts/math/SafeMath.sol";
 contract BondedECDSAKeepFactory is
     IBondedECDSAKeepFactory,
     CloneFactory,
-    AuthorityDelegator
+    AuthorityDelegator,
+    IRandomBeaconConsumer
 {
     using AddressArrayUtils for address[];
     using SafeMath for uint256;
--- a/solidity/package-lock.json
+++ b/solidity/package-lock.json
@@ -879,9 +879,9 @@
       }
     },
     "@keep-network/keep-core": {
-      "version": "0.14.0-pre.0",
-      "resolved": "https://registry.npmjs.org/@keep-network/keep-core/-/keep-core-0.14.0-pre.0.tgz",
-      "integrity": "sha512-etiQWKwtxTCA7QFQ/wsvYAu8owdddEDOEGZ3H1QMPljbyKiDuvHmdAdzeuGsou1+i9fL/XzZPv32ctWob5xT/w==",
+      "version": "0.14.0-pre.4",
+      "resolved": "https://registry.npmjs.org/@keep-network/keep-core/-/keep-core-0.14.0-pre.4.tgz",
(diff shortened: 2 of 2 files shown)


solidity/contracts/BondedECDSAKeepFactory.sol
```diff
@@ -29,7 +29,8 @@ import "openzeppelin-solidity/contracts/math/SafeMath.sol";
 contract BondedECDSAKeepFactory is
     IBondedECDSAKeepFactory,
     CloneFactory,
-    AuthorityDelegator
+    AuthorityDelegator,
+    IRandomBeaconConsumer
```
nkuba, line 33: Aren't we missing import of `IRandomBeaconConsumer`?

nkuba: Oh I see, it gets imported in `import "@keep-network/keep-core/contracts/IRandomBeacon.sol";`.
Can we be explicit what we're importing? E.g.:
```solidity
import { 
    IRandomBeacon, 
    IRandomBeaconConsumer 
} "@keep-network/keep-core/contracts/IRandomBeacon.sol";.
```

pdyraga: I'd love to but I am getting some truly bizarre output from truffle then. 
```
import {
    IRandomBeacon,
    IRandomBeaconConsumer
} from "@keep-network/keep-core/contracts/IRandomBeacon.sol";
``` 
```
/Users/piotr/go/src/github.com/keep-network/keep-ecdsa/solidity/contracts/BondedECDSAKeepFactory.sol:17:1: ParserError: Source "@keep-network/keep-core/contracts/IRandomBeacon.sol" not found: File import callback not supported
import {
^ (Relevant source part starts here and spans across multiple lines).

Compilation failed. See above.
Truffle v5.1.18 (core: 5.1.18)
Node v11.15.0
```

Do you know what could be happening here? I am out of ideas.

nkuba: I'm not sure why it doesn't work. Let's leave it as-is 🤷

pdyraga:
Ah, I forgot to undraft it. Ready @nkuba

solidity/contracts/BondedECDSAKeepFactory.sol
pdyraga: Yeah, I tried everything. I thought it's something about borked `build` artifacts or something not updating in `node_modules`. Tried all the combinations, and have the same error all the time.

PR #400, opened 2020-04-15
keep-test: Reduce ecdsa unbonded deposit

sthompson22:
Drafting this for visibility during the build process.  Will close when done.

We're reducing the unbonded deposit to reduce the amount of ETH needed for application deployment during scale-out testing.

We'll fix this properly against master by making this value configurable via Kube config.  I will do this in a separate PR.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -224,7 +224,7 @@ jobs:
       - run:
           name: Tag Docker images
           command: |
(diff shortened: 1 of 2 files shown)


sthompson22:
https://app.circleci.com/pipelines/github/keep-network/keep-ecdsa/1803/workflows/f6ed95fa-bc8c-44de-afee-2182272576b4/jobs/7093 published.

PR #406, opened 2020-04-16
Minimum stake or active keep membership firewall policy

pdyraga:
Closes https://github.com/keep-network/keep-core/issues/1588
Depends on https://github.com/keep-network/keep-core/pull/1612

For ECDSA keep client, we always connect with another client if it has a minimum stake. In a situation when the remote client has no minimum stake we should check if it is a member in at least one active keep. If so, we should allow establishing the connection.

The reason for that is to allow the client to participate in any outstanding signing in case it's been slashed, most probably by the beacon. The client with no minimum stake is not eligible to be selected to new ECDSA keeps but it can still participate in signing in existing keeps it's a member of.

Figuring out if the given client is a member in at least one active keep will take a bunch of calls to Ethereum client and to minimize them for every single firewall check tick, we cache the result in-memory for 24 hours.

### DIFF
--- a/pkg/firewall/cache.go
+++ b/pkg/firewall/cache.go
@@ -43,7 +43,7 @@ func (tc *timeCache) add(item string) bool {
 		return false
 	}
 
-	// sweep old entries
+	// sweep old entries (those for which caching timespan has passed)
 	for {
 		back := tc.indexer.Back()
 		if back == nil {
--- a/pkg/firewall/cache.go
+++ b/pkg/firewall/cache.go
@@ -9,8 +9,15 @@ import (
 // timeCache provides a time cache safe for concurrent use by
 // multiple goroutines without additional locking or coordination.
 type timeCache struct {
-	indexer  *list.List
-	cache    map[string]time.Time
+	// all items in the cache in the order they were added
+	// most recent items are on the front of the indexer;
+	// it is used to optimize cache sweeping
+	indexer *list.List
+	// item in the cache with the timestamp it's been added
+	// to the cache the last time
+	cache map[string]time.Time
+	// the timespan after which entry in the cache is considered
+	// as outdated and can be removed from the cache
 	timespan time.Duration
 	mutex    sync.RWMutex
 }
--- a/pkg/firewall/cache.go
+++ b/pkg/firewall/cache.go
@@ -53,7 +53,7 @@ func (tc *timeCache) add(item string) bool {
 		item := back.Value.(string)
 		itemTime, ok := tc.cache[item]
 		if !ok {
-			logger.Panicf(
+			logger.Errorf(
 				"inconsistent cache state - expected item [%v] is not present",
 				item,
 			)
--- a/pkg/firewall/cache.go
+++ b/pkg/firewall/cache.go
@@ -1,0 +1,77 @@
+package firewall
+
+import (
+	"container/list"
+	"sync"
+	"time"
+)
+
+// timeCache provides a time cache safe for concurrent use by
+// multiple goroutines without additional locking or coordination.
+type timeCache struct {
+	indexer  *list.List
+	cache    map[string]time.Time
+	timespan time.Duration
+	mutex    sync.RWMutex
+}
+
+// NewTimeCache creates a new cache instance with provided timespan.
+func newTimeCache(timespan time.Duration) *timeCache {
+	return &timeCache{
+		indexer:  list.New(),
+		cache:    make(map[string]time.Time),
+		timespan: timespan,
+	}
+}
+
+// Add adds an entry to the cache. Returns `true` if entry was not present in
+// the cache and was successfully added into it. Returns `false` if
+// entry is already in the cache. This method is synchronized.
+func (tc *timeCache) add(item string) bool {
+	tc.mutex.Lock()
+	defer tc.mutex.Unlock()
+
+	_, ok := tc.cache[item]
+	if ok {
+		return false
+	}
+
+	// sweep old entries
+	for {
+		back := tc.indexer.Back()
+		if back == nil {
+			break
+		}
+
+		item := back.Value.(string)
+		itemTime, ok := tc.cache[item]
+		if !ok {
+			logger.Fatalf(
+				"inconsistent cache state - expected item [%v] is not present",
+				item,
+			)
+			break
+		}
+
+		if time.Since(itemTime) > tc.timespan {
+			tc.indexer.Remove(back)
+			delete(tc.cache, item)
+		} else {
+			break
+		}
+	}
+
+	tc.cache[item] = time.Now()
+	tc.indexer.PushFront(item)
+	return true
+}
+
+// Has checks presence of an entry in the cache. Returns `true` if entry is
+// present and `false` otherwise.
+func (tc *timeCache) has(item string) bool {
+	tc.mutex.RLock()
+	defer tc.mutex.RUnlock()
+
+	_, ok := tc.cache[item]
+	return ok
+}
--- a/pkg/firewall/cache.go
+++ b/pkg/firewall/cache.go
@@ -53,7 +53,7 @@ func (tc *timeCache) add(item string) bool {
 		item := back.Value.(string)
 		itemTime, ok := tc.cache[item]
 		if !ok {
-			logger.Fatalf(
+			logger.Panicf(
 				"inconsistent cache state - expected item [%v] is not present",
 				item,
 			)
--- a/pkg/firewall/cache_test.go
+++ b/pkg/firewall/cache_test.go
@@ -1,0 +1,52 @@
+package firewall
+
+import (
+	"strconv"
+	"sync"
+	"testing"
+	"time"
+)
+
+func TestAdd(t *testing.T) {
+	cache := newTimeCache(time.Minute)
+
+	cache.add("test")
+
+	if !cache.has("test") {
+		t.Fatal("should have 'test' key")
+	}
+}
+
+func TestConcurrentAdd(t *testing.T) {
+	cache := newTimeCache(time.Minute)
+
+	var wg sync.WaitGroup
+	wg.Add(10)
+
+	for i := 0; i < 10; i++ {
+		go func(item int) {
+			cache.add(strconv.Itoa(item))
+			wg.Done()
+		}(i)
+	}
+
+	wg.Wait()
+
+	for i := 0; i < 10; i++ {
+		if !cache.has(strconv.Itoa(i)) {
+			t.Fatalf("should have '%v' key", i)
+		}
+	}
+}
+
+func TestExpiration(t *testing.T) {
+	cache := newTimeCache(500 * time.Millisecond)
+	for i := 0; i < 6; i++ {
+		cache.add(strconv.Itoa(i))
+		time.Sleep(100 * time.Millisecond)
+	}
+
+	if cache.has(strconv.Itoa(0)) {
+		t.Fatal("should have dropped '0' key from the cache already")
+	}
+}
--- a/pkg/firewall/firewall.go
+++ b/pkg/firewall/firewall.go
@@ -28,9 +28,9 @@ var errNoAuthorization = fmt.Errorf("remote peer has no authorization on the fac
 var errNoMinStakeNoActiveKeep = fmt.Errorf("remote peer has no minimum " +
 	"stake and is not a member in any of active keeps")
 
-// NewStakeOrActiveKeepPolicy if a firewall policy checking if the remote peer
-// has a minimum stake and if it has no minimum stake if it is a member in at
-// least one active keep.
+// NewStakeOrActiveKeepPolicy is a firewall policy checking if the remote peer
+// has a minimum stake and in case it has no minimum stake if it is a member of
+// at least one active keep.
 func NewStakeOrActiveKeepPolicy(
 	chain eth.Handle,
 	stakeMonitor coreChain.StakeMonitor,
--- a/pkg/firewall/firewall.go
+++ b/pkg/firewall/firewall.go
@@ -119,7 +119,7 @@ func (soakp *stakeOrActiveKeepPolicy) validateActiveKeepMembership(
 	lastIndex := new(big.Int).Sub(keepCount, one)
 
 	for keepIndex := new(big.Int).Set(lastIndex); keepIndex.Cmp(zero) != -1; keepIndex.Sub(keepIndex, one) {
-		keep, err := soakp.chain.GetKeepAtIndex(keepIndex)
+		keepAddress, err := soakp.chain.GetKeepAtIndex(keepIndex)
 		if err != nil {
 			logger.Errorf(
 				"could not get keep at index [%v]: [%v]",
@@ -133,11 +133,11 @@ func (soakp *stakeOrActiveKeepPolicy) validateActiveKeepMembership(
 		// active, we skip it. We still need to process the rest of the keeps
 		// because it's possible that although this keep is not active some
 		// peers created before this one are still active.
-		isActive, err := soakp.chain.IsActive(keep)
+		isActive, err := soakp.chain.IsActive(keepAddress)
 		if err != nil {
 			logger.Errorf(
 				"could not check if keep [%x] is active: [%v]",
-				keepIndex,
+				keepAddress,
 				err,
 			)
 			continue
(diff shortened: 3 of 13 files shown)


pdyraga:
@nkuba @lukasz-zimnoch Ready for early review, still working on tests.

pdyraga:
All is here, it's ready @nkuba @lukasz-zimnoch

pkg/firewall/firewall.go
```diff
@@ -0,0 +1,151 @@
...
+	// a member in at least one active keep. If so, we let to connect.
+	// Otherwise, we do not let to connect.
+	return soakp.validateActiveKeepMembership(remotePeerPublicKey)
+}
+
+func (soakp *stakeOrActiveKeepPolicy) validateActiveKeepMembership(
+	remotePeerPublicKey *ecdsa.PublicKey,
+) error {
+	remotePeerNetworkPublicKey := coreKey.NetworkPublic(*remotePeerPublicKey)
+	remotePeerAddress := coreKey.NetworkPubKeyToEthAddress(&remotePeerNetworkPublicKey)
```
nkuba, line 67: Do we need this `remote` prefix? 

```suggestion
	peerPublicKey *ecdsa.PublicKey,
) error {
	peerNetworkPublicKey := coreKey.NetworkPublic(*peerPublicKey)
	peerAddress := coreKey.NetworkPubKeyToEthAddress(&peerNetworkPublicKey)
```

pkg/firewall/firewall.go
```diff
@@ -0,0 +1,151 @@
...
+	// ones most recently created as there is a higher chance they are active.
+	keepCount, err := soakp.chain.GetKeepCount()
+	if err != nil {
+		return fmt.Errorf("could not get keep count: [%v]", err)
+	}
+
+	lastIndex := new(big.Int).Sub(keepCount, one)
+
+	for keepIndex := new(big.Int).Set(lastIndex); keepIndex.Cmp(zero) != -1; keepIndex.Sub(keepIndex, one) {
+		keep, err := soakp.chain.GetKeepAtIndex(keepIndex)
```
nkuba, line 98:
```suggestion
		keepAddress, err := soakp.chain.GetKeepAtIndex(keepIndex)
```

pkg/firewall/firewall.go
```diff
@@ -0,0 +1,151 @@
...
+			continue
+		}
+
+		// Get all the members of the active keep and store them in the active
+		// keep members cache.
+		members, err := soakp.chain.GetMembers(keep)
+		if err != nil {
+			logger.Errorf(
+				"could not get members of keep [%x]: [%v]",
+				keepIndex,
```
nkuba, line 131:
```suggestion
				"could not get members of keep [%s]: [%v]",
				keepAddress.String(),
```

pkg/firewall/firewall.go
```diff
@@ -0,0 +1,151 @@
...
+
+		// We are interested only in active keeps. If the current keep is not
+		// active, we skip it. We still need to process the rest of the keeps
+		// because it's possible that although this keep is not active some
+		// peers created before this one are still active.
+		isActive, err := soakp.chain.IsActive(keep)
+		if err != nil {
+			logger.Errorf(
+				"could not check if keep [%x] is active: [%v]",
+				keepIndex,
```
nkuba, line 116:
```suggestion
				"could not check if keep [%s] is active: [%v]",
				keepAddress.String(),
```

pkg/firewall/firewall_test.go
```diff
@@ -0,0 +1,228 @@
...
+		t.Fatal(err)
+	}
+
+	if err := policy.Validate(
+		key.NetworkKeyToECDSAKey(remotePeerPublicKey),
+	); err != nil {
+		t.Fatalf("validation should pass: [%v]", err)
+	}
+}
+
```
nkuba, line 204: Can we add a case where we validate a member that already exists in the cache?

solidity/contracts/BondedECDSAKeepFactory.sol
```diff
@@ -444,7 +444,17 @@ contract BondedECDSAKeepFactory is
...
     }
 
-    /// @dev Gets the stake balance of the specified operator.
+    /// @notice Checks if the factory has the authorization to operate on stake
+    /// represented by the provided operator.
+    ///
+    /// @param _operator operator's address
+    /// @return True if the factory has access to the staked token balance of
+    /// the provided operator and can slash that stake. False otherwise.
+    function hasAuthorization(address _operator) public view returns (bool) {
```
nkuba, line 453: We've got `isOperatorRegistered` and `isOperatorEligible` functions already, maybe we should name this function `isOperatorAuthorized` for consistency?

solidity/test/BondedECDSAKeepFactoryTest.js
```diff
@@ -1367,6 +1368,27 @@ contract("BondedECDSAKeepFactory", async (accounts) => {
     })
   })
 
+  describe("hasAuthorization", async () => {
+    before(async () => {
+      await initializeNewFactory()
+      await initializeMemberCandidates()
+    })
+
+    it("returns true if operator authorized the factory", async () => {
```
nkuba, line 1377: Technically it's not the operator who authorizes the factory.

```suggestion
    it("returns true if operator is authorized for the factory", async () => {
```

solidity/test/BondedECDSAKeepFactoryTest.js
```diff
@@ -1367,6 +1368,27 @@ contract("BondedECDSAKeepFactory", async (accounts) => {
...
+      assert.isTrue(
+        await keepFactory.hasAuthorization(members[0]),
+        "the operator has authorized the factory"
+      )
+    })
+
+    it("returns false if operator has not authorized the factory", async () => {
```
nkuba, line 1384:
```suggestion
    it("returns false if operator is not authorized for the factory", async () => {
```

solidity/test/BondedECDSAKeepFactoryTest.js
```diff
@@ -1367,6 +1368,27 @@ contract("BondedECDSAKeepFactory", async (accounts) => {
...
+  describe("hasAuthorization", async () => {
+    before(async () => {
+      await initializeNewFactory()
+      await initializeMemberCandidates()
+    })
+
+    it("returns true if operator authorized the factory", async () => {
+      assert.isTrue(
+        await keepFactory.hasAuthorization(members[0]),
+        "the operator has authorized the factory"
```
nkuba, line 1380:
```suggestion
        "the operator is authorized for the factory"
```

solidity/test/BondedECDSAKeepFactoryTest.js
```diff
@@ -1367,6 +1368,27 @@ contract("BondedECDSAKeepFactory", async (accounts) => {
...
+      )
+    })
+
+    it("returns false if operator has not authorized the factory", async () => {
+      assert.isFalse(
+        await keepFactory.hasAuthorization(notMember),
+        "the operator has not authorized the factory"
```
nkuba, line 1387:
```suggestion
        "the operator is not authorized for the factory"
```

pkg/firewall/cache.go
```diff
@@ -0,0 +1,77 @@
...
+	mutex    sync.RWMutex
+}
+
+// NewTimeCache creates a new cache instance with provided timespan.
+func newTimeCache(timespan time.Duration) *timeCache {
+	return &timeCache{
+		indexer:  list.New(),
+		cache:    make(map[string]time.Time),
+		timespan: timespan,
+	}
```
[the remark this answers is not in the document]
pdyraga: No, we should be fine because it's not a pointer.

pkg/firewall/cache.go
```diff
@@ -0,0 +1,77 @@
...
+func (tc *timeCache) add(item string) bool {
+	tc.mutex.Lock()
+	defer tc.mutex.Unlock()
+
+	_, ok := tc.cache[item]
+	if ok {
+		return false
+	}
+
+	// sweep old entries
```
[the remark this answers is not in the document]
pdyraga: Yes, I'll clarify it in the comment.

pkg/firewall/cache.go
```diff
@@ -0,0 +1,77 @@
...
+	for {
+		back := tc.indexer.Back()
+		if back == nil {
+			break
+		}
+
+		item := back.Value.(string)
+		itemTime, ok := tc.cache[item]
+		if !ok {
+			logger.Fatalf(
```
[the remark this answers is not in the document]
pdyraga: It's a tough decision. It's not generally possible the cache can be inconsistent if there are no bugs in the implementation so this logger should be never printed.

I am fine with both. Fatal printed to the logs multiple times should wake up the operator anyway.

pkg/firewall/cache_test.go
```diff
@@ -0,0 +1,52 @@
...
+		}(i)
+	}
+
+	wg.Wait()
+
+	for i := 0; i < 10; i++ {
+		if !cache.has(strconv.Itoa(i)) {
+			t.Fatalf("should have '%v' key", i)
+		}
+	}
```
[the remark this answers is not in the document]
pdyraga: We don't generally care if they are stored this way. It's just an implementation detail allowing us to sweep the cache faster. What we do care about is to remove only old entries and this is covered in `TestExpiration`.

pkg/firewall/firewall.go
```diff
@@ -0,0 +1,151 @@
...
+import (
+	"crypto/ecdsa"
+	"fmt"
+	"math/big"
+	"time"
+
+	"github.com/ipfs/go-log"
+
+	coreChain "github.com/keep-network/keep-core/pkg/chain"
+	coreFirewall "github.com/keep-network/keep-core/pkg/firewall"
```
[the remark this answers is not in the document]
pdyraga: > It was my initial plan but then I realized we don't want `net` to depend on `chain`. Right now, `net` defines `Firewall` interface so it doesn't care what's behind it.

pkg/firewall/firewall.go
```diff
@@ -0,0 +1,151 @@
...
+	minimumStakePolicy     coreNet.Firewall
+	activeKeepMembersCache *timeCache
+}
+
+func (soakp *stakeOrActiveKeepPolicy) Validate(
+	remotePeerPublicKey *ecdsa.PublicKey,
+) error {
+	// Validate minimum stake policy. If the remote peer has the minimum stake,
+	// we are fine and we should let to connect.
+	if err := soakp.minimumStakePolicy.Validate(remotePeerPublicKey); err == nil {
```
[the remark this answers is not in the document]
pdyraga: I am not sure about this. I think it's a normal situation from the operator's perspective, we shouldn't probably warn about anything here.

pkg/firewall/firewall.go
pdyraga: It's a convention we use in the network layer to indicate it's not the current's peer property.

pkg/firewall/firewall_test.go
pdyraga: I added a test case ensuring we cache active keep members but it's hard to write a good test ensuring we don't hit the chain in case the given member is already in the cache. This would be possible if we had a separate chain interface for the factory, similarly how we have a separate chain interface for signing or threshold relay or groups in `keep-core`. Without this refactoring, it's hard to implement such a case without dirty hacks. Or at least I have no idea how to do it without dirty hacks.

pdyraga:
Ready for another chance!

pkg/firewall/cache.go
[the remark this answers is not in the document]
nkuba: > Fatal printed to the logs multiple times should wake up the operator anyway.

Just to confirm we are on the same page, fatal is killing the process, right?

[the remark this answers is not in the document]
pdyraga: I think we are not on the same page :)

You asked
> do we want to kill the client in case of this error?

And my answer was *I do not know* while the current code here is NOT killing the process. We are logging on a `FATAL` level but we are not panicking. 

The severity is:
```
DEBUG < INFO < WARN < ERROR < DPANIC < PANIC < FATAL
```

When I look at it, I think we probably should change this log level to panic from fatal but I think we should avoid panicking here. It does not stop the application from processing, firewall checks may just get more expensive. I think it's better than killing the client and risking slashing.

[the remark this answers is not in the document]
pdyraga: Updated in https://github.com/keep-network/keep-ecdsa/pull/406/commits/c5823ee900043c50a46058a11bbf02aab82c0a6a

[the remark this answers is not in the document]
nkuba: > And my answer was I do not know while the current code here is NOT killing the process. We are logging on a FATAL level but we are not panicking.

We are logging with `Fatalf` which as per [documentation](https://godoc.org/github.com/op/go-logging#Logger.Fatalf) is killing the client.
> Fatalf is equivalent to l.Critical followed by a call to os.Exit(1). 

---

> When I look at it, I think we probably should change this log level to panic from fatal but I think we should avoid panicking here.

So do you want to panic or not? 

---

> It does not stop the application from processing, firewall checks may just get more expensive. I think it's better than killing the client and risking slashing.

Panic is also killing the client.

---

The difference between `Panic` and `Fatal` will be in handling defers. But as I understand the client will end up closed in each case.

See: https://stackoverflow.com/a/35997323

pkg/firewall/firewall_test.go
nkuba: Do we want to create an issue to do that in the future?

pkg/firewall/firewall.go
```diff
@@ -0,0 +1,175 @@
...
+	coreKey "github.com/keep-network/keep-core/pkg/net/key"
+
+	eth "github.com/keep-network/keep-ecdsa/pkg/chain"
+)
+
+var logger = log.Logger("keep-firewall")
+
+// KeepCacheLifetime is the time the cache maintains the list of active keep
+// members. We use the cache to minimize calls to Ethereum client.
+const KeepCacheLifetime = 24 * time.Hour
```
nkuba, line 24: When we were discussing this PR we were thinking about increasing the cache lifetime to 7 days. I just want to confirm that we didn't miss an update here and you finally decided to go with 1 day.

pkg/firewall/cache.go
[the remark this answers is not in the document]
pdyraga: Good catch. I was looking at `ipfs/go-log` docs in the source code and it's not mentioned there so I assumed logger is just writing the message to the correct level and it's up to the programmer to decide what to do next. I experimented with it and `logger.Fatalf` results in:
```
Starting keep-ecdsa client...
14:58:15.310 CRITI   keep-cmd: inconsistent cache state - expected item [123] is not present start.go:90
```

While `logger.Panicf` results in:
```
14:54:16.769 CRITI   keep-cmd: inconsistent cache state - expected item [123] is not present start.go:90
panic: inconsistent cache state - expected item [123] is not present

goroutine 1 [running]:
github.com/whyrusleeping/go-logging.(*Logger).Panicf(0xc00011daa0, 0x4b5bc63, 0x3c, 0xc000221810, 0x1, 0x1)
        /Users/piotr/go/pkg/mod/github.com/whyrusleeping/go-logging@v0.0.0-20170515211332-0457bb6b88fc/logger.go:206 +0x145
github.com/keep-network/keep-ecdsa/cmd.Start(0xc00016a2c0, 0x0, 0xc00016a2c0)
        /Users/piotr/go/src/github.com/keep-network/keep-ecdsa/cmd/start.go:90 +0x1a0
github.com/urfave/cli.HandleAction(0x49b1080, 0x4cbfcd0, 0xc00016a2c0, 0xc0004aa200, 0x0)
        /Users/piotr/go/pkg/mod/github.com/keep-network/cli@v1.20.0/app.go:490 +0xc8
github.com/urfave/cli.Command.Run(0x4b26082, 0x5, 0x0, 0x0, 0x0, 0x0, 0x0, 0x4b52aec, 0x2f, 0x0, ...)
        /Users/piotr/go/pkg/mod/github.com/keep-network/cli@v1.20.0/command.go:210 +0x9e8
github.com/urfave/cli.(*App).Run(0xc0001201a0, 0xc000138000, 0x4, 0x4, 0x0, 0x0)
        /Users/piotr/go/pkg/mod/github.com/keep-network/cli@v1.20.0/app.go:255 +0x741
main.main()
        /Users/pi:53 +0x3a2
```

I'd suggest we go with `Errorf` in this case. I am not happy with it because I think it's worse than ERROR but at the same time, I don't think we should `panic` here on production.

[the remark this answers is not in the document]
nkuba: So maybe [`Criticalf`](https://github.com/op/go-logging/blob/master/logger.go#L203)?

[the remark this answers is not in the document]
pdyraga: I'd love to but there is no `Criticalf` in `ipfs` logger: https://github.com/ipfs/go-log/blob/master/log.go#L14

pkg/firewall/firewall.go
pdyraga: Done in https://github.com/keep-network/keep-ecdsa/pull/406/commits/263b1773e379e0d81245dde9f4ef1c4f7caaf2a7

pkg/firewall/firewall_test.go
pdyraga: https://github.com/keep-network/keep-ecdsa/issues/420

Issue #395, opened 2020-04-13
Retries loop is missing time delay


nkuba:
There is a retry loop for a key generation that has no delay interval between runs that causes problems when an early step fails quickly like e.g. in https://github.com/keep-network/keep-ecdsa/issues/392. Single loop run is so quick that it drains disk space due to a number of logged messages. A short retry delay should be considered to be added.

We should also handle signature calculation retries too.

PR #409, opened 2020-04-17
Adding a delay of 1 sec in retry loops

dimpar:
Closes: https://github.com/keep-network/keep-ecdsa/issues/395

Adding a short delay interval of 1 sec. upon failure when calculating a signature and signer generation.

### DIFF
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -312,6 +312,7 @@ func (n *Node) publishSignature(
 				keepAddress.String(),
 				err,
 			)
+			time.Sleep(retryDelay)
 			continue
 		}
 		if hasSigningTimedOut {
@@ -332,6 +333,7 @@ func (n *Node) publishSignature(
 				keepAddress.String(),
 				err,
 			)
+			time.Sleep(retryDelay)
 			continue
 		}
 
@@ -360,6 +362,7 @@ func (n *Node) publishSignature(
 					keepAddress.String(),
 					err,
 				)
+				time.Sleep(retryDelay)
 				continue
 			}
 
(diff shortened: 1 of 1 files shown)


pkg/node/node.go
```diff
@@ -255,6 +257,7 @@ func (n *Node) CalculateSignature(
 				keepAddress.String(),
 				err,
 			)
+			time.Sleep(time.Second)
```
nkuba, line 260: Maybe we could extract the value to a constant defined on top of the file and reuse it?

dimpar: Hm, I guess I'm open, but what do you think we should name this constant? We use `time.Second` in 3 different places in slightly different contexts and it is already a constant. Naming `const second = time.Second` feels redundant. Smth general like `contextWaitInterval`?
Also we have similar case in `publishSignature` -> `time.Sleep(1 * time.Minute)`, but this one is used only in 1 place.

nkuba: All the three places are related, maybe we could name it just `retryDelay`?

The value in `publishSignature` is a different story and can remain as is.

dimpar: done

nkuba:
In the other PR we're adding yet another step with `continue` we will need to cover with a changes from this PR: https://github.com/keep-network/keep-ecdsa/pull/412/files#diff-440b0b0f55b1efa66d8a9a49cf4df99eR311

nkuba:
We should consider adding the delay in every retry loop we do `continue`. An elegant solution would be having exponential backoff proposed by @Shadowfiend (e.g. https://github.com/cenkalti/backoff).

For this PR we could start with a simple 1-second sleeps and TODOs left in the code. Then it could be replaced with the backoff.

nkuba:
@pdyraga @dimpar I introduced some updates to this PR. I would love to have you guys reviewed this PR.

Issue #392, opened 2020-04-13
Requested keep with group size 1 "group should have at least 2 members" failures loop


nkuba:
When requested keep with group size 1 it is accepted by the factory contract but the client fails in a loop with:
```
2020-04-13 08:01:04.270632 ERROR keep-ecdsa node.go:174: failed to generate threshold signer: [group should have at least 2 members but got: [1]]
```

To do:
- [x] update `BondedECDSAKeepFactory` to require `group size` > 0 and `threshold` > 0 - https://github.com/keep-network/keep-ecdsa/pull/410
- [ ] the client should check requested `groupSize` before starting the retry loop, log an error if group size < 2 (we will implement it proper way later in https://github.com/keep-network/keep-ecdsa/issues/408) and return - https://github.com/keep-network/keep-ecdsa/pull/415

Issue #408, opened 2020-04-17
Implement single signer support


nkuba:
A keep with a single member can be created, we need to implement single signer support in the client.

This issue is a followup from https://github.com/keep-network/keep-ecdsa/issues/392

PR #410, opened 2020-04-17
Requiring group size and threshold to be > 0

dimpar:
Refs: https://github.com/keep-network/keep-ecdsa/issues/392

It is not possible to create a group with `0` members. Similarly, it is not possible to create a group with `0` honest threshold. Added appropriate validations to `openKeep` contract functions.

### DIFF
--- a/solidity/contracts/BondedECDSAKeepFactory.sol
+++ b/solidity/contracts/BondedECDSAKeepFactory.sol
@@ -239,9 +239,12 @@ contract BondedECDSAKeepFactory is
         uint256 _bond,
         uint256 _stakeLockDuration
     ) external payable returns (address keepAddress) {
-        require(_groupSize > 0, "Minimum signing group size must be greater than 0");
+        require(_groupSize > 0, "Minimum signing group size is 1");
         require(_groupSize <= 16, "Maximum signing group size is 16");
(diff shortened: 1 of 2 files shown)


solidity/contracts/BondedECDSAKeepFactory.sol
```diff
@@ -239,7 +239,9 @@ contract BondedECDSAKeepFactory is
         uint256 _bond,
         uint256 _stakeLockDuration
     ) external payable returns (address keepAddress) {
+        require(_groupSize > 0, "Minimum signing group size must be greater than 0");
```
pdyraga, line 242: Maybe `Signing group size must be greater than 0` or `Minimum signing group size is 1`?

pdyraga:
@nkuba Is what we have here enough to close https://github.com/keep-network/keep-ecdsa/issues/392? Do we need to add a validation to the client to fail on 1-1 case and scenarios not currently supported by tss-lib?

pdyraga:
Ah, alright. I see the second checkbox in the issue:

> the client should check requested groupSize before starting the retry loop, log an error if group size < 2 (we will implement it proper way later in #408) and return.

Let's address it in a separate PR.

Issue #403, opened 2020-04-16
Check for signing request confirmations from the block it was requested


nkuba:
Currently, we check for signing request confirmation starting from the current block. This should be updated to the moment the signature has been actually requested. It's not much problem for received events, but if the client is started it checks if keep is awaiting signatures.

This change will require an update in the contract to track request time for digests. And aligning the client.

PR #411, opened 2020-04-17
Check for signing request confirmations from the block it was requested

nkuba:
Currently, we check for signing request confirmation starting from the current block when starting the client. This may cause the client to wait too long compared to other members. In this PR we store request's block number for each digest and use it as starting point for waiting for confirmations.

Closes https://github.com/keep-network/keep-ecdsa/issues/403


nkuba:
Thank you for the updates! Please merge it sir.

Issue #404, opened 2020-04-16
Abort signing when signing timeout has passed


nkuba:
A client doesn't check if signing process has timed out which leaves it in a retry loop.

Currently when client submits the signature it receives the message:
> 12:22:23.900 ERROR keep-ecdsa: failed to submit signature for keep [0xe7FB2CB75209F8FE246641125De06E6A44bAb6b6]: [got error [VM Exception while processing transaction: revert Signing timeout elapsed] while resolving original error [failed to estimate gas needed: VM Exception while processing transaction: revert Signing timeout elapsed]]; will retry after 1 minute node.go:359

The client should use keep's `hasSigningTimedOut` function to determine if the signing process should be abandoned in retry loops or when starting the signing process.

PR #412, opened 2020-04-17
Abort signing when signing timeout has passed

nkuba:
The client was not checking if signing has timed out so it was trying to publish the signature over and over. Here we update the client to check if the signing timeout has been hit in the contract so there is no point in publishing the signature or starting signature calculation on client start.

Closes: https://github.com/keep-network/keep-ecdsa/issues/404

### DIFF
--- a/pkg/chain/chain.go
+++ b/pkg/chain/chain.go
@@ -3,11 +3,12 @@
 package eth
 
 import (
+	"math/big"
+
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/keep-network/keep-common/pkg/subscription"
 	"github.com/keep-network/keep-core/pkg/chain"
 	"github.com/keep-network/keep-ecdsa/pkg/ecdsa"
-	"math/big"
 )
 
 // Handle represents a handle to an ethereum blockchain.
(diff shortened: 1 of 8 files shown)


pkg/node/node.go
```diff
@@ -295,7 +295,23 @@ func (n *Node) publishSignature(
...
+		hasSigningTimedOut, err := n.ethereumChain.HasSigningTimedOut(keepAddress)
+		if err != nil {
+			logger.Errorf(
+				"failed to verify if signing has timed out for keep [%s]: [%v]",
+				keepAddress.String(),
+				err,
+			)
+			continue
+		}
+		if !hasSigningTimedOut {
```
pdyraga, line 313:
```
if hasSigningTimedOut
```
should it be.

pkg/chain/chain.go
```diff
@@ -113,6 +114,9 @@ type BondedECDSAKeep interface {
 	// calculated for the given digest.
 	IsAwaitingSignature(keepAddress common.Address, digest [32]byte) (bool, error)
 
+	// HasSigningTimedOut checks if signing has timed out for keep.
```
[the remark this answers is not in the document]
pdyraga: Done in https://github.com/keep-network/keep-ecdsa/pull/412/commits/2c65e240dd15af752d1df2a85011886c96306e49

pkg/node/node.go
pdyraga: Done in https://github.com/keep-network/keep-ecdsa/pull/412/commits/755c8ba8e6432c0f458153df223cd12643b197c5

pdyraga:
Looks good to me now but I did two changes so I'll leave the final decision about merging to you @nkuba.

nkuba:
Oh yeah, thanks for the updates dude!

nkuba:
Let's ship it!

pdyraga:
Shippin'

PR #415, opened 2020-04-18
Catch unsupported keep size and threshold in client

nkuba:
Currently, the client supports keep size of at least 2 members and threshold equal to keep size. In this PR we catch requests with unsupported values, log an error and don't start the key generation process.

I also noticed that we were missing the threshold parameter passing to the event on keep creation or being able to read threshold of already created keep - fixed it in this PR.

Also added some parameters indexation for events as these may be helpful later for events filtering.

Closes: https://github.com/keep-network/keep-ecdsa/issues/392

### DIFF
--- a/pkg/chain/ethereum/ethereum.go
+++ b/pkg/chain/ethereum/ethereum.go
@@ -58,7 +58,7 @@ func (ec *EthereumChain) OnBondedECDSAKeepCreated(
 			handler(&eth.BondedECDSAKeepCreatedEvent{
 				KeepAddress:     KeepAddress,
 				Members:         Members,
-				HonestThreshold: HonestThreshold,
+				HonestThreshold: HonestThreshold.Uint64(),
 			})
 		},
 		func(err error) error {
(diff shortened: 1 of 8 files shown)


pkg/chain/ethereum/ethereum.go
```diff
@@ -431,6 +456,20 @@ func (ec *EthereumChain) GetMembers(
...
+// GetHonestThreshold returns keep's honest threshold.
+func (ec *EthereumChain) GetHonestThreshold(
+	keepAddress common.Address,
+) (*big.Int, error) {
+	keepContract, err := ec.getKeepContract(keepAddress)
+	if err != nil {
+		return nil, err
+	}
+
+	return keepContract.HonestThreshold()
```
pdyraga, line 468: What do you think about returning `uint` from this function? The comparison later can be easier. The maximum group size in the contract is 16 and threshold cannot be higher.

nkuba: I had this thought. I wasn't sure if we want to have the same type as in the contract or want to convert it. I'm fine with returning `uint` here.

pdyraga: We do it in the beacon for group size, threshold and all timeouts. Operating on big ints was too painful.

pdyraga: This is the only thing I'd change in this PR. Once we eliminate big int from here I am happy to merge.

Issue #356, opened 2020-04-04
Signature fraud should still be reportable if KEEP slashing fails


Shadowfiend:
Currently, `submitSignatureFraud` [calls](https://github.com/keep-network/keep-ecdsa/blob/3f2d73fe07e9f897ff0fb5e138ac7047a590aea0/solidity/contracts/BondedECDSAKeep.sol#L354) `slashSignerStakes`, which in turn asks the Keep factory to [slash the keep's members](https://github.com/keep-network/keep-ecdsa/blob/3f2d73fe07e9f897ff0fb5e138ac7047a590aea0/solidity/contracts/BondedECDSAKeep.sol#L587). The factory in turn asks the staking contract to [slash those stakes](https://github.com/keep-network/keep-ecdsa/blob/3f2d73fe07e9f897ff0fb5e138ac7047a590aea0/solidity/contracts/BondedECDSAKeepFactory.sol#L421). By design, the token staking contract only allows stake slashing if the [operator contract is approved](https://github.com/keep-network/keep-core/blob/master/solidity/contracts/TokenStaking.sol#L270). However, if the ECDSA keep is bonded, this bond should not be invalidated. Signature fraud reporting should still occur even if the stake slashing failed, so the consuming application can complete any bond-related actions.

Pending confirmation of whether we want to enable permanent approvals for operator contracts.

PR #416, opened 2020-04-18
Do not revert signature fraud submission TX if slashing failed

pdyraga:
Closes: https://github.com/keep-network/keep-ecdsa/issues/356

In practice, the situation when KEEP token slashing failed should never happen (see https://github.com/keep-network/keep-core/pull/1625). However, we want to be very explicit in this contract and protect the owner that even if it happens, the transaction submitting fraud proof is not going to fail and keep owner can seize and liquidate bonds in the same transaction.

### DIFF
--- a/solidity/test/BondedECDSAKeepTest.js
+++ b/solidity/test/BondedECDSAKeepTest.js
@@ -1189,8 +1189,6 @@ contract("BondedECDSAKeep", (accounts) => {
         "Caller is not the keep member"
       )
     })
-
-    it("does not fail if slashing failed", async () => {})
   })
 
   describe("closeKeep", () => {
--- a/solidity/test/BondedECDSAKeepTest.js
+++ b/solidity/test/BondedECDSAKeepTest.js
@@ -986,46 +986,60 @@ contract("BondedECDSAKeep", (accounts) => {
         )
       }
     })
-  })
 
-  describe("slashSignerStakes", async () => {
-    it("is not public", async () => {
-      try {
-        await keep.slashSignerStakes()
-      } catch (err) {
-        assert.equal(err, "TypeError: keep.slashSignerStakes is not a function")
-      }
-    })
-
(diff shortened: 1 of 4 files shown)


solidity/test/BondedECDSAKeepTest.js
```diff
@@ -1175,6 +1189,8 @@ contract("BondedECDSAKeep", (accounts) => {
         "Caller is not the keep member"
       )
     })
+
+    it("does not fail if slashing failed", async () => {})
```
nkuba, line 1193: Missing implementation?

solidity/test/BondedECDSAKeepTest.js
```diff
@@ -986,46 +986,60 @@ contract("BondedECDSAKeep", (accounts) => {
...
+          signature1.R,
+          signature1.S,
+          hash256Digest1,
+          preimage1
+        ),
+        "Keep is not active"
+      )
     })
 
-    it("slashes keep members stakes", async () => {
```
nkuba, line 1012: Are we removing this test on purpose?

solidity/test/BondedECDSAKeepTest.js
pdyraga: I had problems with linter, I do not know what happens here. I removed this line. Test for this scenario is implemented [here](https://github.com/keep-network/keep-ecdsa/blob/77b95f9e47b908dc30503e315aa04e80b2f0c8ce/solidity/test/BondedECDSAKeepTest.js#L1020-L1042).

solidity/test/BondedECDSAKeepTest.js
pdyraga: Yes, this is not covered [here](https://github.com/keep-network/keep-ecdsa/blob/77b95f9e47b908dc30503e315aa04e80b2f0c8ce/solidity/test/BondedECDSAKeepTest.js#L932-L959).

PR #418, opened 2020-04-19
Use simple delegate authority instead of claims

pdyraga:
Depends on https://github.com/keep-network/keep-core/pull/1628

When keep factory creates a keep it can delegate its authority to the keep directly, without having the keep claiming it. This solution is simpler and does not cause gas estimation problems like the claim path.


pdyraga:
We decided not to implement https://github.com/keep-network/keep-core/pull/1628 so I am closing this one without merging.

PR #421, opened 2020-04-20
keep-dev/test: Reduce unbonded deposit on init

sthompson22:
This is a backport for a change we made against the `v0.13.0` tag after it was deployed.

Reducing the unbonded deposit amount from 50 to 10, to reduce overall ETH need on each new deploy.


sthompson22:
@nkuba We need this one in before the next tag.

PR #427, opened 2020-04-22
Circle: Kill Docker image tagging

sthompson22:
This doesn't work the way we expect with tag builds, so I'm nuking it for now.  Will look at it again soon for a proper solution.

This one is really on me for not testing the tag build path with the original PR.


sthompson22:
@nkuba We should merge this before the 0.14.0 tag is cut.

PR #430, opened 2020-04-23
Pull 0.15.0 artifacts for keep-core

Shadowfiend:
Forgot to version-bump the keep-core dep in addition to the keep-ecdsa
version.


sthompson22:
(after builds are green)

PR #431, opened 2020-04-24
Rename magpie to beneficiary and Registry to KeepRegistry

eth-r:
In https://github.com/keep-network/keep-core/pull/1681 we renamed `magpie` to `beneficiary`.
In https://github.com/keep-network/keep-core/pull/1687 we renamed `Registry` to `KeepRegistry`.

Here we are adjusting `keep-ecdsa` to accommodate these changes.


pdyraga:
Double-checked with the client, looks good:

image

PR #432, opened 2020-04-28
Fixed from address for contract call options 

pdyraga:
Updated dependencies to the most recent versions of `keep-core` and `keep-common` containing https://github.com/keep-network/keep-common/pull/33 with a fix to `From` address in call options for generated contract bindings.

Tested with clients and smoke test, everything works as expected.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,8 +18,8 @@ require (
 	github.com/gogo/protobuf v1.3.1
 	github.com/google/gofuzz v1.1.0
 	github.com/ipfs/go-log v0.0.1
-	github.com/keep-network/keep-common v0.3.0-rc
-	github.com/keep-network/keep-core v0.14.0-rc
+	github.com/keep-network/keep-common v0.3.0-rc.0.20200427145435-e24eeb9599e4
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -18,8 +18,8 @@ require (
 	github.com/gogo/protobuf v1.3.1
 	github.com/google/gofuzz v1.1.0
 	github.com/ipfs/go-log v0.0.1
-	github.com/keep-network/keep-common v0.3.0-rc
-	github.com/keep-network/keep-core v0.14.0-rc
+	github.com/keep-network/keep-common v0.3.0-rc.0.20200427145435-e24eeb9599e4
```
nkuba, line 21: Do we want to update this reference to a commit merging the actual PR to the master? https://github.com/keep-network/keep-common/commit/07d2a79e2e9815f8e7ed361a6f18801594879e28

nkuba: Or maybe we should provide a new tag in `keep-common`?

pdyraga: We'll do a new tag before `keep-core` release and then we'll update. We have to do the same in `keep-core`.

nkuba: 🆒

PR #436, opened 2020-04-30
Bumping up contracts version to 0.16.0-pre

pdyraga:
After tagging `0.15.0-rc` it's time to bump up the version to `-pre`.


pdyraga:
We are at 1.0 now so this one is no longer applicable.

PR #437, opened 2020-05-02
Load balancer

eth-r:
Currently includes a basic design for how we could implement a transparent load balancer factory.

Implements subfactories that let the load balancer factory pass the calling application in an argument, and provide a function to query the weight of the sortition pool of a given application.

The load balancer has one or two subfactories that it divides work between, according to the `_choiceFunction`.


pdyraga:
Replaced by https://github.com/keep-network/tbtc/pull/610

PR #441, opened 2020-05-06
Expose getSortitionPoolWeight in keep factory

lukasz-zimnoch:
<s>Depends on https://github.com/keep-network/sortition-pools/pull/80</s>

Here we expose getSortitionPoolWeight in BondedECDSAKeepFactory contract to allow getting the weight of the underlying sortition pool.

### DIFF
--- a/solidity/contracts/BondedECDSAKeepFactory.sol
+++ b/solidity/contracts/BondedECDSAKeepFactory.sol
@@ -480,6 +480,24 @@ contract BondedECDSAKeepFactory is
         return tokenStaking.balanceOf(_operator);
     }
 
+    /// @notice Gets the total weight of operators
+    /// in the sortition pool for the given application.
+    /// @dev Reverts if sortition does not exits for the application.
+    /// @param _application Address of the application.
+    /// @return The sum of all registered operators' weights in the pool.
+    /// If no pool exists, returns 0.
+    function getSortitionPoolWeight(
+        address _application
+    ) public view returns (uint256) {
+        address poolAddress = candidatesPools[_application];
+
+        if (poolAddress == address(0)) {
+            return 0;
+        }
+
+        return BondedSortitionPool(poolAddress).totalWeight();
+    }
+
     /// @notice Gets bonded sortition pool of specific application for the
     /// operator.
     /// @dev Reverts if the operator is not registered for the application.
--- a/solidity/contracts/BondedECDSAKeepFactory.sol
+++ b/solidity/contracts/BondedECDSAKeepFactory.sol
@@ -486,15 +486,14 @@ contract BondedECDSAKeepFactory is
     /// @param _application Address of the application.
     /// @return The sum of all registered operators' weights in the pool.
     /// Reverts if sortition pool for the application does not exist.
-    function getSortitionPoolWeight(
-        address _application
-    ) public view returns (uint256) {
+    function getSortitionPoolWeight(address _application)
+        public
+        view
+        returns (uint256)
+    {
         address poolAddress = candidatesPools[_application];
 
-        require(
-            poolAddress != address(0),
-            "No pool found for the application"
-        );
+        require(poolAddress != address(0), "No pool found for the application");
 
         return BondedSortitionPool(poolAddress).totalWeight();
     }
--- a/solidity/contracts/BondedECDSAKeepFactory.sol
+++ b/solidity/contracts/BondedECDSAKeepFactory.sol
@@ -491,9 +491,10 @@ contract BondedECDSAKeepFactory is
     ) public view returns (uint256) {
         address poolAddress = candidatesPools[_application];
 
-        if (poolAddress == address(0)) {
-            return 0;
-        }
+        require(
+            poolAddress != address(0),
+            "No pool found for the application"
+        );
 
         return BondedSortitionPool(poolAddress).totalWeight();
     }
--- a/solidity/contracts/BondedECDSAKeepFactory.sol
+++ b/solidity/contracts/BondedECDSAKeepFactory.sol
@@ -485,7 +485,7 @@ contract BondedECDSAKeepFactory is
     /// @dev Reverts if sortition does not exits for the application.
     /// @param _application Address of the application.
     /// @return The sum of all registered operators' weights in the pool.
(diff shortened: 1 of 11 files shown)


solidity/contracts/api/IBondedECDSAKeepFactory.sol
```diff
@@ -22,4 +22,12 @@ interface IBondedECDSAKeepFactory {
...
     /// @return Uint256 estimate.
     function openKeepFeeEstimate() external view returns (uint256);
+
+    /// @notice Gets the total weight of operators
+    /// in the sortition pool for the given application.
+    /// @param _application Address of the application.
+    /// @return The sum of all registered operators' weights in the pool.
+    function getSortitionPoolWeight(
+        address _application
+    ) external view returns (uint256 poolWeight);
```
pdyraga, line 32: Let's just do `returns (uint256)`, the function is `getSortitionPoolWeight` so it's clear what are we returning here. It's the same case as with `openKeepFeeEstimate`

solidity/contracts/BondedECDSAKeepFactory.sol
```diff
@@ -480,6 +480,24 @@ contract BondedECDSAKeepFactory is
...
+    /// @return The sum of all registered operators' weights in the pool.
+    /// If no pool exists, returns 0.
+    function getSortitionPoolWeight(
+        address _application
+    ) public view returns (uint256) {
+        address poolAddress = candidatesPools[_application];
+
+        if (poolAddress == address(0)) {
+            return 0;
+        }
```
pdyraga, line 496: I think we should rather revert if the pool does not exist:
```
require(
            candidatesPools[_application] != address(0),
            "No pool found for the application"
        );
```

We'll have this behavior consistent with `getSortitionPool`.

pdyraga: Let's also make sure to convert the two possible paths in this function in unit tests: https://github.com/keep-network/keep-ecdsa/blob/master/solidity/test/BondedECDSAKeepFactoryTest.js

eth-r: Reverting if the pool doesn't exist can cause some problems with the load balancer strategy if the pool hasn't been created. If we're only using it in an application-specific way this is fine. Also, the `getSortitionPoolWeight` function doesn't necessarily need to be exposed here and we can query `getSortitionPool`, handling errors there if relevant, and then directly call the sortition pool to get total weight.

pdyraga: > Reverting if the pool doesn't exist can cause some problems with the load balancer strategy if the pool hasn't been created. If we're only using it in an application-specific way this is fine.

I think this is what we expect. My intuition is that the load balancer is an application-specific strategy, it has a reference to two ECDSA keep factories and the load balancer should expect those factories to be initialized. If they are not, it means something is wrong.

Regarding `getSortitionPoolWeight`, I kinda like we return just `uin256` because the application does not have to import `SortitionPool` that is kinda internal to how keep is doing group selection. Do you think just the weight is not going to be enough?

eth-r: The weight is enough in realistic situations. There are some cases (whales with lots of KEEP but not much bondable ETH) where the weight would give false positives wrt the ability to select KEEP-backed members but those would be hard to account for anyway and we can just treat them as situations where you won't be able to create a new keep and fail in some reasonable way.

pdyraga: Yep and the good part of having the load balancing strategy on the application side is that the application may `CALL` the selected factory and if that call fails, it may fallback to the other factory if that's what the application wants to do. We are not doing any assumptions about that on Keep side.

lukasz-zimnoch:
@pdyraga ready for another chance!

pdyraga:
@lukasz-zimnoch You need to `lint:fix:js`

lukasz-zimnoch:
> @lukasz-zimnoch You need to `lint:fix:js`

Ok, done!

solidity/integration/smoke_test.js
```diff
@@ -1,12 +1,7 @@
-const BondedECDSAKeepVendor = artifacts.require("./BondedECDSAKeepVendor.sol")
-const BondedECDSAKeepVendorImplV1 = artifacts.require(
-  "./BondedECDSAKeepVendorImplV1.sol"
-)
 const BondedECDSAKeepFactory = artifacts.require("./BondedECDSAKeepFactory.sol")
 const BondedECDSAKeep = artifacts.require("./BondedECDSAKeep.sol")
 
 const RandomBeaconService = artifacts.require("IRandomBeacon")
-const {RandomBeaconAddress} = require("../migrations/external-contracts")
```
pdyraga, line 5: Are we sure we want to modify this file? It looks like the linter went crazy. I don't think it's going to work with the changes here.

solidity/scripts/lcl-initialize.js
```diff
@@ -5,11 +5,6 @@ const TokenStaking = artifacts.require(
   "@keep-network/keep-core/build/truffle/TokenStaking"
 )
 
-const {
-  TokenStakingAddress,
-  TBTCSystemAddress,
-} = require("../migrations/external-contracts")
-
```
pdyraga, line 12: Same here, I think this will not going to work without those lines.

solidity/integration/smoke_test.js
lukasz-zimnoch: Hmm they were marked as unused and I didn't see any usage in fact. But OK, I will restore them.

solidity/scripts/lcl-initialize.js
lukasz-zimnoch: Same as above

pdyraga:
@lukasz-zimnoch I had to merge master and approve linting, please take a look at the changes and confirm I did not mess something up. If everything looks good to you, I am good to merge.

lukasz-zimnoch:
> @lukasz-zimnoch I had to merge master and approve linting, please take a look at the changes and confirm I did not mess something up. If everything looks good to you, I am good to merge.

Yep, everything looks good!

Issue #438, opened 2020-05-05
sha256 vs keccak


wjmelements:
Why do you use sha256 instead of keccak256?

Shadowfiend:
Bonded ECDSA keeps, in their first incarnation, are pretty strongly oriented towards backing tBTC deposits, so they're built with BTC parameters in mind. As a result, we sha256 preimages to get the digest for signatures. Bitcoin uses double-sha256, but to maintain a *little* more generality, ECDSA keeps use single sha256, and let the caller produce the preimage from the first sha256 external to the keep itself.

Supporting different digest functions for preimages is something we want to explore in future versions, particularly if we have specific use cases in mind, so we'd love to hear it if you have one!

nkuba:
Closing this issue due to inactivity. Please feel free to reopen if followup is needed.

PR #443, opened 2020-05-06
Removed double sha256 mention

nkuba:
We no longer require double sha256 as per #217. We missed updating the doc for interface and error message. Here we fix this oversight.


Shadowfiend:
h/t to @wjmelements whose question in #438 made me realize we still mentioned double SHA256s in a couple of places.

PR #445, opened 2020-05-07
v0.17.0-pre

sthompson22:
There's a few things we need to sort here with migrations for keep-core being disabled on the keep-dev workflow.  Going ahead and opening this while we sort it out.

### DIFF
--- a/solidity/package.json
+++ b/solidity/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@keep-network/keep-ecdsa",
-  "version": "0.15.0-pre",
+  "version": "0.17.0-pre",
   "description": "Smart contracts for ECDSA Keep",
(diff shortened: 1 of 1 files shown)


pdyraga:
I think we can close #436 with this change. We need to resolve conflicts and make sure lock is in sync - build is complaining about it.

sthompson22:
This is what we need to talk about - I can't update the lock because we disabled NPM publishing in the `keep-dev` workflow on `keep-core`.

sthompson22:
We need to either reenable NPM  publishing even though we aren't migrating in keep-dev anymore, to keep this versioning dance going, or come up with something else.

pdyraga:
We are at 1.0 now so this one is no longer applicable.

PR #449, opened 2020-05-13
1.0.0 preparation

pdyraga:
Versions updated:
- `keep-core` go dependency updated to `v1.1.3`. This is the latest release of the client available.
- `keep-common` go dependency updated to `v1.0.0`. This is the lastest version of the library available.
- `sortition-pools` contracts dependency updated to `v1.0.0`. This is the latest version of the library available.
- `keep-core` contracts dependency updated to `v1.1.2`. This is the latest version of contracts available.

There are two additional changes I had to make because of `keep-common` update:
- We are instantiating the nonce manager and use it for all contracts.
- We use the first 20 bytes of member ID when constructing the file name to respect 128 characters limit.

Finally, bumped up contracts version to `1.0.0` :tada:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,7 +18,7 @@ require (
 	github.com/gogo/protobuf v1.3.1
 	github.com/google/gofuzz v1.1.0
 	github.com/ipfs/go-log v0.0.1
-	github.com/keep-network/keep-common v0.3.2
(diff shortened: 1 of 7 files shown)


pdyraga:
Tested:
- smoke test several times
- signing with existing keep with client restarts
- closing keep and archiving storage data

PR #453, opened 2020-05-15
Figment bootstrap nodes

mattharrop:
Add figment seed nodes


pdyraga:
Thank you @mattharrop 🚀

PR #463, opened 2020-05-19
Use alpine as the runtime container since Go binaries are self-contained

alphastorm:
This cuts the final Docker image size down by 90% from 390MB (v1.1.0) to 36.5MB.

Same thing as https://github.com/keep-network/keep-core/pull/1811/files.


lukasz-zimnoch:
Works good!

PR #469, opened 2020-05-27
Reduce instance type recommendations

sthompson22:
After spending some time doing targeted testing with single client deployments across providers it feels safe to lower the instance type recommendations here.  Worth noting that depending on how you deploy your keep-ecdsa-client, particularly if you're in a containerized environment
sharing VM resources across multiple applications, be sure to take that into consideration when selecting a machine.


docs/run-keep-ecdsa.adoc
```diff
@@ -20,22 +20,27 @@ these capabilities consider the following criteria:
...
-|n1-standard-4
+|n1-standard-2
 
 |AWS
-|m5.xlarge
+|t3.medium
 
 |Azure
-|D4s v3
+|A2 v2
```
nkuba, line 40: For our keep network use, `F2s v2` may be a good fit. F series is compute-optimized, hence could handle messages processing peaks and prime numbers generation better. F is slightly cheaper than A and I believe it has broader availability. A series is recommended for dev/test usage.

docs/run-keep-ecdsa.adoc
```diff
@@ -20,22 +20,27 @@ these capabilities consider the following criteria:
...
 
 |Self-hosted
-|4 vCPU / 4 GiB RAM / 1 GiB Persistent Storage
+|2 vCPU / 4 GiB RAM / 1 GiB Persistent Storage
```
nkuba, line 43: I think we can reduce memory requirements pretty safely to 2 GB. We may be able to reduce it even more but it requires more monitoring of nodes performance with higher network usage.

docs/run-keep-ecdsa.adoc
sthompson22: Having a look.

docs/run-keep-ecdsa.adoc
sthompson22: Have you  run testnet nodes on 2 reliably?

docs/run-keep-ecdsa.adoc
sthompson22: Yeah this looks appropriate, making the update - also double  checking the `c` class on AWS.

sthompson22: Updated both  - prices were close to what was set.   There's some other options for GCP  but need to  make sure 2gb is safe.

docs/run-keep-ecdsa.adoc
nkuba: Yes, I've been also running them on 1GB pretty stable but don't want to recommend it just yet. 2 GB should be more than enough.

sthompson22: Great thanks, updated.

sthompson22:
@nkuba Can you give this a look/merge, all is green?

sthompson22:
Having this more up to date with the stakedrop influx would be awesome.

PR #470, opened 2020-05-30
Allow grantee withdraw unbonded value from KeepBonding

pdyraga:
Depends on: https://github.com/keep-network/keep-core/pull/1834
Once https://github.com/keep-network/keep-core/pull/1834 is merged and new `1.2.0-pre` version of `keep-core` is published to NPM registry, please update `"@keep-network/keep-core" version referenced in `package.json`.

The previous construction allowed only the operator or token owner to withdraw the unbonded value. We were not allowing the token grantee to do that.

There are four valid cases for unbonded value withdrawal:
- operator withdraws,
- token owner withdraws (when the delegated stake are tokens owned by the delegating party),
- grantee withdraws (when the delegated stake are tokens granted to the delegating party),
- managed grantee withdraws (when the delegated stake are tokens granted to the delegating party via `ManagedGrant` contract).

All those cases are now supported in `KeepBonding` contract. The last case requires separate function since it is not possible to lookup `ManagedGrant` contract address in another way than by using an event.

### DIFF
--- a/solidity/contracts/KeepBonding.sol
+++ b/solidity/contracts/KeepBonding.sol
@@ -356,6 +356,9 @@ contract KeepBonding {
         return authorizedPools[_operator][_poolAddress];
     }
 
+    /// @notice Withdraws the provided amount from unbonded value of the
+    /// provided operator to operator's beneficiary. If there is no enough
+    /// unbonded value or the transfer failed, function fails.
     function withdrawBond(uint256 amount, address operator) internal {
         require(
             unbondedValue[operator] >= amount,
--- a/solidity/contracts/KeepBonding.sol
+++ b/solidity/contracts/KeepBonding.sol
@@ -16,6 +16,8 @@ pragma solidity 0.5.17;
 
 import "@keep-network/keep-core/contracts/KeepRegistry.sol";
 import "@keep-network/keep-core/contracts/TokenStaking.sol";
+import "@keep-network/keep-core/contracts/TokenGrant.sol";
+import "@keep-network/keep-core/contracts/libraries/RolesLookup.sol";
 
 import "openzeppelin-solidity/contracts/math/SafeMath.sol";
 
@@ -24,12 +26,16 @@ import "openzeppelin-solidity/contracts/math/SafeMath.sol";
 /// @notice Contract holding deposits from keeps' operators.
 contract KeepBonding {
     using SafeMath for uint256;
+    using RolesLookup for address payable;
 
     // Registry contract with a list of approved factories (operator contracts).
     KeepRegistry internal registry;
 
     // KEEP token staking contract.
     TokenStaking internal tokenStaking;
+
+    // KEEP token grant contract.
+    TokenGrant internal tokenGrant;
 
     // Unassigned value in wei deposited by operators.
     mapping(address => uint256) public unbondedValue;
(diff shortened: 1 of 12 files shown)


solidity/contracts/KeepBonding.sol
```diff
@@ -329,4 +355,20 @@ contract KeepBonding {
     {
         return authorizedPools[_operator][_poolAddress];
     }
+
+    function withdrawBond(uint256 amount, address operator) internal {
```
nkuba, line 359: Can we add docs for this function?

pdyraga: Done.

solidity/contracts/KeepBonding.sol
```diff
@@ -109,29 +121,43 @@ contract KeepBonding {
     }
 
     /// @notice Withdraws amount from operator's value available for bonding.
-    /// Can be called only by the operator or by the stake owner.
+    /// Can be called only by:
+    /// - operator
+    /// - staked tokens owner
```
nkuba, line 126: It's not possible in case of a token grant. Can we update docs to be explicit about cases that are supported?

nkuba:
It works fine. I tested the scenarios listed below with Keep Dashboard changes from https://github.com/keep-network/keep-core/pull/1835. 

I. Bonding withdraw without delegation - https://github.com/keep-network/keep-ecdsa/pull/478

II. Bonding withdraw with a delegation

a. Delegated by Owner

token owner:	`0x3d373D872B7BA29d92Ed47CAA8605b4DD6Ec84eF`
operator: 	`0xF0002dFf3113400f953DFc2F341129f8D4F40650`
beneficiary: 	`0xD4aF650b6664059C55D1abdE5156b08884318c06`
authorizer: 	`0xdf42641326140B45e23E178aE6FAaFF08E48cdEa`

withdraw by:
- token owner - ✅ 
- operator - ✅ 


b. Delegated by grantee (Grant - TokenGrant.receiveApproval)
token owner: 				`0x3d373D872B7BA29d92Ed47CAA8605b4DD6Ec84eF`

grantee: 			`0xaaB0ca8E2C3d86120b5A840E0006d18E6ea84000`
operator: 		`0xC21cB9C99C316d1863142F7dD86dd5496D81A8D6`
beneficiary: 	`0x86AD7e828F7DDd39D4Ac9Ec2050FA1578418aDC3`
authorizer: 	`0x68ad60CC5e8f3B7cC53beaB321cf0e6036962dBc`


withdraw by:
- grantee - ✅ 
- operator - ✅ 
- token owner - not allowed


c. Delegated by grantee (ManagedGrant - ManagedGrantFactory.createManagedGrant)
token owner: 				`0x3d373D872B7BA29d92Ed47CAA8605b4DD6Ec84eF`

grantee: 			`0xfe04e65Ae59BAbdf862EcA1ADc287F2B1C3075e7`
operator: 		`0x45B756a5f662c4f99Bef0397a55f94DDDeF4C62E`
beneficiary: 	`0xfd42E065d5b128a65d752F11A1dA089f74DBf49A`
authorizer: 	`0x1C948F07dF428E79c5821D6215b66269c286aAaA`

withdraw by:
- managed grantee - ✅ 
- operator - ✅ 
- token owner - not allowed

solidity/contracts/KeepBonding.sol
pdyraga: Please let me know what do you think about this:
```
Withdraws amount from operator's value available for bonding.
Should not be used by grantee of managed grants. For this case,
please use `withdrawAsManagedGrantee` instead.

This function can be called only by: 
- operator,
- liquid, staked tokens owner (not a grant),
- direct staked tokens grantee (not a managed grant).
```

nkuba: Looks fine! Thanks.

PR #472, opened 2020-05-30
Added KEEP asciiart to all contracts

pdyraga:
All contracts KEEP asciiart.

I have also removed an old TODO in `KeepBonding`. Repositories are not merged for mainnet and once we merge them, it's questionable if the requested feature will be a good move. I have doubts now so instead of having an old TODO there, let's give ourselves a chance to revisit package organization. Also, this TODO looks terrible on Etherscan.


pdyraga:
@nkuba I rebased it onto the recent `master`.

PR #473, opened 2020-05-30
Include distributed rewards amount in events

pdyraga:
It is gas-cheap but should help with splitting rewards in case someone uses third party staking provider for running nodes.


solidity/contracts/BondedECDSAKeep.sol
```diff
@@ -447,7 +447,7 @@ contract BondedECDSAKeep is IBondedECDSAKeep {
             dividend.add(remainder)
         );
 
-        emit ERC20RewardDistributed();
+        emit ERC20RewardDistributed(_value);
```
nkuba, line 450: I'll also add token address to the event.

nkuba: 9b9893f716fe6d87ffd557978e80ac8c1ad9dab9

PR #476, opened 2020-06-04
Update Run-keep-ecdsa.doc

afmsavage:
Removed the final comma in the bootstrap peers list.  If a user doesn't have a linter to catch it, this will cause their node to fail.


Shadowfiend:
Hmm... TOML supports trailing commas I thought? I’ll try to give this a test whirl to confirm it’s broken as is and merge if so.

afmsavage:
I got some errors when I tried with it @Shadowfiend , but I'm not too familiar with TOML.  I can retest when the new testnet nodes come out later today as well and report back.

sthompson22:
> I got some errors when I tried with it @Shadowfiend , but I'm not too familiar with TOML. I can retest when the new testnet nodes come out later today as well and report back.

Is this still an issue for you @afmsavage ?

afmsavage:
> > I got some errors when I tried with it @Shadowfiend , but I'm not too familiar with TOML. I can retest when the new testnet nodes come out later today as well and report back.
> 
> Is this still an issue for you @afmsavage ?

Nope, you're right it is valid. I must have made a mistake somewhere else in my config that was causing the issue. Close away

PR #478, opened 2020-06-10
Require beneficiary to be set for withdraw

nkuba:
In this PR we update ETH withdraws with checks that a beneficiary is set for the operator. 

Previously it was possible that bonding value was deposited for the operator before stake delegation happened, so the beneficiary was not set yet. In such a case, the operator was able to withdraw the unbonded value, but the `tokenStaking.beneficiaryOf` returned zero address, to which funds could be transferred. 

Here we added verification if the beneficiary is defined. If not the transaction will revert.

### DIFF
--- a/solidity/contracts/BondedECDSAKeep.sol
+++ b/solidity/contracts/BondedECDSAKeep.sol
@@ -485,14 +485,10 @@ contract BondedECDSAKeep is IBondedECDSAKeep {
 
         memberETHBalances[_member] = 0;
 
-        address beneficiary = tokenStaking.beneficiaryOf(_member);
-        require(
-            beneficiary != address(0),
-            "Beneficiary not defined for the operator"
-        );
-
         /* solium-disable-next-line security/no-call-value */
-        (bool success, ) = beneficiary.call.value(value)("");
+        (bool success, ) = tokenStaking.beneficiaryOf(_member).call.value(
+            value
+        )("");
 
         require(success, "Transfer failed");
     }
--- a/solidity/contracts/BondedECDSAKeep.sol
+++ b/solidity/contracts/BondedECDSAKeep.sol
@@ -485,10 +485,14 @@ contract BondedECDSAKeep is IBondedECDSAKeep {
 
         memberETHBalances[_member] = 0;
 
+        address beneficiary = tokenStaking.beneficiaryOf(_member);
+        require(
+            beneficiary != address(0),
+            "Beneficiary not defined for the operator"
+        );
+
         /* solium-disable-next-line security/no-call-value */
-        (bool success, ) = tokenStaking.beneficiaryOf(_member).call.value(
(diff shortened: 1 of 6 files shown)


solidity/contracts/BondedECDSAKeep.sol
```diff
@@ -485,10 +485,14 @@ contract BondedECDSAKeep is IBondedECDSAKeep {
 
         memberETHBalances[_member] = 0;
 
+        address beneficiary = tokenStaking.beneficiaryOf(_member);
+        require(
+            beneficiary != address(0),
+            "Beneficiary not defined for the operator"
+        );
```
pdyraga, line 492: Do we need this check here? I think it's not possible to happen. To become a keep member, stake delegation has to exist and the beneficiary has to be defined. If we want to guard against `0x0` as a beneficiary, it should be done either in the dApp or in the staking contract.

solidity/contracts/KeepBonding.sol
```diff
@@ -371,11 +375,15 @@ contract KeepBonding {
 
         unbondedValue[operator] = unbondedValue[operator].sub(amount);
 
-        (bool success, ) = tokenStaking.beneficiaryOf(operator).call.value(
-            amount
-        )("");
+        address beneficiary = tokenStaking.beneficiaryOf(operator);
+        require(
```
pdyraga, line 379: Can we add a comment for future-us that this situation happens only when someone deposited ETH in `KeepBonding` before defining the beneficiary?

solidity/contracts/BondedECDSAKeep.sol
nkuba: 3d2c4513add72429288c48f48b59527aad246fb1

solidity/test/BondedECDSAKeepTest.js
```diff
@@ -1612,11 +1616,23 @@ contract("BondedECDSAKeep", (accounts) => {
...
+    it("reverts in case of not defined beneficiary", async () => {
+      const member = members[0]
+
+      await tokenStaking.setBeneficiary(member, constants.ZERO_ADDRESS)
+
+      await expectRevert(
+        keep.withdraw(member),
+        "Beneficiary not defined for the operator"
+      )
+    })
```
pdyraga, line 1628: Do we still need this guy? I think this functionality was removed.

pdyraga: Yeah, just checked locally and this test is failing now.

nkuba:
@pdyraga it's ready

Shadowfiend:
> Previously it was possible that bonding value was deposited for the operator before stake delegation happened, so the beneficiary was not set yet.

Do we want to fix this by preventing *withdrawal* for an unset beneficiary, or by preventing *bonding*?

PR #483, opened 2020-06-17
ETH Delegating

nkuba:
This PR implements ETH Delegating that will be integrated into ECDSA Keeps.

`ETHDelegating` contract is a very basic draft of the contract that will evolve as we integrate it with Keep Factory. We need delegation details like: `beneficiary` and `authorizer` for the given `operator`. The idea is that the owner will call `delegate` function providing these details so they can be later used for bonding.

TODO:
- [x] Determine mainnet value for minimum value to be deposited on delegation
- [x] Decide whether we want to have a setter for the value above or just hardcode it in the contract

Depends on https://github.com/keep-network/keep-core/pull/1873
Depends on https://github.com/keep-network/keep-ecdsa/pull/525
Depends on https://github.com/keep-network/sortition-pools/pull/89


mhluongo:
@nkkuba should this be against `no-beneficiary-withdraw`?

solidity/contracts/AbstractBonding.sol
```diff
@@ -0,0 +1,357 @@
...
+import "@keep-network/keep-core/contracts/KeepRegistry.sol";
+import "@keep-network/keep-core/contracts/Authorizations.sol";
+import "@keep-network/keep-core/contracts/StakeDelegatable.sol";
+import "@keep-network/sortition-pools/contracts/api/IBonding.sol";
+import "openzeppelin-solidity/contracts/math/SafeMath.sol";
+
+
+/// @title Keep Bonding
+/// @notice Contract holding deposits from keeps' operators.
+contract AbstractBonding is IBonding {
```
pdyraga, line 26: Would it be possible to extract `AbstractBonding` into a separate PR? I'd love to merge it quickly, I don't want to block on the rest of the code.

nkuba: https://github.com/keep-network/keep-ecdsa/pull/525

solidity/contracts/ETHStaking.sol
```diff
@@ -0,0 +1,65 @@
...
+        require(
+            operators[operator].owner == address(0),
+            "Operator already in use"
+        );
+
+        operators[operator] = Operator(0, _from, beneficiary, authorizer);
+        ownerOperators[_from].push(operator);
+
+        emit Staked(_from, operator, beneficiary, authorizer);
+    }
```
pdyraga, line 64: Worth noting, we'll need to have an initialization period here similar to how we do it for KEEP staking. Maybe you already have it in plans, just calling it out.

nkuba: Good point. We may need to handle it in a separate PR as this may require updates to FullyBackedSortitionPool.

solidity/contracts/ETHStaking.sol
```diff
@@ -0,0 +1,66 @@
...
+    /// addresses of operator, beneficiary and authorizer for a stake from the
+    /// caller.
+    /// @param operator Address of the operator.
+    /// @param beneficiary Address of the beneficiary.
+    /// @param authorizer Address of the authorizer.
+    function stake(
+        address operator,
+        address payable beneficiary,
+        address authorizer
+    ) external payable {
```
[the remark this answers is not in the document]
pdyraga: We discussed it with @nkuba and we agreed on adding a minimum bond expectation here as well sending the entire `msg.value` to keep bonding, allowing the user to set up the delegation along with a bond in one transaction.

solidity/contracts/ETHStaking.sol
```diff
@@ -0,0 +1,66 @@
...
+        // TODO: Do we need to verify the caller in any way?
```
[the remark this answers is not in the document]
nkuba: I was thinking about it in the context of an attack when an attacker calls this function before the owner.

solidity/contracts/ETHStaking.sol
```diff
@@ -0,0 +1,66 @@
...
+        address _from = msg.sender;
```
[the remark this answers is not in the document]
nkuba: copy-paste, will change to `staker`

solidity/contracts/ETHBonding.sol
```diff
@@ -0,0 +1,70 @@
...
+*/
+
+pragma solidity 0.5.17;
+
+import "./AbstractBonding.sol";
+import "./ETHStaking.sol";
+
+/// @title ETH Bonding
+/// @notice Contract holding deposits from keeps' operators.
+contract ETHBonding is AbstractBonding {
```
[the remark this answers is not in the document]
nkuba: Sure, I will name it consistently. Although my personal preference is `KEEPBonding` and `ETHBonding`.

solidity/contracts/ETHStaking.sol
```diff
@@ -0,0 +1,66 @@
...
+import "@keep-network/keep-core/contracts/KeepRegistry.sol";
+import "@keep-network/keep-core/contracts/Authorizations.sol";
+import "@keep-network/keep-core/contracts/StakeDelegatable.sol";
+
+import "openzeppelin-solidity/contracts/math/SafeMath.sol";
+
+/// @title ETH Staking
+/// @notice A staking contract for ETH staking. An owner of the ETH can delegate
+/// ETH as a stake to an operator. The value of ETH the owner is willing to stake
+/// should be deposited in `ETHBonding` contract for the given operator.
```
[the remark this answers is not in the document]
nkuba: I'm not sure about how to name this contract, I had ideas like `ETHStaking` and `ETHDelegating`. Decided on `ETHStaking` to be consistent with `TokenStaking`. 
It is not strictly about staking as in the token staking, but on the other hand, we are naming the functionality ETH-only staking, which means we stake ETH. The bond will become a stake.

We could combine ETHBonding and ETHStaking contracts but the resulting contract may get too big, too complex and too confusing at some point.

It's important for us to decide on the naming now as this will impact other places where we want to use this contract.

Waiting for your thoughts and suggestions.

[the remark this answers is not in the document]
pdyraga: I'd probably go with `EthDelegation`.

`TokenStaking` is about KEEP token staking and we are staking neither KEEP nor ETH in `EthStaking`. What we do instead, is that we are defining *a delegation* of ETH - from owner to operator, using the given beneficiary and authorizer.

solidity/contracts/ETHStaking.sol
```diff
@@ -0,0 +1,66 @@
...
+        // TODO: Do we need to verify the caller in any way?
+        address _from = msg.sender;
+
+        require(
+            operators[operator].owner == address(0),
+            "Operator already in use"
+        );
+
+        operators[operator] = Operator(0, _from, beneficiary, authorizer);
+        ownerOperators[_from].push(operator);
```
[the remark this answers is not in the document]
nkuba: 6bfe87683605df2186e4e50017b8bec86bb21751

solidity/contracts/ETHStaking.sol
```diff
@@ -0,0 +1,66 @@
...
+        public
+        Authorizations(keepRegistry)
+    {}
+
+    event Staked(
+        address owner,
+        address indexed operator,
+        address indexed beneficiary,
+        address indexed authorizer
+    );
```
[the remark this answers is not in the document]
nkuba: 6bfe87683605df2186e4e50017b8bec86bb21751

solidity/contracts/ETHBonding.sol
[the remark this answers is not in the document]
nkuba: 96ecec0194d5ed997ec66ecabff200717618f05f

solidity/contracts/ETHBonding.sol
```diff
@@ -0,0 +1,70 @@
...
+    ETHStaking ethStaking;
+
+    /// @notice Initializes Keep Bonding contract.
```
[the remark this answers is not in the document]
nkuba: 96ecec0194d5ed997ec66ecabff200717618f05f

solidity/contracts/ETHStaking.sol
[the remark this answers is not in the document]
nkuba: Thoughts on `EthDelegating` or `EthDelegations`? To me `EthDelegation`suggests a single delegation unit, something appropriate for a structure or an event.

`EthDelegating` seems to be consistent with `TokenStaking`, `KeepBonding` and `EthBonding`.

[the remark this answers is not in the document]
pdyraga: I ❤️ `EthDelegating`, it's consistent with all the other contract names you mentioned.

solidity/contracts/ETHStaking.sol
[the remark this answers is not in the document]
nkuba: I decided to rename it to `owner` as we're no longer talking about staking in the contract.
a89b18a5064108ddacc406457efc029d9afbf025

solidity/contracts/ETHBonding.sol
```diff
@@ -0,0 +1,70 @@
...
+                           Trust math, not hardware.
+*/
+
+pragma solidity 0.5.17;
+
+import "./AbstractBonding.sol";
+import "./ETHStaking.sol";
+
+/// @title ETH Bonding
+/// @notice Contract holding deposits from keeps' operators.
```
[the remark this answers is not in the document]
nkuba: 5c0c536cb1f9fb98eeeb744d93328687760c993c

solidity/contracts/ETHStaking.sol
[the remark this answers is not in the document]
nkuba: 5c0c536cb1f9fb98eeeb744d93328687760c993c

solidity/contracts/ETHBonding.sol
```diff
@@ -0,0 +1,70 @@
...
+contract ETHBonding is AbstractBonding {
+    ETHStaking ethStaking;
+
+    /// @notice Initializes Keep Bonding contract.
+    /// @param registryAddress Keep registry contract address.
+    /// @param ethStakingAddress ETH Staking contract address.
+    constructor(address registryAddress, address ethStakingAddress)
```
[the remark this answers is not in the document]
nkuba: 3c720641f4a7ea8c80ea0c1b06c96ff6689755a4

solidity/contracts/ETHStaking.sol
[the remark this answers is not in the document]
nkuba: It's covered by https://github.com/keep-network/keep-ecdsa/pull/483#discussion_r472879989

solidity/contracts/ETHStaking.sol
nkuba: 1e86e1a197b477af6ff7ebc09477afadc5dd8817

solidity/contracts/ETHStaking.sol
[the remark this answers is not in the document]
nkuba: Just wondering if it's enough to hardcode the value or do we want it to be editable via setter? Maybe we want to pass it in a constructor?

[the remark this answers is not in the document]
nkuba: 666238cb4afdf9f3d9e680b88da29b4b908779e3

solidity/contracts/ETHStaking.sol
```diff
@@ -0,0 +1,66 @@
...
+        );
+
+        operators[operator] = Operator(0, _from, beneficiary, authorizer);
+        ownerOperators[_from].push(operator);
+
+        // TODO: Add initialization period as per https://github.com/keep-network/keep-ecdsa/pull/483#discussion_r468628872
+
+        emit Staked(_from, operator, beneficiary, authorizer);
+    }
+}
```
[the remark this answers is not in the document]
nkuba: 8a8219de75b5e64f24e8d99f48532f58fe24a8a6

solidity/contracts/EthBonding.sol
```diff
@@ -0,0 +1,164 @@
...
+import "@keep-network/keep-core/contracts/Authorizations.sol";
+import "@keep-network/keep-core/contracts/KeepRegistry.sol";
+import "@keep-network/keep-core/contracts/StakeDelegatable.sol";
+
+import "@keep-network/sortition-pools/contracts/api/IFullyBackedBonding.sol";
+
+/// @title ETH Bonding
+/// @notice Contract holding deposits and delegations for ETH-only keeps'
+/// operators. An owner of the ETH can delegate ETH to an operator. The value
+/// of ETH the owner is willing to delegate should be deposited for the given
```
pdyraga, line 28: Instead:

> should be deposited for the given operator.

Perhaps we should say:

> is deposited as the unbonded value for the given operator.

solidity/contracts/EthBonding.sol
```diff
@@ -0,0 +1,164 @@
...
+/// operator.
+contract EthBonding is
+    AbstractBonding,
+    Authorizations,
+    StakeDelegatable,
+    IFullyBackedBonding
+{
+    event Delegated(address indexed owner, address indexed operator);
+
+    event OperatorDelegated(
```
pdyraga, line 38: In `TokenStaking` we have `StakeDelegated` and `OperatorStaked`. Ideally those wo should be just one event but given the constraint about the maximum number of indexed parameters, we had to split it into two events. The way I think about it is:

_Stake has been Delegated so the Operator is now Staked._

It does not fit here:

_ETH has been Delegated so the Operator is now... Delegated?_ 

The only two ideas that come to my mind are:
- _ETH has been Delegated so the Operator is now Configured_ 
  translating into `EthDelegated` and `OperatorConfigured`
- _ETH has been Delegated and the Operator Received the Delegation_
  translating into `EthDelegated` and `DelegationReceived`

solidity/contracts/EthBonding.sol
```diff
@@ -0,0 +1,164 @@
...
+        address indexed operator,
+        address indexed beneficiary,
+        address indexed authorizer
+    );
+
+    // TODO: Decide on the final value and if we want a setter for it.
+    uint256 public constant MINIMUM_DELEGATION_DEPOSIT = 12345;
```
pdyraga, line 45: I'd name it just `MINIMUM_DELEGATION` given that we have _a minimum stake_.

solidity/contracts/EthBonding.sol
```diff
@@ -0,0 +1,164 @@
...
+        initializationPeriod = _initializationPeriod;
+    }
+
+    /// @notice Registers delegation details. The function is used to register
+    /// addresses of operator, beneficiary and authorizer for a delegation from
+    /// the caller.
+    /// The function requires ETH to be submitted in the call as a protection
+    /// against attacks blocking operators. The value should be at least equal
+    /// to the minimum delegation deposit. Whole amount is deposited as operator's
+    /// unbonded value for the future bonding.
```
pdyraga, line 68: The fact we are preventing operator griefing this way sounds to me like a candidate for `/// @dev` detail.

solidity/contracts/fully-backed/FullyBackedBonding.sol
```diff
@@ -0,0 +1,191 @@
...
+        address indexed authorizer
+    );
+
+    // The ether value (in wei) that should be passed along with the delegation
+    // and deposited for bonding.
+    uint256 public constant MINIMUM_DELEGATION_DEPOSIT = 12345; // TODO: Decide right value
+
+    // Once a delegation to an operator is received the delegator has to wait for
+    // specific time period before being able to pull out the funds.
+    uint256 public constant DELEGATION_LOCK_PERIOD = 1 days; // TODO: Decide right value
```
pdyraga, line 50: I know it's just a dummy value now but to do not lose it: intuition tells me that we should be talking about at least a week or two.

pdyraga (commented):
More to come.

solidity/contracts/fully-backed/FullyBackedBondedECDSAKeepFactory.sol
```diff
@@ -0,0 +1,254 @@
...
+    using SafeMath for uint256;
+
+    // Sortition pool is created with a minimum bond of 20 ETH to avoid
+    // small operators joining and griefing future selections before the
+    // minimum bond is set to the right value by the application.
+    //
+    // Anyone can create a sortition pool for an application with the default
+    // minimum bond value but the application can change this value later, at
+    // any point.
+    uint256 public constant minimumBond = 20e18; // 20 ETH // TODO: Decide on value
```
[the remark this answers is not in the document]
pdyraga: (Probably not but let's keep this convo open for now)

solidity/contracts/KeepBonding.sol
```diff
@@ -31,16 +31,16 @@ contract KeepBonding is AbstractBonding {
...
-        address tokenGrantAddress
-    ) public AbstractBonding(registryAddress) {
-        tokenStaking = TokenStaking(tokenStakingAddress);
-        tokenGrant = TokenGrant(tokenGrantAddress);
+        KeepRegistry _keepRegistry,
+        TokenStaking _tokenStaking,
+        TokenGrant _tokenGrant
+    ) public AbstractBonding(_keepRegistry) {
+        tokenStaking = _tokenStaking;
+        tokenGrant = _tokenGrant;
```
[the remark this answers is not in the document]
nkuba: We also updated types from addresses to params. But yeah, we don't have to do it. I'll try to revert changes.

solidity/contracts/fully-backed/FullyBackedBondedECDSAKeep.sol
```diff
@@ -0,0 +1,71 @@
...
+import "../AbstractBondedECDSAKeep.sol";
+import "./FullyBackedBonding.sol";
+import "./FullyBackedBondedECDSAKeepFactory.sol";
+
+/// @title Fully Backed Bonded ECDSA Keep
+/// @notice ECDSA keep with additional signer bond requirement that is fully backed
+/// by ETH only.
+/// @dev This contract is used as a master contract for clone factory in
+/// BondedECDSAKeepFactory as per EIP-1167. It should never be removed after
+/// initial deployment as this will break functionality for all created clones.
```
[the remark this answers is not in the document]
nkuba: Copy-paste from `BondedECDSAKeep`. We should remove this part of comment.

pdyraga (commented):
Completed review of `FullyBackedBonding`. For the records, questions I tried to answer while reviewing:

Q: Can I top-up my delegation?
A: Yes, though it could be hard to KEEP Token Dashboard to display the total amount of ETH ever deposited. It's also a challenge for `KeepBonding`, that's why instead of displaying all ETH ever deposited, the dashboard displays the amount that is locked and the amount that is available. I think we can address this problem for ETH-only option by adding a `topUp` function in the contract. Left a comment about that.

Q: Can I decide to do not candidate to new keeps so that I can withdraw all funds later?
A: Yes. However, instead of "undelegating" ETH, you have to deauthorize sortition pool for the given operator. I was thinking about adding a concept of undelegating to `FullyBackedBonding` for a moment, but I think it does not fit here well as we do not work with a stake. We will have to think about how to best present this process in KEEP Token Dashboard but I prefer to solve it with a good UX instead of complicating contracts.

Q: Can I cancel my delegation?
A: No. You need to deauthorize sortition pool and then wait for unlocking period.

solidity/contracts/fully-backed/FullyBackedBonding.sol
```diff
@@ -0,0 +1,191 @@
...
+            beneficiary,
+            authorizer
+        );
+
+        deposit(operator);
+
+        emit Delegated(owner, operator);
+        emit OperatorDelegated(operator, beneficiary, authorizer);
+    }
+
```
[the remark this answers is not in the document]
nkuba: We've got `deposit` function on bonding contract that emits `UnbondedValueDeposited`. I'm not sure about duplicating it. What if someone top-ups directly with `deposit` function? This is how we handle it currently for KEEP backed bonding.
To calculate the values we will need to analyze events, we started working on that in https://github.com/keep-network/local-setup/pull/48.

solidity/contracts/fully-backed/FullyBackedECDSAKeepFactory.sol
```diff
@@ -0,0 +1,254 @@
...
+    using SafeMath for uint256;
+
+    // Sortition pool is created with a minimum bond of 20 ETH to avoid
+    // small operators joining and griefing future selections before the
+    // minimum bond is set to the right value by the application.
+    //
+    // Anyone can create a sortition pool for an application with the default
+    // minimum bond value but the application can change this value later, at
+    // any point.
+    uint256 public constant minimumBond = 20e18; // 20 ETH // TODO: Decide on value
```
pdyraga, line 60: Let's use the same value as in `BondedECDSAKeepFactory`. What's proposed here is fine.

solidity/contracts/fully-backed/FullyBackedECDSAKeepFactory.sol
```diff
@@ -0,0 +1,254 @@
...
+
+    using SafeMath for uint256;
+
+    // Sortition pool is created with a minimum bond of 20 ETH to avoid
+    // small operators joining and griefing future selections before the
+    // minimum bond is set to the right value by the application.
+    //
+    // Anyone can create a sortition pool for an application with the default
+    // minimum bond value but the application can change this value later, at
+    // any point.
```
pdyraga, line 59: Thoughts on explaining how this value relates to the minimum delegation from `FullyBackedBonding` for future-us and other developers looking at this code?

solidity/contracts/fully-backed/FullyBackedBonding.sol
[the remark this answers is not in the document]
pdyraga: I am also not sold on this idea but I also can't imagine analyzing all the events when rendering a page in token dashboard, it would take too much time.

solidity/migrations/2_deploy_contracts.js
```diff
@@ -7,12 +8,24 @@ const BondedECDSAKeepVendorImplV1 = artifacts.require(
   "BondedECDSAKeepVendorImplV1"
 )
 
+const FullyBackedBonding = artifacts.require("FullyBackedBonding")
+const FullyBackedECDSAKeep = artifacts.require("FullyBackedECDSAKeep")
+const FullyBackedECDSAKeepFactory = artifacts.require(
+  "FullyBackedECDSAKeepFactory"
+)
+
+const EthBonding = artifacts.require("EthBonding")
```
pdyraga, line 17: Where is this guy defined?

solidity/migrations/2_deploy_contracts.js
```diff
@@ -68,4 +83,20 @@ module.exports = async function (deployer) {
     BondedECDSAKeepVendorImplV1.address,
     implInitializeCallData
   )
+
+  // ETH-only Staking
```
pdyraga, line 87: `ETH bonding only`?

solidity/migrations/2_deploy_contracts.js
```diff
@@ -37,6 +51,7 @@ module.exports = async function (deployer) {
     RegistryAddress = (await deployer.deploy(KeepRegistry)).address
   }
 
+  // KEEP and ETH backed bonding
```
pdyraga, line 54: `KEEP staking and ETH bonding`?

solidity/migrations/2_deploy_contracts.js
```diff
@@ -68,4 +83,20 @@ module.exports = async function (deployer) {
...
+  await deployer.deploy(
+    FullyBackedECDSAKeepFactory,
+    FullyBackedECDSAKeep.address,
+    FullyBackedSortitionPoolFactory.address,
+    FullyBackedBonding.address,
+    RandomBeaconAddress
+  )
+
+  // ETH-only backed bonding
+  await deployer.deploy(EthBonding, RegistryAddress)
```
pdyraga, line 101: I don't think we need it, `FullyBackedBonding` is already defined, I think it's a result of some wrong merge conflict resolution.

solidity/migrations/3_initialize.js
```diff
@@ -3,12 +3,16 @@ const BondedECDSAKeepVendorImplV1 = artifacts.require(
...
+const FullyBackedECDSAKeepFactory = artifacts.require(
+  "FullyBackedECDSAKeepFactory"
+)
 const KeepRegistry = artifacts.require("KeepRegistry")
 
 const {RegistryAddress} = require("./external-contracts")
 
 module.exports = async function (deployer) {
   await BondedECDSAKeepFactory.deployed()
+  await BondedECDSAKeepFactory.deployed()
```
pdyraga, line 15: Why are we doing the same call twice?

solidity/contracts/fully-backed/FullyBackedBonding.sol
```diff
@@ -0,0 +1,191 @@
...
+import "@keep-network/keep-core/contracts/StakeDelegatable.sol";
+import "@keep-network/keep-core/contracts/KeepRegistry.sol";
+
+import "@keep-network/sortition-pools/contracts/api/IFullyBackedBonding.sol";
+
+/// @title Fully Backed Bonding
+/// @notice Contract holding deposits and delegations for ETH-only keeps'
+/// operators. An owner of the ETH can delegate ETH to an operator. The value
+/// of ETH the owner is willing to delegate should be deposited for the given
+/// operator.
```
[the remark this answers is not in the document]
nkuba: 🐮 🤠🥛

solidity/test/FullyBackedBondingTest.js
```diff
@@ -0,0 +1,451 @@
...
+        "Only operator or the owner is allowed to withdraw bond"
+      )
+    })
+
+    it("transfers unbonded value to beneficiary", async () => {
+      const expectedUnbonded = initialDeposit.sub(value)
+
+      const expectedBeneficiaryBalance = web3.utils
+        .toBN(await web3.eth.getBalance(beneficiary))
+        .add(value)
```
pdyraga, line 309: Can we move those declarations to be next to assertions? It took me a couple of minutes to understand what `.sub(value)` is doing in `expectedUnbonded` declaration.

I'd like to see `await bonding.withdraw(value, operator, {from: operator})` first and then the expected values.

solidity/migrations/2_deploy_contracts.js
nkuba: 💣

solidity/contracts/test/FullyBackedECDSAKeepFactoryStub.sol
```diff
@@ -0,0 +1,36 @@
...
+
+    function getGroupSelectionSeed() public view returns (uint256) {
+        return groupSelectionSeed;
+    }
+
+    function addKeep(address keep) public {
+        keeps.push(keep);
+        /* solium-disable-next-line security/no-block-members*/
+        keepOpenedTimestamp[keep] = block.timestamp;
+    }
```
pdyraga, line 35: Is this function used anywhere?

solidity/test/FullyBackedBondingTest.js
```diff
@@ -0,0 +1,451 @@
+const {accounts, contract, web3} = require("@openzeppelin/test-environment")
```
pdyraga, line 1: I am getting test failures locally from time to time:
```
  1) FullyBackedBonding
       withdraw
         cannot be called before delegation lock period passes:
     AssertionError: Expected an exception but none was received
      at expectException (node_modules/@openzeppelin/test-helpers/src/expectRevert.js:25:10)
      at processTicksAndRejections (internal/process/task_queues.js:86:5)

  2) FullyBackedBonding
       isInitialized
         returns false when authorized but initialization period not passed yet:

      AssertionError: expected true to be false
      + expected - actual

      -true
      +false
      
      at Context.it (test/FullyBackedBondingTest.js:399:14)
      at processTicksAndRejections (internal/process/task_queues.js:86:5)
```

I think we need to set a little safer time margin for those tests.

solidity/test/FullyBackedBondingTest.js
nkuba: We usually first define expectations, next execute test action and then run verification. I can move pieces here as you wish if this helps with understanding the test case.

nkuba: We cannot move `expectedBeneficiaryBalance` after `withdraw` as we have to call `web3.eth.getBalance` to get initial account balance before calling the action changing the state.

solidity/test/FullyBackedBondingTest.js
nkuba: Should be fine now.

solidity/test/FullyBackedECDSAKeepFactoryTest.js
```diff
@@ -0,0 +1,1633 @@
...
+    before(async () => {
+      await initializeNewFactory()
+      await initializeMemberCandidates()
+
+      const pool = await FullyBackedSortitionPool.at(signerPool)
+      minimumBondableValue = await pool.getMinimumBondableValue()
+    })
+
+    // TODO: Add a case to check if operator is initialized
+    // github.com/keep-network/sortition-pools/blob/413abd66536032ee757f190a29fda2705c7b70cf/contracts/FullyBackedSortitionPool.sol#L155
```
pdyraga, line 355: Should we?

solidity/test/FullyBackedECDSAKeepFactoryTest.js
```diff
@@ -0,0 +1,1633 @@
...
+      // (20 * 10^18 * 10) / 10^18                 =  200
+      // ((20 * 10^18 * 10) + (10^18 - 1)) / 10^18 =~ 200.99
+      // Ethereum uint256 division performs implicit floor:
+      // The weight dit not change: 200 == floor(200.99)
+      assert.isTrue(
+        await keepFactory.isOperatorUpToDate(members[0], application)
+      )
+    })
+
+    it("returns false if the operator stake dropped below minimum", async () => {
```
pdyraga, line 426: s/stake/bond

solidity/test/FullyBackedECDSAKeepFactoryTest.js
```diff
@@ -0,0 +1,1633 @@
...
+      assert.isTrue(
+        await keepFactory.isOperatorUpToDate(members[0], application),
+        "unexpected status of the operator after bonding value change"
+      )
+
+      await expectRevert(
+        keepFactory.updateOperatorStatus(members[0], application),
+        "Operator already up to date"
+      )
+    })
```
pdyraga, line 498: I think we should have another test that would increase this value by more ETH, then assert operator is no longer up to date, update the operator, and assert the operator is up to date again.

The name of this test suggests we are testing the update operation completes successfully but what the test implementation really does is checking that the update is not needed and if executed, it fails.

solidity/test/FullyBackedECDSAKeepFactoryTest.js
```diff
@@ -0,0 +1,1633 @@
...
+        await keepFactory.isOperatorUpToDate(members[0], application),
+        "unexpected status of the operator after bonding value change"
+      )
+
+      await keepFactory.updateOperatorStatus(members[0], application)
+
+      await expectRevert(
+        keepFactory.isOperatorUpToDate(members[0], application),
+        "Operator not registered for the application"
+      )
```
pdyraga, line 484: We should have another test updating operator status if the value dropped but not removing it from the pool (unbonded value still above the minimum).

solidity/contracts/fully-backed/FullyBackedBonding.sol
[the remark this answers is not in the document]
nkuba: I can imagine that we will have to analyze  in the dashboard all the events anyway, to cover cases where top-up function was omitted and eth has been depositet directly.

solidity/test/FullyBackedECDSAKeepFactoryTest.js
nkuba: It was already covered in registration test: https://github.com/keep-network/keep-ecdsa/blob/d1edab304c4ace73f720f6c4e6d429a14ba6a19b/solidity/test/FullyBackedECDSAKeepFactoryTest.js#L221

It won't be possible to happen on `isOperatorUpToDate` as operator would be initialized already once they join the pool.

I'll remove this comment.

solidity/contracts/fully-backed/FullyBackedBonding.sol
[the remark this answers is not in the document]
pdyraga: Yesterday, I tried to run the script from `local-setup` analyzing bonds per event locally against Infura. One hour was not enugh to produce results for ~100 mainnet operators.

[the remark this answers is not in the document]
nkuba: I'm not sure how much optimized is the script you mentioned. Tracking the value in contract instead of calculating it from events would be the fastest solution.

I believe we just have to agree to disagree. I'll add a top-up function you wanted.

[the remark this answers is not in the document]
pdyraga: > Tracking the value in contract instead of calculating it from events would be the fastest solution.

We'd need a separate function to return bonds to the bonding contract. So `deposit` would be used only by the user and `returnBonds` (or any other function name) would be used to return partial bonds by application. Too late for this one and it still means we need two functions and there is no guarantee user calls the right one.

solidity/test/FullyBackedECDSAKeepFactoryTest.js
nkuba: It's already covered pretty well in sortition pool tests https://github.com/keep-network/sortition-pools/blob/678a7b7d10274329fe77b36fe59a2f34e7fc0f65/test/fullyBackedSortitionPoolTest.js#L324-L359

Do we want to duplicate these tests in the factory tests?

solidity/test/FullyBackedECDSAKeepFactoryTest.js
nkuba: It's covered by sortition pool test https://github.com/keep-network/sortition-pools/blob/678a7b7d10274329fe77b36fe59a2f34e7fc0f65/test/fullyBackedSortitionPoolTest.js#L293-L323

solidity/test/FullyBackedECDSAKeepFactoryTest.js
pdyraga: I think it's fair to have such a test here, especially that we have one for `BondedECDSAKeepFactory`, please see `updates operator if stake has changed above minimum`.

We don't need to assert weight and all sortition-pool specific properties, but `isOperatorUpToDate`/`updateOperatorStatus` are factory's functions and we should cover this behavior with a unit test.

solidity/test/FullyBackedECDSAKeepFactoryTest.js
pdyraga: Yeah, but we are testing factory functions here  - we don't need to assert properties of sortition pool but it is worth to cover all `isOperatorUpToDate` and `updateOperatorStatus` paths. The scenario I am writing about is covered in `removes operator if stake has changed below minimum` test of `BondedECDSAKeepFactoryTest`.

solidity/migrations/3_initialize.js
```diff
@@ -32,7 +35,12 @@ module.exports = async function (deployer) {
...
   await registry.approveOperatorContract(BondedECDSAKeepFactory.address)
   console.log(
-    `approved operator contract [${BondedECDSAKeepFactory.address}] in registry`
+    `approved BondedECDSAKeepFactory operator contract [${BondedECDSAKeepFactory.address}] in registry`
+  )
+
+  await registry.approveOperatorContract(FullyBackedECDSAKeepFactory.address)
+  console.log(
+    `approved FullyBackedECDSAKeepFactory operator contract [${FullyBackedECDSAKeepFactory.address}] in registry`
   )
```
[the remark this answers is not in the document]
nkuba: We can remove this call.

solidity/test/FullyBackedECDSAKeepFactoryTest.js
nkuba: It's sortition-pool based logic that we're testing. But, OK I'll duplicate the tests.

solidity/test/FullyBackedECDSAKeepFactoryTest.js
pdyraga: To summarize:
- `removes operator if bonding value has changed below minimum` - bond dropped below the minimum, operator is removed from the pool as a result of `updateOperatorStatus` and `isOperatorUpToDate` reverts.
- `updates operator if bonding value has changed above minimum` - bond value has changed but the change was insignificant (less than pool's divisor) and pool assumes no change happened so `updateOperatorStatus` reverts.

We do not have the scenario where:
- bonding value dropped, but it is still above the minimum. Operator is updated as a result of `updateOperatorStatus` call and it stays in the pool, so `isOperatorUpToDate` works.
- bonding value changed, it's still above the minimum but the change was significant and update is required. This is covered by `updates operator if stake has changed above minimum` scenario in `BondedECDSAKeepFactoryTest`.

nkuba: Just to clarify the summary: scenarios you mentioned are covered well in sortition-pools, which is where they belong; I'll duplicate them in factory tests as you wish.

solidity/test/FullyBackedECDSAKeepFactoryTest.js
pdyraga: If we are worried about duplicating tests of sortition pool - and that's fine we are worried about it - we should remove the current test as the fact we add 1 wei to the bond and this update is ignored by the pool is the pool's logic.

We are missing the very basic scenario here when bond has changed and update needs to be performed.

nkuba: The tests we have currently here were added just to test integration with sortition pool. Any test would be fine to cover that. I'll enhance them to cover more cases.

pdyraga: A good practice is to test all paths for the component under the test, without going to deep into the test of its dependencies. In this context, we are testing the factory and we don't care what's under `updateOperatorStatus`/`isOperatorUpToDate`. Maybe those functions hit the factory directly and maybe they implement some local caching because hitting the factory every time is expensive. Also, the factory defines the concept of the minimum bond. So I do agree that testing things like no update because the change is insigificant (below pool weight divisor) is not something we should do in tests for the factory, all paths regarding update and minimum stake should be covered.

solidity/migrations/3_initialize.js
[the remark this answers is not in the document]
pdyraga: Do you know what was the point of having it here?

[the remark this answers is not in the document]
nkuba: Probably loading deployed contract so we can use `address` property, right now it works without this call.

solidity/test/FullyBackedECDSAKeepFactoryTest.js
```diff
@@ -0,0 +1,1685 @@
...
+
+    before(async () => {
+      await initializeNewFactory()
+      await initializeMemberCandidates()
+      await registerMemberCandidates()
+
+      const pool = await FullyBackedSortitionPool.at(signerPool)
+      minimumBondableValue = await pool.getMinimumBondableValue()
+
+      await setUnbondedValue(members[0], minimumBondableValue.muln(3))
```
pdyraga, line 457: This call happens after `joinPool` executed inside `registerMemberCandidates` so all tests are starting with the operator being not up-to-date. That's why `revers if operator is up to date` is failing. We should probably move this call inside `registerMemberCandidates` or call an update here.

solidity/test/FullyBackedECDSAKeepFactoryTest.js
```diff
@@ -0,0 +1,1685 @@
...
+    })
+
+    it("revers if operator is up to date", async () => {
```
pdyraga, line 460: s/revers/reverts

solidity/test/FullyBackedECDSAKeepFactoryTest.js
```diff
@@ -0,0 +1,1685 @@
...
+      )
+    })
+
+    it("updates operator if bonding value decreased insignificantly above minimum", async () => {
+      bonding.withdraw(new BN(1), members[0], {from: members[0]})
+
+      assert.isFalse(
+        await keepFactory.isOperatorUpToDate(members[0], application),
+        "unexpected status of the operator after stake change"
+      )
```
pdyraga, line 521: Should it be `isFalse` or `isTrue`? If it dropped insignificantly, the operator should still be up to date.

To be clear: I am fine with not having all `*insignificantly*` scenarios in `FullyBackedECDSAKeepFactoryTest.js` as I do agree it kinda duplicates sortition pool tests. What I want to have here all other paths added in this commit 👌 

We can leave it, we can drop it, I am fine with both.

solidity/contracts/fully-backed/FullyBackedBonding.sol
```diff
@@ -104,6 +106,18 @@ contract FullyBackedBonding is
         emit OperatorDelegated(operator, beneficiary, authorizer, msg.value);
     }
 
+    /// @notice Top-ups operator's unbonded value.
+    /// @dev This function should be used to add new unbonded value to the system
+    /// for an operator. The `deposit` function defined in parent abstract contract
+    /// should be called only by applications returning value that has been already
+    /// initially deposited and seized later.
```
pdyraga, line 113: Worth expanding`@dev` and explaining why this function should be used instead of `deposit` (event and tracking ETH).

solidity/test/FullyBackedBondingTest.js
```diff
@@ -204,6 +204,54 @@ describe("FullyBackedBonding", function () {
     })
   })
 
+  describe("topUp", async () => {
+    const value = new BN(123)
+
+    let initialDeposit
+
+    before(async () => {
```
pdyraga, line 212: Should this one be `before` or `beforeEach`? I think `before` does not work the way we expect it to work in this context.

solidity/test/FullyBackedECDSAKeepFactoryTest.js
nkuba: > Should it be `isFalse` or `isTrue`? If it dropped insignificantly, the operator should still be up to date.

https://github.com/keep-network/sortition-pools/blob/678a7b7d10274329fe77b36fe59a2f34e7fc0f65/contracts/FullyBackedSortitionPool.sol#L193-L195

It depends on how much it dropped. Due to the `floor` calculation for division an insignificant change has different implication on results between addition and subtraction. In this case an insignificant drop in unbonded value causes drop of the eligible weight so the change is noticable.

solidity/test/FullyBackedBondingTest.js
nkuba: We defined `beforeEach` with snapshots at the top level, these all doesn't work well together. Let me figure it out.

pdyraga (approved):

PR #486, opened 2020-06-23
Added Log rotation to docker run command

afmsavage:
Added in log rotation to the docker run command so that only 300mb of logs are stored at a time.  Logs over time, especially when run with `debug` will create GB of log files.


sthompson22:
> Ya sure, let's party 🎉

As soon as builds are green 😬

afmsavage:
w00t

PR #487, opened 2020-06-24
Save last seen peers and use them as bootstrap upon restart

experiencedft:
Related to [this ``keep-core`` issue](https://github.com/keep-network/keep-core/issues/1856). 

See the [corresponding PR](https://github.com/keep-network/keep-core/pull/1877) on ``keep-core`` for more explanations.

This simply extends the feature to the ECDSA client.

### DIFF
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"fmt"
 	"time"
+	"strings"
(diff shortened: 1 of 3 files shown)


Shadowfiend:
We're deep in contract-land at the moment, so it might be a few days before someone circles back here; rest assured, we'll have a look!

Issue #491, opened 2020-07-06
Expect the delegation to exist before depositing ETH in the bonding contract


pdyraga:
In https://github.com/keep-network/keep-ecdsa/pull/478 we addressed the issue when the beneficiary is not yet defined when staker wants to withdraw tokens from the escrow.

Instead of reverting the withdrawal transaction, we should consider blocking unbonded value deposits when the beneficiary is not yet defined (delegation does not exist).

PR #492, opened 2020-07-07
Checking beneficiary address before depositing funds

dimpar:
Closes https://github.com/keep-network/keep-ecdsa/issues/491

Blocking unbonded value deposits when the beneficiary is not yet defined (delegation does not exist).
Removing check from `withdraw` function that previously blocked withdrawal if the beneficiary was not defined.

### DIFF
--- a/solidity/contracts/KeepBonding.sol
+++ b/solidity/contracts/KeepBonding.sol
@@ -21,7 +21,6 @@ import "@keep-network/keep-core/contracts/libraries/RolesLookup.sol";
 
 import "openzeppelin-solidity/contracts/math/SafeMath.sol";
 
-
 /// @title Keep Bonding
 /// @notice Contract holding deposits from keeps' operators.
 contract KeepBonding {
@@ -231,11 +230,11 @@ contract KeepBonding {
     /// @param holder Address of the holder of the bond.
     /// @param referenceID Reference ID of the bond.
     /// @return Amount of wei in the selected bond.
-    function bondAmount(address operator, address holder, uint256 referenceID)
-        public
-        view
-        returns (uint256)
-    {
+    function bondAmount(
+        address operator,
+        address holder,
+        uint256 referenceID
+    ) public view returns (uint256) {
         bytes32 bondID = keccak256(
             abi.encodePacked(operator, holder, referenceID)
         );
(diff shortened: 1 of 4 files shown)


solidity/contracts/KeepBonding.sol
```diff
@@ -314,6 +327,7 @@ contract KeepBonding {
 
         lockedBonds[bondID] = lockedBonds[bondID].sub(amount);
 
+        /* solium-disable-next-line */
```
pdyraga, line 330: Isn't it disabled in the global `.soliumrc.json`?

solidity/test/BondedECDSAKeepFactoryTest.js
```diff
@@ -43,6 +43,7 @@ contract("BondedECDSAKeepFactory", async (accounts) => {
   const notMember = accounts[5]
 
   const keepOwner = accounts[6]
+  const beneficiary = notMember
```
pdyraga, line 46: Why don't we use `accounts[7]`?

solidity/test/BondedECDSAKeepTest.js
```diff
@@ -1865,6 +1865,11 @@ contract("BondedECDSAKeep", (accounts) => {
   }
 
   async function depositForBonding(member1Value, member2Value, member3Value) {
+    const beneficiary = accounts[4]
```
pdyraga, line 1868: Can we define `beneficiary` on the top of the file, where we define the `bondCreator`, `members`, `authorizers` and all the rest of the roles?

solidity/contracts/KeepBonding.sol
dimpar: Yes, you're right. My VCS underlined it red and that mislead me.

solidity/test/BondedECDSAKeepFactoryTest.js
dimpar: No particular reason, just picked an account that not a member. Changed to accounts[7]

solidity/test/BondedECDSAKeepTest.js
dimpar: Sure, done.

PR #493, opened 2020-07-08
Bumping up keep-core dependency to 1.3.0-pre

dimpar:
Bumping up keep-core dependency to 1.3.0-pre and hdwallet-provider to 1.0.37
Updating failed integration tests.

### DIFF
--- a/solidity/package.json
+++ b/solidity/package.json
@@ -32,7 +32,7 @@
   },
   "homepage": "https://github.com/keep-network/keep-ecdsa",
   "dependencies": {
-    "@keep-network/keep-core": ">1.3.0-pre <1.3.0",
+    "@keep-network/keep-core": ">1.2.4-rc <1.2.4",
     "@keep-network/sortition-pools": "1.0.0",
     "@openzeppelin/upgrades": "^2.7.2",
     "openzeppelin-solidity": "2.3.0",
--- a/solidity/package.json
+++ b/solidity/package.json
@@ -32,7 +32,7 @@
   },
   "homepage": "https://github.com/keep-network/keep-ecdsa",
   "dependencies": {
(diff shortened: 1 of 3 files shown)


solidity/package.json
```diff
@@ -32,7 +32,7 @@
   },
   "homepage": "https://github.com/keep-network/keep-ecdsa",
   "dependencies": {
-    "@keep-network/keep-core": ">1.2.4-rc <1.2.4",
+    "@keep-network/keep-core": ">1.3.0-pre <1.3.0",
```
pdyraga, line 35: `>1.3.0-pre <1.3.0-rc` perhaps?

solidity/test/BondedECDSAKeepIntegrationTest.js
```diff
@@ -254,6 +256,9 @@ contract("BondedECDSAKeepFactory", async (accounts) => {
       keepTokenGrant.address
     )
 
+    await TokenStaking.link("GrantStaking", (await GrantStaking.new()).address)
+    await TokenStaking.link("Locks", (await Locks.new()).address)
```
pdyraga, line 260: Let's keep them next to `MinimumStakeSchedule` linkage.

solidity/package.json
dimpar: we don't want to include release candidates?

pdyraga: `-pre` should refer to `-pre`s. We should also update the version of this package on the top of the file.

dimpar: done

solidity/test/BondedECDSAKeepIntegrationTest.js
dimpar: done

Issue #490, opened 2020-07-06
Remove bonded ECDSA keep signature submission timeout


pdyraga:
Bonded ECDSA keep requires the signature to be provided within a certain timeout. When the signature is not provided on time, the new signature can not be requested and signature, and the pending signature can not be submitted. This blocks the keep forever and means there is no opportunity for the application to retry.

It seems that this restriction is quite severe and letting the application decide whether to give another chance or seize bonds could be a better solution. In case of tBTC, it could mean that the signers can retry until some maintainer reports the timeout.

We should analyze the flow and consider removing the signature submission timeout from bonded ECDSA keep.

PR #494, opened 2020-07-09
Remove BondedECDSAKeep signing timeout

nkuba:
This PR removes signing timeouts handling for BondedECDSAKeep. We expect the application to track the timeout and close the keep when it's exceeded. Until then the keep should expect the signature to be delivered by a client.

Please expect two more PRs:
1. We will introduce a similar change for key generation timeout (https://github.com/keep-network/keep-ecdsa/pull/496).
2. We will make client timeouts configurable in a config file.

Closes: https://github.com/keep-network/keep-ecdsa/issues/490


pdyraga:
@nkuba Can you please update this branch with `master`?

PR #497, opened 2020-07-10
Configurable look-back period and keygen and signing timeouts

nkuba:
In this PR we let operator configure values for a look-back period for awaiting key generation keeps and key generation and signing timeouts.

Values can be configured in config file as on following example:
```toml
[Client]
  AwaitingKeyGenerationLookback = "24h"
  KeyGenerationTimeout = "3h"
  SigningTimeout = "2h"
```

Providing the values is optional. If the values are not defined the default ones will be used.

Refs: https://github.com/keep-network/keep-ecdsa/issues/490

### DIFF
--- a/configs/config.toml.SAMPLE
+++ b/configs/config.toml.SAMPLE
@@ -44,7 +44,7 @@
 #   # AnnouncedAddresses = ["/dns4/example.com/tcp/3919", "/ip4/80.70.60.50/tcp/3919"]
 
 [Client]
-# Look-back period to check if existing keeps are awaiting signer generation.
+# Look-back period to check if existing keeps are awaiting signing key generation.
 # The value should be provided based on the sanctioned application requirements.
 #  AwaitingKeyGenerationLookback = "24h"	# optional
 
--- a/configs/config.toml.SAMPLE
+++ b/configs/config.toml.SAMPLE
@@ -43,6 +43,17 @@
 #   # Uncomment to override the node's default addresses announced in the network
 #   # AnnouncedAddresses = ["/dns4/example.com/tcp/3919", "/ip4/80.70.60.50/tcp/3919"]
 
+[Client]
+# Look-back period to check if existing keeps are awaiting signer generation.
+# The values should be provided based on the sanctioned application requirements.
+#  AwaitingKeyGenerationLookback = "24h"	# optional
+
+# Timeouts for processes execution. Within these timeouts the process will keep
+# retrying to generate a signer or calculate a signature. The values should be
+# provided based on the sanctioned application requirements.
+#  KeyGenerationTimeout = "3h" 				# optional
+#  SigningTimeout = "2h"					# optional
+
 [TSS]
 # Timeout for TSS protocol pre-parameters generation. The value
 # should be provided based on resources available on the machine running the client.
--- a/configs/config.toml.SAMPLE
+++ b/configs/config.toml.SAMPLE
@@ -44,8 +44,11 @@
 #   # AnnouncedAddresses = ["/dns4/example.com/tcp/3919", "/ip4/80.70.60.50/tcp/3919"]
 
 [Client]
-# Look-back period to check if existing keeps are awaiting signing key generation.
-# The value should be provided based on the sanctioned application requirements.
+# Look-back period to check if existing, active keeps are awaiting signer generation.
+# When the client starts, it goes through all keeps registered on-chain to check
+# whether it's a member of one of them and to generate the signing key if needed.
+# The client does not check keeps older than `AwaitingKeyGenerationLookback` to
+# minimize the number of calls to the chain.
 #  AwaitingKeyGenerationLookback = "24h"	# optional
 
 # Timeouts for processes execution. Within these timeouts the process will keep
--- a/configs/config.toml.SAMPLE
+++ b/configs/config.toml.SAMPLE
@@ -45,7 +45,7 @@
 
 [Client]
 # Look-back period to check if existing keeps are awaiting signer generation.
-# The values should be provided based on the sanctioned application requirements.
+# The value should be provided based on the sanctioned application requirements.
 #  AwaitingKeyGenerationLookback = "24h"	# optional
 
 # Timeouts for processes execution. Within these timeouts the process will keep
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -24,10 +24,10 @@ import (
 var logger = log.Logger("keep-ecdsa")
 
 const (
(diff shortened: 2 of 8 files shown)


configs/config.toml.SAMPLE
```diff
@@ -43,6 +43,17 @@
 #   # Uncomment to override the node's default addresses announced in the network
 #   # AnnouncedAddresses = ["/dns4/example.com/tcp/3919", "/ip4/80.70.60.50/tcp/3919"]
 
+[Client]
+# Look-back period to check if existing keeps are awaiting signer generation.
```
pdyraga, line 47: `are awaiting signing key generation` perhaps?

configs/config.toml.SAMPLE
```diff
@@ -43,6 +43,17 @@
...
+# The value should be provided based on the sanctioned application requirements.
+#  AwaitingKeyGenerationLookback = "24h"	# optional
+
+# Timeouts for processes execution. Within these timeouts the process will keep
+# retrying to generate a signer or calculate a signature. The values should be
+# provided based on the sanctioned application requirements.
+#  KeyGenerationTimeout = "3h" 				# optional
```
pdyraga, line 54: What's the reasoning behind separating `AwaitingKeyGenerationLookback` and `KeyGenerationTimeout`? There is no point in looking back further than `KeyGenerationTimeout`.

pkg/client/config.go
```diff
@@ -0,0 +1,15 @@
...
+
+// Config contains configuration for tss protocol execution.
+type Config struct {
+	// Period to check keeps for awaiting key generation on client start.
+	AwaitingKeyGenerationLookback configtime.Duration
+
+	// Timeout for key generation and signature calculation.
+	KeyGenerationTimeout configtime.Duration
+	SigningTimeout       configtime.Duration
+}
```
pdyraga, line 15: What do you think about defining three `Get*` functions in `Config` and handling there the default/non-default values? We can also define the defaults in this file.

configs/config.toml.SAMPLE
nkuba: It's used in two different ways. When starting a client we may want to look back for keeps from e.g. 2 days, but having such a long timeout for a single key generation process doesn't make much sense. It's strictly related to the application but having it separated makes the client more flexible and configurable.

configs/config.toml.SAMPLE
pdyraga: My proposition:
```
# Look-back period to check if existing, active keeps are awaiting signer generation.
# When the client starts, it goes through all keeps registered on-chain to check
# whether it's a member of one of them and to generate the signing key if needed.
# The client does not check keeps older than AwaitingKeyGenerationLookback to
# minimize the number of calls to the chain.
```

pkg/client/client.go
```diff
@@ -24,10 +24,10 @@ import (
 var logger = log.Logger("keep-ecdsa")
 
 const (
-	awaitingKeyGenerationLookback = 24 * time.Hour
-	keyGenerationTimeout          = 120 * time.Minute
-	signingTimeout                = 90 * time.Minute
-	blockConfirmations            = 12
+	defaultAwaitingKeyGenerationLookback = 24 * time.Hour
```
pdyraga: My proposition:
```
// The deault look-back period to check if existing, active keeps are awaiting signer generation.
// When the client starts, it goes through all keeps registered on-chain to check
// whether it's a member of one of them and to generate the signing key if needed.
// The client does not check keeps older than the awaiting key generation lookback value
// allows to minimize the number of calls to the chain calls. 
```

pkg/client/client.go
```diff
@@ -24,10 +24,10 @@ import (
...
+	defaultKeyGenerationTimeout          = 3 * time.Hour
```
pdyraga:
```
// The default value of a timeout for a keep key generation timeout
```
?

pkg/client/client.go
```diff
@@ -24,10 +24,10 @@ import (
...
+	defaultSigningTimeout                = 2 * time.Hour
```
pdyraga:
```
// The default value of a timeout for a signature generation by keep.
```

pkg/client/client.go
```diff
@@ -24,10 +24,10 @@ import (
...
+	blockConfirmations                   = 12
```
pdyraga:
```
// The default number of block confirmations the client waits for until
// it starts the requested signing process. This value prevents from
// reporting unauthorized signings by adversaries in case of a chain
// fork.
```

pkg/client/config.go
```diff
@@ -0,0 +1,54 @@
...
+	timeout := c.KeyGenerationTimeout.ToDuration()
+	if timeout == 0 {
+		timeout = defaultKeyGenerationTimeout
+	}
+
+	return timeout
+}
+
+// GetSigningTimeout return key signing timeout as `time.Duration`.
+func (c *Config) GetSigningTimeout() time.Duration {
```
pdyraga, line 47: Can we document `Get*` function in the way I proposed in the previous comments?

pkg/client/client.go
```diff
@@ -23,12 +23,10 @@ import (
 
 var logger = log.Logger("keep-ecdsa")
 
-const (
-	awaitingKeyGenerationLookback = 24 * time.Hour
-	keyGenerationTimeout          = 120 * time.Minute
-	signingTimeout                = 90 * time.Minute
-	blockConfirmations            = 12
-)
+// The default number of block confirmations the client waits until it starts the
```
pdyraga, line 26: Actually... it shouldn't be `The default` but just `The number of` - we don't overwrite this value anywhere in the config file, it's the same for all the clients.

pkg/client/config.go
```diff
@@ -0,0 +1,67 @@
...
+	// The default value of a timeout for a keep key generation.
+	defaultKeyGenerationTimeout = 3 * time.Hour
+
+	// The default value of a timeout for a signature calculation.
+	defaultSigningTimeout = 2 * time.Hour
+)
+
+// Config contains configuration for tss protocol execution.
+type Config struct {
+	// Period to check keeps for awaiting key generation on client start.
```
pdyraga, line 27: I'd improve this one a little. Perhaps:
```
// Defines the look-back period to check if existing, active keeps are awaiting
// signer generation on the client start. The client does not check keeps older 
// than the look-back value.
```

pdyraga:
Opening new keeps works, signing with existing keep works, default values are applied if no other are specified in the configuration, and values from the configuration take priority if specified. LGTM.

PR #499, opened 2020-07-10
Cache no active keep members checks

### LINKED ISSUE #498: Cache negative active keep check results
In https://github.com/keep-network/keep-core/issues/1895 we decided to cache negative results of has minimum stake checks. We should do the same for `stakeOrActiveKeepPolicy` in `keep-ecdsa` except that negative checks.

lukasz-zimnoch:
Closes #498

Introduced cache for no active keep members check results in the same way as for active keep members.

### DIFF
--- a/pkg/firewall/firewall_test.go
+++ b/pkg/firewall/firewall_test.go
@@ -412,8 +412,9 @@ func TestSweepsActiveKeepMembersCache(t *testing.T) {
 // Has authorization.
 // Is not a member of an active keep.
 // Should NOT allow to connect.
-// After some time, becomes member of an active keep.
-// Then, it should allow to connect.
+// After some time, has a minimum stake again and becomes a member of an active keep.
+// Shortly after that, the minimum stake drops below the required minimum but
+// the membership in an active keep remains.
 func TestSweepsNoActiveKeepMembersCache(t *testing.T) {
 	chain := local.Connect()
 	coreFirewall := newMockCoreFirewall()
--- a/pkg/firewall/firewall_test.go
+++ b/pkg/firewall/firewall_test.go
@@ -20,9 +20,10 @@ func TestHasMinimumStake(t *testing.T) {
 	chain := local.Connect()
 	coreFirewall := newMockCoreFirewall()
 	policy := &stakeOrActiveKeepPolicy{
-		chain:                  chain,
-		minimumStakePolicy:     coreFirewall,
-		activeKeepMembersCache: cache.NewTimeCache(cacheLifeTime),
+		chain:                    chain,
+		minimumStakePolicy:       coreFirewall,
+		activeKeepMembersCache:   cache.NewTimeCache(cacheLifeTime),
+		noActiveKeepMembersCache: cache.NewTimeCache(cacheLifeTime),
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/firewall/firewall_test.go
```diff
@@ -320,16 +328,16 @@ func TestCachesAllActiveKeepMembers(t *testing.T) {
...
 		t.Errorf("should cache active keep members")
 	}
-	if cache.Has(closedKeepMembers[0].String()) {
+	if activeKeepMembersCache.Has(closedKeepMembers[0].String()) {
 		t.Errorf("should not cache non-active keep members")
 	}
-	if cache.Has(closedKeepMembers[1].String()) {
+	if activeKeepMembersCache.Has(closedKeepMembers[1].String()) {
 		t.Errorf("should not cache non-active keep members")
 	}
```
pdyraga, line 342: Shouldn't we add assertions for `noActiveKeepMembersCache`?

pkg/firewall/firewall_test.go
```diff
@@ -399,6 +408,70 @@ func TestSweepsCache(t *testing.T) {
 
 }
 
+// Has no minimum stake.
+// Has authorization.
+// Is not a member of an active keep.
+// Should NOT allow to connect.
+// After some time, becomes member of an active keep.
```
pdyraga, line 415: To be precise, it should be:
```
// After some time, has a minimum stake again and becomes a member of an active keep.
// Shortly after that, the minimum stake drops below the required minimum but
// the membership in an active keep remains.
```

pkg/firewall/firewall_test.go
lukasz-zimnoch: The `noActiveKeepMembersCache` works a bit differently. We can't put members of an inactive keep inside it because those members can be part of an other active keep. Because of that, just the checked member may be inserted into the `noActiveKeepMembersCache` so making similar assertons for this cache here is not possible.

pdyraga: Right. This is what I propose to assert. We can assert neither `activeKeepMembers` nor `closedKeepMembers` are in that cache but also that the guy for which we are executing the check - `key.NetworkPubKeyToEthAddress(remotePeerPublicKey)` - is there:
```
	// We don't put members of an active keep inside noActiveKeepMembersCache.
	// We don't put members of an inactive keep inside noActiveKeepMembersCache
	// because those members can belong to another active keep. We put there
	// only those members for which we executed the check.
	if noActiveKeepMembersCache.Has(activeKeepMembers[0].String()) {
		t.Errorf("should not cache member until all keeps are checked")
	}
	if noActiveKeepMembersCache.Has(activeKeepMembers[1].String()) {
		t.Errorf("should not cache member until all keeps are checked")
	}
	if noActiveKeepMembersCache.Has(closedKeepMembers[0].String()) {
		t.Errorf("should not cache member until all keeps are checked")
	}
	if noActiveKeepMembersCache.Has(closedKeepMembers[1].String()) {
		t.Errorf("should not cache member until all keeps are checked")
	}
	if !noActiveKeepMembersCache.Has(key.NetworkPubKeyToEthAddress(remotePeerPublicKey)) {
		t.Errorf("should be in the no active keep members cache")
	}
```

Issue #501, opened 2020-07-13
Migrate to OpenZeppelin test environment


pdyraga:
We should update `keep-ecdsa` tests to OpenZeppelin test environment just like we did in `keep-core` (https://github.com/keep-network/keep-core/issues/1528).

While working on this issue, let's also explore if it is possible to move stubs somewhere into `solidity/tests` directory tree, as proposed in https://github.com/keep-network/keep-core/pull/1739. 

References:
- https://github.com/keep-network/keep-core/pull/1533
- https://github.com/keep-network/tbtc/pull/441

eth-r:
It would be great if we could also migrate to a [parallel test runner, e.g. Jest](https://docs.openzeppelin.com/test-environment/0.1/choosing-a-test-runner) to make the tests even faster, as `keep-core` is getting rather slow again. Depending on the amount of work necessary for changing the runner (Jest seems quite similar to Mocha) it may be convenient to do it at the same time as the OZ migration.

dimpar:
Agree, `Jest` is worth trying.  Per their doc, there are 2 main differences. I believe that migrating `keep-ecdsa` to OZ would align all keep-network apps using the same testing env. Since [PR](https://github.com/keep-network/keep-ecdsa/pull/502) for OZ is already opened, I would explore `Jest` in a separate PR.

PR #502, opened 2020-07-15
Migration to openzeppelin test environment

dimpar:
Closes https://github.com/keep-network/keep-ecdsa/issues/501

Migrating tests to OpenZeppelin test environment. Summary of running times in the table below

|       | Truffle+Ganache | OpenZeppelin | Diff    |
|-------|---------|--------------|---------|
| Local | 2m22sec | 1m15sec      | 1m07sec |
| CI    | 3m18sec | 1m43sec      | 1m35sec |

In this PR I've tried to leave the `Stubs` under `/solidity/test` directory, but it seems that truffle won't put compiled Stubs artifacts to `build/contracts` which resulted in bunch of errors when running with OZ. Stubs were put under `/solidity/contracts/test` dir in order to run with OZ.


solidity/test/BondedECDSAKeepFactoryTest.js
```diff
@@ -1,32 +1,35 @@
...
 const {expectRevert, time} = require("@openzeppelin/test-helpers")
 
-import {mineBlocks} from "./helpers/mineBlocks"
+const {mineBlocks} = require("./helpers/mineBlocks")
 
 const truffleAssert = require("truffle-assertions")
 
-const KeepRegistry = artifacts.require("KeepRegistry")
-const BondedECDSAKeepFactoryStub = artifacts.require(
+const StackLib = contract.fromArtifact("StackLib")
```
pdyraga, line 10: What is `StackLib`? I cannot find it anywhere.

solidity/test/BondedECDSAKeepTest.js
```diff
@@ -1658,16 +1666,18 @@ contract("BondedECDSAKeep", (accounts) => {
...
       // Check balances of all keep members' and beneficiary.
       const newBalances = await getERC20BalancesFromList(accountsInTest, token)
       assert.equal(newBalances.toString(), expectedBalances.toString())
     })
 
     async function initializeTokens(token, keep, account, amount) {
-      await token.mint(account, amount)
-      await token.approve(keep.address, amount)
+      await token.mint(account, amount, {from: accounts[0]})
+      await token.approve(keep.address, amount, {from: accounts[0]})
```
pdyraga, line 1680: I think we should be doing `await token.approve(keep.address, amount, {from: account})` here.

We are minting tokens to `account` and that `account` should approve tokens.

solidity/test/BondedECDSAKeepVendorImplV1Test.js
```diff
@@ -1,14 +1,20 @@
...
-const BondedECDSAKeepVendorImplV1Stub = artifacts.require(
+const KeepRegistry = contract.fromArtifact("KeepRegistry")
+const BondedECDSAKeepVendorImplV1Stub = contract.fromArtifact(
   "BondedECDSAKeepVendorImplV1Stub"
 )
 
+const BN = web3.utils.BN
+const chai = require("chai")
+chai.use(require("bn-chai")(BN))
+const assert = chai.assert
```
pdyraga, line 14: We are not asserting any big numbers so I think it's enough to have:
```
const chai = require("chai")
const assert = chai.assert
```

solidity/test/BondedECDSAKeepVendorUpgradeTest.js
```diff
@@ -1,16 +1,22 @@
...
-const BondedECDSAKeepVendorImplV2Stub = artifacts.require(
+const BondedECDSAKeepVendorImplV2Stub = contract.fromArtifact(
   "BondedECDSAKeepVendorImplV2Stub"
 )
 
-contract("BondedECDSAKeepVendorUpgrade", async (accounts) => {
+const BN = web3.utils.BN
+const chai = require("chai")
+chai.use(require("bn-chai")(BN))
+const assert = chai.assert
```
pdyraga, line 17: Same here, it should be enough to have:
```
const chai = require("chai")
const assert = chai.assert
```

pdyraga:
Can we also set `local` network configuration in `truffle.js` back to:
```
    local: {
      host: "127.0.0.1",     // Localhost (default: none)
      port: 8546,            // Standard Ethereum port (default: none)
      network_id: "*",       // Any network (default: none)
      websockets: true,     // To enable EventEmitter interface for web3 change it to 'true'
    },
```

We no longer need Ganache for unit tests :tada: and having the config ready for smoke test will be helpful.

solidity/test/BondedECDSAKeepFactoryTest.js
dimpar: [StackLib](https://github.com/keep-network/sortition-pools/blob/master/contracts/StackLib.sol)

You'd get this if we won't link it:
```
Contract contains unresolved libraries. You must deploy and link the following libraries before you can deploy a new version of Contract: StackLib
```

pdyraga: Right, it's from sortition pools. I have no better idea of how to solve it, let's leave it as-is.

pdyraga: In `2_deploy_contract.js` we solve all the linkage problems with
```
const {
  deployBondedSortitionPoolFactory,
} = require("@keep-network/sortition-pools/migrations/scripts/deployContracts")
const BondedSortitionPoolFactory = artifacts.require(
  "BondedSortitionPoolFactory"
)
```

but it's not that easy here. That's fine, let's leave it as-is and approach it again in future in case we'll have any better idea.

solidity/test/BondedECDSAKeepTest.js
dimpar: Yup, you're right. Changed.

solidity/test/BondedECDSAKeepVendorImplV1Test.js
dimpar: Removed

pdyraga:
I love it.

Issue #506, opened 2020-07-21
Allow the application to configure expectations about the minimum bond


pdyraga:
Right now, bonded sortition pool updates the minimum bondable value for each successful signer set selection. This way, the gas cost of skipping ineligible signers can be minimized and the sortition pool automatically adjusts the requirements based on the current application's expectations.

Instead of doing it automatically, we could have the application set the minimum bondable value and leave the responsibility of keeping this value sane, balancing signer selection cost, to the application. For tBTC, this value could be the minimum lot size.

pdyraga:
@eth-r I am very interested in your thoughts here.

eth-r:
This is reasonable. With sortition pools being fairly optimized wrt querying costs, we should be fine even if we need to skip a lot of low-bond operators for larger lot sizes.

PR #508, opened 2020-07-23
Allow the application to set pool's minimum bondable value

pdyraga:
Refs: keep-network/keep-ecdsa#506
Depends on https://github.com/keep-network/sortition-pools/pull/84

`BondedSortitionPool` allows specifying a global minimum bondable value the operator needs to have so that it can join and stay in the pool. The responsibility for setting it to some reasonable value goes to the application.

We now have two minimum bond values in the sortition pool (see https://github.com/keep-network/sortition-pools/pull/84): one defining the minimum unbonded value the operator needs to have so that it can join and stay in the pool, and another defining the required bond value for the current selection. If the given operator has enough unbonded value to stay in the pool but it does not have enough unbonded value for the current group selection, it is
skipped.

Sortition pool is created with a minimum bond of 20 ETH to avoid small operators joining and griefing future selections before the minimum bond is set to the right value by the application. Anyone can create a sortition pool for an application with the default minimum bond value but the application can change this value later, at any point. 20 ETH per operator should be more than enough to cover 1 BTC deposit in 3-operators group.

The next step is going to be: modify tBTC to set the minimum bondable value for itself on initialization and for each lot size update.

### DIFF
--- a/solidity/contracts/BondedECDSAKeepFactory.sol
+++ b/solidity/contracts/BondedECDSAKeepFactory.sol
@@ -170,21 +170,17 @@ contract BondedECDSAKeepFactory is
         return candidatesPools[_application];
     }
 
-    /// @notice Gets the sortition pool address for the given application.
-    /// @dev Reverts if sortition does not exits for the application.
-    /// @param _application Address of the application.
-    /// @return Address of the sortition pool contract.
-    function getSortitionPool(address _application)
-        external
-        view
-        returns (address)
-    {
-        require(
-            candidatesPools[_application] != address(0),
-            "No pool found for the application"
+    /// @notice Sets the minimum bondable value required from the operator to
+    /// join the sortition pool of the given application. It is up to the
+    /// application to specify a reasonable minimum bond for operators trying to
+    /// join the pool to prevent griefing by operators joining without enough
+    /// bondable value.
+    /// @param _minimumBondableValue The minimum unbonded value allowing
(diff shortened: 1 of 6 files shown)


solidity/contracts/BondedECDSAKeepFactory.sol
```diff
@@ -384,6 +370,23 @@ contract BondedECDSAKeepFactory is
         groupSelectionSeed = _relayEntry;
     }
 
+    /// @notice Gets the sortition pool address for the given application.
+    /// @dev Reverts if sortition does not exits for the application.
```
Shadowfiend, line 374: exits->exist

solidity/test/BondedECDSAKeepFactoryTest.js
```diff
@@ -1491,6 +1491,27 @@ describe("BondedECDSAKeepFactory", function () {
     })
   })
 
+  describe("setMinimumBondableValue", async () => {
+    before(async () => {
+      await initializeNewFactory()
+      await initializeMemberCandidates()
+    })
+
+    it("fails for unknown application", async () => {
```
nkuba, line 1500: Can we add such a case for `registerMemberCandidate`? I believe we missed it before.

pdyraga: Sure! Added in https://github.com/keep-network/keep-ecdsa/pull/508/commits/74a261f8753517f629501f6c8c2d749c832edcb5.

Issue #500, opened 2020-07-12
Bonded sortition pool re-joining after running out of Eth


knarz:
Over the past couple of days I've observed that once a node gets kicked from the bonded sortition pool, because it ran out of Eth for bonding, it has a hard time rejoining. Calls to `RegisterAsMemberCandidate` in `pkg/chain/ethereum/ethereum.go` are queued, I think, but it seems like they're never executed even if more Eth are added for bonding.

1. Run out of Eth for bonding
2. Operator gets kicked from bonded sortition pool
3. Node logs `Operator not eligible` errors
4. Add more Eth for bonding
5. Node continues to log errors

A restart of the node remedies this, which is not perfect.

I was also able to fix it by just calling `registerMemberCandidate` on chain via an external script. This returns the node to normal operation but also requires outside intervention

mhluongo:
Thank you for doing the legwork here @knarz! We'll check it out

pdyraga:
I tried to replicate this problem with no luck, I see the client behaves as expected.

[Looking at the code in `pkg/client/registration.go`](https://github.com/keep-network/keep-ecdsa/blob/4cf63d2e018bc0da8e49c1d2302e0c3fbdc82fdf/pkg/client/registration.go), `checkStatusAndRegisterForApplication` registers the client to the application and triggers status monitoring; in the case when `monitorSignerPoolStatus` exits with an error - and it happens when the client is kicked out of the sortition pool - it tries to register the operator again and monitors the progress every block in `waitUntilRegistered`. From the code perspective, it all looks as expected.

I created two simple scripts - one withdrawing unbonded value from `KeepBonding` and another depositing 50 ETH to `KeepBonding`. I executed withdraw-deposit cycle multiple times and the client handled it all correctly, automatically registering the staker back to the pool. I have also executed the smoke test opening a keep and requesting a signature from that keep and I see that the client who registered back to the sortition pool was properly selected to the signer set and cooperated with other clients.

```
14:38:58.060 DEBUG keep-ecdsa: operator status check for application [0x4f7C771Ab173bEc2BbE980497111866383a21172] triggered at block [5266] registration.go:241
14:38:58.061 DEBUG keep-ecdsa: operator status is up to date for application [0x4f7C771Ab173bEc2BbE980497111866383a21172] registration.go:258
[...]

PR #509, opened 2020-07-24
Add 20% safety margin to sortition pool registration TX gas limit

pdyraga:
Closes: #500 

If we have multiple sortition pool join transactions queued - and that happens when multiple operators get eligible to join, e.g. after lowering the minimum bond requirement - transactions mined at the end may no longer have valid gas limits as they were estimated based on a different state of the pool. We add 20% safety margin to the original gas estimation to account for that.

### DIFF
--- a/pkg/chain/ethereum/ethereum.go
+++ b/pkg/chain/ethereum/ethereum.go
@@ -35,11 +35,11 @@ func (ec *EthereumChain) RegisterAsMemberCandidate(application common.Address) e
 	}
 
 	// If we have multiple sortition pool join transactions queued - and that
-	// happens when multiple operators gets eligible to join, e.g. after
-	// lowering the minimum bond requirement - transactions mined at the end may
-	// no longer have valid gas limits as they were estimated based on a
-	// different state of the pool. We add 20% safety margin to the original gas
-	// estimation to account for that.
+	// happens when multiple operators become eligible to join at the same time,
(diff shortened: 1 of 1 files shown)


pkg/chain/ethereum/ethereum.go
```diff
@@ -29,8 +29,23 @@ func (ec *EthereumChain) Address() common.Address {
 // RegisterAsMemberCandidate registers client as a candidate to be selected
 // to a keep.
 func (ec *EthereumChain) RegisterAsMemberCandidate(application common.Address) error {
+	gasEstimate, err := ec.bondedECDSAKeepFactoryContract.RegisterMemberCandidateGasEstimate(application)
+	if err != nil {
+		return fmt.Errorf("failed to estimate gas [%v]", err)
+	}
+
+	// If we have multiple sortition pool join transactions queued - and that
+	// happens when multiple operators gets eligible to join, e.g. after
```
dimpar, line 38: "..operators gets eligible.." -> maybe just "operators are eligible.."

pdyraga: Maybe *become eligible to join at the same time*? I intend to express that they become eligible at the same time, that's why transactions land in the mempool more or less at the same time.

dimpar: sure, this sounds better

PR #510, opened 2020-07-29
Calculate the required unbonded value per keep member

pdyraga:
Refs keep-network/keep-ecdsa#506

Instead of expecting the application (sortition pool owner) to calculate
the minimum unbonded value and perform all the necessary rounding, we
now calculate everything in BondedECDSAKeepFactory based on the provided
bond required from a keep and the number of signers in that keep.

Although we currently do not use threshold when calculating the required bond,
we want the applications to pass this parameter so that we can later add
proper support for n>t keeps without changing the public API.

### DIFF
--- a/solidity/contracts/BondedECDSAKeepFactory.sol
+++ b/solidity/contracts/BondedECDSAKeepFactory.sol
@@ -173,7 +173,7 @@ contract BondedECDSAKeepFactory is
     /// bondable value.
     /// @param _minimumBondableValue The minimum bond value the application
     /// requires from a single keep.
-    /// @param _groupSize Number of singers in the keep.
+    /// @param _groupSize Number of signers in the keep.
     /// @param _honestThreshold Minimum number of honest keep signers.
     function setMinimumBondableValue(
         uint256 _minimumBondableValue,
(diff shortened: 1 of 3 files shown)


solidity/contracts/BondedECDSAKeepFactory.sol
```diff
@@ -171,11 +171,18 @@ contract BondedECDSAKeepFactory is
     /// application to specify a reasonable minimum bond for operators trying to
     /// join the pool to prevent griefing by operators joining without enough
     /// bondable value.
-    /// @param _minimumBondableValue The minimum unbonded value allowing
-    /// an operator to join and stay in the sortition pool for the application.
-    function setMinimumBondableValue(uint256 _minimumBondableValue) external {
+    /// @param _minimumBondableValue The minimum bond value the application
+    /// requires from a single keep.
+    /// @param _groupSize Number of singers in the keep.
```
Shadowfiend, line 176: signers*

pdyraga: 🧑‍🎤

PR #511, opened 2020-07-30
Try to get sortition pool before creating a new one

lukasz-zimnoch:
Sortition pool existence should be checked before trying to create a new one during `lcl-initialize.js` script execution. This is a must when executing `local-setup` scripts because keep-network/tbtc#715 introduced sortition pool creation during `tbtc` contract migration. If we always create a sortition pool in that case, we can expect the `lcl-initialize.js` to fail.

### DIFF
--- a/solidity/scripts/lcl-initialize.js
+++ b/solidity/scripts/lcl-initialize.js
@@ -76,16 +76,31 @@ module.exports = async function () {
     }
 
     try {
-      await bondedECDSAKeepFactory.createSortitionPool(application)
-      console.log(`created sortition pool for application: [${application}]`)
+      console.log(`getting sortition pool for application [${application}]`)
 
       sortitionPoolAddress = await bondedECDSAKeepFactory.getSortitionPool(
-        application
+          application
       )
     } catch (err) {
-      console.error("failed to create sortition pool", err)
-      process.exit(1)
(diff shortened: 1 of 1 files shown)


solidity/scripts/lcl-initialize.js
```diff
@@ -76,17 +76,32 @@ module.exports = async function () {
...
-      console.log(`created sortition pool for application: [${application}]`)
+      console.log(`getting sortition pool for application [${application}]`)
 
       sortitionPoolAddress = await bondedECDSAKeepFactory.getSortitionPool(
         application
       )
     } catch (err) {
-      console.error("failed to create sortition pool", err)
-      process.exit(1)
+      console.error("failed to get sortition pool", err, "; creating a new one")
```
nkuba, line 85: Should we do `process.exit(1)` in the next line?

lukasz-zimnoch: But we don't :)

nkuba: So shouldn't we?

lukasz-zimnoch: Nope, if we catch an error from `getSortitionPool` we should try to create a new one instead of exit.

nkuba: 🤦 Please forget

PR #513, opened 2020-07-31
ECDSA keep reward specialization

eth-r:
New implementation of ECDSA keep rewards based on the abstract rewards contract.


solidity/contracts/ECDSAKeepRewards.sol
```diff
@@ -0,0 +1,96 @@
+pragma solidity ^0.5.4;
+
+import "openzeppelin-solidity/contracts/token/ERC20/IERC20.sol";
+import "openzeppelin-solidity/contracts/token/ERC20/SafeERC20.sol";
+import "openzeppelin-solidity/contracts/math/SafeMath.sol";
+import "@keep-network/keep-core/contracts/Rewards.sol";
+
+contract ECDSAKeepRewards is Rewards {
+    uint256 constant MAX_UINT20 = 1 << 20 - 1;
```
benlongstaff, line 9: where does this get used? I couldn't find it in keep-ecdsa or keep-core?

solidity/contracts/ECDSAKeepRewards.sol
```diff
@@ -0,0 +1,96 @@
...
+    uint256 constant _minimumKeepsPerInterval = 2;
```
benlongstaff, line 10: should this number be higher? i'm totally ok with it staying at 2 though

solidity/contracts/ECDSAKeepRewards.sol
```diff
@@ -0,0 +1,96 @@
...
+    IBondedECDSAKeepFactory factory;
+
+    constructor (
+        address _token,
+        address factoryAddress,
+        uint256 _firstIntervalStart,
+        uint256[] memory _intervalWeights
```
benlongstaff, line 17: why is intervalWeights as a constructor param and _minimumKeepsPerInterval a constant?

solidity/contracts/ECDSAKeepRewards.sol
pdyraga: Same here, it looks unused so I'll drop it.

solidity/contracts/ECDSAKeepRewards.sol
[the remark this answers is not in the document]
pdyraga: It's going to be higher, I am pretty sure this value is just a temporary one. I'll drop a TODO here.

solidity/contracts/ECDSAKeepRewards.sol
pdyraga: Good point. I think we are going to make all those parameters constants to minimize the risk of making a mistake during deployment.

pdyraga:
Closing in favor of https://github.com/keep-network/keep-ecdsa/pull/567 and https://github.com/keep-network/keep-ecdsa/pull/571 that were based on this code.

PR #516, opened 2020-08-05
Repair eth_connectivity metric

lukasz-zimnoch:
`eth_connectivity` metric didn't work properly for `keep-ecdsa` because it invoked the `hasMinimumStake` method using a non-existing address taken from a non-existing config property. This PR fixes the problem by taking the address directly from the parsed key.


pdyraga:
Side note, we should remove the address property from `keep-core` config in a separate PR and use the same approach as in `keep-ecdsa`. I'll fill an issue for that.

PR #517, opened 2020-08-05
Repair lcl-initialize script

lukasz-zimnoch:
The sortition pool creation logic in `lcl-initialize.js` script, didn't consider a case when the `getSortitionPool` method returned a zero address. This PR adds some improvements to that logic.

### DIFF
--- a/solidity/scripts/lcl-initialize.js
+++ b/solidity/scripts/lcl-initialize.js
@@ -82,9 +82,18 @@ module.exports = async function () {
         application
       )
     } catch (err) {
-      console.error("failed to get sortition pool", err, "; creating a new one")
+      console.error("failed to get sortition pool", err)
+    }
 
+    if (
+      !sortitionPoolAddress ||
+      sortitionPoolAddress === "0x0000000000000000000000000000000000000000"
+    ) {
       try {
+        console.error(
+          `no sortition pool for application: [${application}]; creating a new one`
+        )
+
         await bondedECDSAKeepFactory.createSortitionPool(application)
 
         console.log(`created sortition pool for application: [${application}]`)
--- a/solidity/scripts/lcl-initialize.js
+++ b/solidity/scripts/lcl-initialize.js
@@ -86,8 +86,8 @@ module.exports = async function () {
     }
 
     // FIXME: Workaround for https://github.com/ethereum/go-ethereum/pull/21083 - instead of failing,
-    // contract call returns address 0x0. Once we are able to update to the most recent abigen 
-    // (see https://github.com/keep-network/keep-common/issues/45), the additional case for 0x0 address 
+    // contract call returns address 0x0. Once we are able to update to the most recent abigen
+    // (see https://github.com/keep-network/keep-common/issues/45), the additional case for 0x0 address
     // should be removed.
     if (
       !sortitionPoolAddress ||
--- a/solidity/scripts/lcl-initialize.js
+++ b/solidity/scripts/lcl-initialize.js
@@ -85,6 +85,10 @@ module.exports = async function () {
       console.error("failed to get sortition pool", err)
     }
 
+    // FIXME: Workaround for https://github.com/ethereum/go-ethereum/pull/21083 - instead of failing,
+    // contract call returns address 0x0. Once we are able to update to the most recent abigen 
+    // (see https://github.com/keep-network/keep-common/issues/45), the additional case for 0x0 address 
+    // should be removed.
     if (
       !sortitionPoolAddress ||
       sortitionPoolAddress === "0x0000000000000000000000000000000000000000"


pdyraga:
If there is no sortition pool for the application, `getSortitionPool` call should fail:

https://github.com/keep-network/keep-ecdsa/blob/8c809460a740ac4794fb7e38514879cbb3e74787/solidity/contracts/BondedECDSAKeepFactory.sol#L384-L387

In what circumstances is the zero address returned?

lukasz-zimnoch:
> If there is no sortition pool for the application, `getSortitionPool` call should fail:
> 
> https://github.com/keep-network/keep-ecdsa/blob/8c809460a740ac4794fb7e38514879cbb3e74787/solidity/contracts/BondedECDSAKeepFactory.sol#L384-L387
> 
> In what circumstances is the zero address returned?

I observed this while initializing a local `keep-ecdsa` setup.

pdyraga:
> I observed this while initializing a local `keep-ecdsa` setup.

Is it something that's happening for you all the time? I've just done a local setup from scratch (with `install.sh`) and it works for me. Looking at the `master` version of the file, we should never have `address(0x0)` there assuming `createSortitionPool` one line earlier did not fail.

Or maybe the issue is when the environment is not set up from scratch?

lukasz-zimnoch:
> > I observed this while initializing a local `keep-ecdsa` setup.
> 
> Is it something that's happening for you all the time? I've just done a local setup from scratch (with `install.sh`) and it works for me. Looking at the `master` version of the file, we should never have `address(0x0)` there assuming `createSortitionPool` one line earlier did not fail.
> 
> Or maybe the issue is when the environment is not set up from scratch?

Yeah, I observed this several times. Anyway, changes from this PR handles both scenarios.

lukasz-zimnoch:
Checked once again on fresh master (`38bc2571`). Executed a fresh deploy using `install.sh` and then `initialize.sh` using the default inputs:
```
Initializing contracts...
Using network 'local'.

Browserslist: caniuse-lite is outdated. Please run next command `npm update`
getting sortition pool for application [0x1a12D46b3648FF04d169F13cd478658590525a0f]
sortition pool for application [0x1a12D46b3648FF04d169F13cd478658590525a0f] is [0x0000000000000000000000000000000000000000]
```

pdyraga:
`sortitionPoolAddress` in the script is `0x0000000000000000000000000000000000000000` only if this value was assigned by `getSortitionPool`. I see no other assignment there. It means that `getSortitionPool` returned this value instead of failing. But we do have tests making sure `getSortitionPool` reverts for an unknown application and those tests are passing:

https://github.com/keep-network/keep-ecdsa/blob/7ccb09bb526d78df638571433cc9751516a9b3c8/solidity/test/BondedECDSAKeepFactoryTest.js#L278-L283

I made some investigation and it seems that the reason we are seeing `0x0` here is the fact `view` calls are not failing on revert and it's been fixed here: https://github.com/ethereum/go-ethereum/pull/21083. I've tested it with geth `1.9.12` and I was also getting `0x0` instead of a failure. Then, I updated my `geth` to `v1.9.18` and it works fine now - the function fails instead of returning `0x0`.

I think we should stick with the approach presented in #511 and use the most recent `geth` version for development. I don't like having the separate if for `0x0` case because it's something different from what we expect in unit tests.

lukasz-zimnoch:
Nice findings! However, I think the `lcl-initialize.js` is rather a convenience script for developers who may use different versions of `geth` and it isn't something bad if we handle all possible cases. If we don't handle the `0x0` case, I bet someone will waste their time soon trying to debug what wrong happened during the `initialize.sh` script  😀 .  Also, this PR doesn't change the behavior introduced by #511 because if you receive a revert, the `sortitionPoolAddress` variable will be undefined an also handled by the conditional sortition pool creation. 

By the way, did you try `go generate ./...` having the most recent version of `geth`? IIRC I observed some problems with the generated ABIs using a version above `1.9.12`

pdyraga:
I don't think we need to support all versions of `geth`, it is important that we support the most recent one. Live goes by, bugs are introduced and fixed. It's the same situation as with the incorrect gas estimates because of not passing gas price in the estimation context. After it's been fixed in `geth`, we waited for a while and removed the workarounds we had in the code.

The special case for `0x0` is nothing more than the workaround for https://github.com/ethereum/go-ethereum/pull/21083. We expect this code to fail when sortition pool does not exist and the only reason it does not fail is the mentioned issue in `go-ethereum`.

I updated `abigen` to the most recent version (`1.9.18`) as you suggested and indeed, there is some problem with generating contract bindings. I opened an issue for that: https://github.com/keep-network/keep-common/issues/45. It worked for me before because I used `geth` client in version `1.9.18` but I was generating contract bindings using abigen `1.9.10`.

If we were able to update to `1.9.18`, I'd suggest to update `geth`, mention it explicitly in development docs that at least `1.9.18` is required, and remove the special case for `0x0` because it is misleading. However, given that we cannot simply update to `1.9.18` - at least without expecting contributors to have another version of `abigen` what greatly complicates things - I think we should merge this change but add a `FIXME` next to the `0x0` case explaining why it's there. Without additional context it's really misleading because it's not what the contract is supposed to do and it's not what we expect in unit tests. For example:
```
// FIXME: Workaround for https://github.com/ethereum/go-ethereum/pull/21083 - instead of failing,
// contract call returns address 0x0. Once we are able to update to the most recent geth & abigen 
// (see https://github.com/keep-network/keep-common/issues/45), the additional case for 0x0 address 
// should be removed.
```

pdyraga:
Linter is not happy about the changes:
```
/home/circleci/project/solidity/scripts/lcl-initialize.js
  89:95   error  Delete `·`  prettier/prettier
  90:104  error  Delete `·`  prettier/prettier
```

PR #522, opened 2020-08-07
DefaultMaxGasPrice updated to 70Gwei

pdyraga:
Given the current trends of gas prices, it makes sense to increase the `DefaultMaxGasPrice` to `70Gwei`. As for today, Etherscan indicates `60Gwei` is the safe price and `75Gwei` is the proposed price, both with ~20 seconds expected mining time of a TX.


pdyraga:
Updated the default value in `config.toml.SAMPLE` as @dimpar caught in https://github.com/keep-network/keep-core/pull/1975#issuecomment-670422813.

PR #525, opened 2020-08-11
Abstract bonding contract

nkuba:
This PR is preparation for ETH only staking. We extracted some common parts of `KeepBonding` contract that almost entirely can be used for ETH bonding and placed them in `AbstractBonding`. The only difference, for now, is `withdraw` that in case of KEEP token staking has to support grantee roles check. This is an abstract function in `AbstractBonding` implemented by `KeepBonding`.


solidity/contracts/AbstractBonding.sol
```diff
@@ -0,0 +1,357 @@
...
+    using SafeMath for uint256;
+
+    // Registry contract with a list of approved factories (operator contracts).
+    KeepRegistry internal registry;
+
+    // Staking Authorizations contract.
+    Authorizations internal authorizations;
+
+    // Stake Delegatable contract.
+    StakeDelegatable internal stakeDelegatable;
```
pdyraga, line 36: Those two are a replacement for `TokenStaking` contract reference?

I was thinking about defining an interface in this file for staking contract containing both `Authorizations` and `StakeDelegatable` required function declarations but I am not sure how many function we'd have to delcare there and whether the current approach is not cleaner. What do you think?

solidity/contracts/AbstractBonding.sol
```diff
@@ -0,0 +1,357 @@
...
+        address indexed operator,
+        uint256 indexed referenceID,
+        address destination,
+        uint256 amount
+    );
+
+    /// @notice Initializes Keep Bonding contract.
+    /// @param registryAddress Keep registry contract address.
+    /// @param authorizationsAddress Staking Authorizations contract address.
+    /// @param stakeDelegatableAddress Stake Delegatable contract address.
```
pdyraga, line 83: Technically, it's just one contract, that's why accepting two parameters looks a little weird to me.

solidity/contracts/AbstractBonding.sol
pdyraga: I think we are talking about just 3 functions:
- `isAuthorizedForOperator`
- `beneficiaryOf`
- `authorizerOf`

So I think delcaring `ITokenStaking` with those three functions on a top of this file could be a little cleaner.

nkuba: > Those two are a replacement for TokenStaking contract reference?

Yes. `TokenStaking` implements two contracts which functions we use here. We already declared functions in `Authorizations` and `StakeDelegatable`. If we decide to define an interface we have to update `TokenStaking` to implement that interface, because I don't like the idea of using an interface that is not explicitly implemented by `TokenStaking`.

pdyraga: It is quite a complex situation. We may define `ITokenStaking` in `keep-core` but then we need to decide which functions should go there.

We could include there all `public` functions of `TokenStaking` contract. That's not the best option because `lockStake` and `unlockStake` are `public` in `TokenStaking` and the ETH-only staking (bonding) contract will not use locks.

We could include there all `public view` functions of `TokenStaking` contract. That's also not the best option for similar reasons - `isStakeLocked` and `getLocks` are `public view` functions in `TokenStaking` and the ETH-only staking (bonding) contract will not use locks.

We could include there all `public view` functions from `StakeDelegatable` and `Authorizations` but it's also less than perfect from `keep-core`'s perspective, we are limiting the public API of `TokenStaking` because some contract - ETH only staking (bonding) contract - in the repository `keep-core` is not even aware of does not need those functions.

Now, looking at the current situation, we don't have any deployment-time guarantee the address of the contract for `stakeDelegatable` reference and `authorizations` reference will be a deployed instance of `Authorizations` or `StakeDelegatable`. We don't have a compile-time check as well. Unit tests check the code works but they work against contracts we deploy during the test set-up. If we screw anything during the mainnet deployment, we'll be busted with the current code anyway.

Given that, I think there is no much difference between declaring this interface here and using `Authorizations` / `StakeDelegatable` references. The only guarantee are proper unit tests implemented against correct contract versions. That's a language/tooling-level issue I am afraid we can't solve here. To make us feel a little better I think we can use `AuthorityVerifier` interface we already have in `keep-core` and define `IDelegationInfo` interface for `StakeDelegatable` with all `public view` functions from there in `keep-core`. Then, we can combine those two interfaces together into `IStaking` but `IStaking` should be declared in `keep-ecdsa`. This interface describes the expectations of `keep-ecdsa`.

Thoughts on that?

solidity/contracts/AbstractBonding.sol
pdyraga: We have this discussion [here](https://github.com/keep-network/keep-ecdsa/pull/525#discussion_r468477260) so I'll resolve this conversation.

solidity/contracts/AbstractBonding.sol
nkuba: > To make us feel a little better I think we can use AuthorityVerifier interface we already have in keep-core

`AuthorityVerifier` doesn't define `isAuthorizedForOperator` function currently, we could add this.

> define IDelegationInfo interface for StakeDelegatable with all public view functions from there in keep-core

I'm not sure about a point of duplicating the code from the `StakeDelegatable` in `IDelegationInfo`. Why can't we use 
`StakeDelegatable` directly?

> Then, we can combine those two interfaces together into IStaking but IStaking should be declared in keep-ecdsa

It would make sense if only we could do this. Interface inheritance was [introduced in solidity 0.6](https://github.com/ethereum/solidity/issues/3419), do you have in mind any specific workaround for the solidity version we use?

pdyraga: I was sure we can inherit from multiple interfaces in Solidity 5, I was wrong. We can probably do some tricks with abstract contracts but I am afraid it will only make the code even more complex.

Knowing that we can't have two interfaces, one for authorizations and one for delegation info, and later combining them together as well as knowing that defining the good interface on `keep-core` is hard, if possible at all given that we need just three functions here, what if we use the fact `AbstractBonding` is abstract and declare three abstract functions on it (`beneficiaryOf`, `authorizerOf`, `isAuthorizedForOperator`) and have the implementations of `AbstractBonding` implement those functions?

It looks cleaner to me than the current approach. My reasoning:
- `StakeDelegatable` and `Authorizations` are kind of an implementation detail of `TokenStaking` rather than a public API and we shouldn't use them as a public API.
- ETH-only staking is actually not about staking but about bonding. I am not sure how you plan to proceed with the implementation of owner-operator-authorizer-beneficiary relationship but it looks to me that the best option could be to have this code implemented in ETH-specific version of `AbstractBonding`.
- Having two references to one contract is acceptable but weird and may be tricky for anyone not very familiar with this code.
- We can have simpler tests, we shouldn't need the two additional stubs.

PR #526, opened 2020-08-12
Extracted random beacon consumer from keep factory

nkuba:
Some part of the code in ECDSA keep factory related to the random beacon seeding could be extracted to a separate contract that is inherited by the factory. This operation is one of the extractions from the factory to create an abstract contract that can be used for both KEEP and ETH staking.

Refs: #271

### DIFF
--- a/solidity/contracts/KeepFactoryBeaconConsumer.sol
+++ b/solidity/contracts/KeepFactoryBeaconConsumer.sol
@@ -1,0 +1,144 @@
+/**
+▓▓▌ ▓▓ ▐▓▓ ▓▓▓▓▓▓▓▓▓▓▌▐▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓ ▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓ ▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▄
+▓▓▓▓▓▓▓▓▓▓ ▓▓▓▓▓▓▓▓▓▓▌▐▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓ ▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓ ▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓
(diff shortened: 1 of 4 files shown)


solidity/contracts/KeepFactoryBeaconConsumer.sol
```diff
@@ -0,0 +1,144 @@
...
+                           Trust math, not hardware.
+*/
+
+pragma solidity 0.5.17;
+
+import "@keep-network/keep-core/contracts/IRandomBeacon.sol";
+
+import "openzeppelin-solidity/contracts/math/SafeMath.sol";
+
+contract KeepFactoryBeaconConsumer is IRandomBeaconConsumer {
```
pdyraga, line 21: Thoughts on naming this contract `KeepFactoryReseed`? The fact it's using the random beacon under the hood is an implementation detail. What matters from `BondedECDSAKeepFactory`'s perspective is that this abstract class offers "reseed" capabilities.

solidity/contracts/KeepFactoryBeaconConsumer.sol
```diff
@@ -0,0 +1,144 @@
...
+    using SafeMath for uint256;
+
+    IRandomBeacon randomBeacon;
```
pdyraga, line 24: Could be wrong but I think linter will complain about having this field before constants.

solidity/contracts/KeepFactoryBeaconConsumer.sol
```diff
@@ -0,0 +1,144 @@
...
+    /// @notice Gets a fee estimate for a new random entry.
+    /// @return Uint256 estimate.
+    function newEntryFeeEstimate() public view returns (uint256) {
+        return randomBeacon.entryFeeEstimate(callbackGas);
+    }
+
+    /// @notice Calculates the fee requestor has to pay to reseed the factory
+    /// for signer selection. Depending on how much value is stored in the
+    /// reseed pool and the price of a new relay entry, returned value may vary.
+    function newGroupSelectionSeedFee() public view returns (uint256) {
```
nkuba, line 72: I was thinking about renaming some of the functions to drop the relation to group selection to have this contract more generic. But this might require exposing functions with old names in the factory and many changes in docs and namings in this contract.

solidity/contracts/KeepFactoryBeaconConsumer.sol
nkuba: Thoughts on `KeepFactorySeed` or `KeepFactorySeeder`?

solidity/contracts/KeepFactoryBeaconConsumer.sol
pdyraga: I think the current naming is fine, it's not worth it. The only seeding/reseeding that may take place here is about group selection.

PR #529, opened 2020-08-12
Extract generic keep factory code from Bonded ECDSA Keep Factory

nkuba:
Extracted some parts of the code that are generic for keep factory that uses clone factory to create keep instance and stores keeps data along with creations timestamps.

Refs: #271

Should be reviewed after https://github.com/keep-network/keep-ecdsa/pull/528 is merged to master.


solidity/contracts/KeepFactory.sol
```diff
@@ -0,0 +1,68 @@
...
+▓▓▓▓▓▓▓▓▓▓ ▓▓▓▓▓▓▓▓▓▓ ▐▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓ ▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓  ▓▓▓▓▓▓▓▓▓▓
+
+                           Trust math, not hardware.
+*/
+
+pragma solidity 0.5.17;
+
+import "./CloneFactory.sol";
+
+contract KeepFactory is CloneFactory {
```
pdyraga, line 19: What do you think about renaming it to `KeepFactoryMasterClone`?

Then, I'd do:
```
contract BondedECDSAKeepFactory is
    IBondedECDSAKeepFactory,
    KeepFactoryMasterClone
    KeepFactorySeed,
    KeepFactoryCandidatesPools,
    AuthorityDelegator
```

My reasoning is that we are not doing anything else than creating clones of the master contract here. Created keeps are not yet initialized and not ready to be used so it's nothing else than a master clone functionality plus some bookkeeping of those clones.

pdyraga: We had this conversation out of band and we decided to strip `KeepFactory*` prefix from all abstract classes and name them by the functionality they offer: `CandidatesPool, GroupSelectionSeed, KeepCreator`. We like the last one because it clearly expresses this is only about creating things and the initialization has to be done someone else. Also, the fact creation uses master contract cloning is just an implementation detail.

PR #530, opened 2020-08-12
Extracted parts of ECDSA keep from Bonded ECDSA Keep

nkuba:
In this PR we extract some parts of the code of BondedECDSAKeep to AbstractBondedECDSAKeep.
It also introduces IBondingManagement to interact with the bonding contract from the keep.

I suggest reviewing this PR commit by commit.

Should be reviewed after #529 is merged to master.

Refs: #271


solidity/contracts/AbstractBondedECDSAKeep.sol
```diff
@@ -0,0 +1,121 @@
...
+import "./api/IBondedECDSAKeep.sol";
+import "./AbstractECDSAKeep.sol";
+import "./AbstractBonding.sol";
+
+import "openzeppelin-solidity/contracts/math/SafeMath.sol";
+
+contract AbstractBondedECDSAKeep is IBondedECDSAKeep, AbstractECDSAKeep {
+    using SafeMath for uint256;
+
+    AbstractBonding bonding;
```
pdyraga, line 26: I don't like the fact we are working against an abstract contract. Abstract contract is a kind of an implementation detail. I think we should rather work against an interface or against a final implementation.

We already have `IBonding` interface defined. We can add `bondAmount`, `seizeBond`, `freeBond` and `deposit` there. Alternatively, and maybe even better, we can have an interface `IBondingManagement` with `seizeBond`, `freeBond`, `deposit` functions, and inheriting from `IBonding`. WDYT?

solidity/contracts/AbstractECDSAKeep.sol
```diff
@@ -0,0 +1,559 @@
...
+            _r,
+            _s,
+            _signedDigest,
+            _preimage
+        );
+
+        require(isFraud, "Signature is not fraudulent");
+
+        if (!fraudulentPreimages[_preimage]) {
+            punishMembersForSignatureFraud();
```
pdyraga, line 372: I'd name it `slashForSignatureFraud`. We can apply any slashing strategy we want slash stake/seize bond, do both.

solidity/contracts/AbstractBondedECDSAKeep.sol
```diff
@@ -0,0 +1,121 @@
...
+
+pragma solidity 0.5.17;
+
+import "./api/IBondedECDSAKeep.sol";
+import "./AbstractECDSAKeep.sol";
+import "./AbstractBonding.sol";
+
+import "openzeppelin-solidity/contracts/math/SafeMath.sol";
+
+contract AbstractBondedECDSAKeep is IBondedECDSAKeep, AbstractECDSAKeep {
```
pdyraga, line 23: I think we are going too far with `AbstractBondedECDSAKeep`. It's nice we are having bond-related logic separated but it's getting too complex and it's really hard to me to figure out what is where, not even speaking about developers not that familiar with the codebase.

If the ETH-only version will extend `AbstractBondedECDSAKeep`, let's have only `AbstractBondedECDSAKeep` and move all `AbstractECDSAKeep` logic here.

If the ETH-only version will extend `AbstractECDSAKeep`, let's have only `AbstractECDSAKeep` and move all `AbstractBondedECDSAKeep` logic down to `BondedECDSAKeep`.

solidity/contracts/AbstractBondedECDSAKeep.sol
nkuba: As `IBonding` is dedicated for sortition pools, I'll create `IBondingManagement`.

nkuba: I'm wondering if we should also add `createBond` and `reassignBond` that are also defined in the bonding contract and are related to other ones we will expose in the new interface.

solidity/contracts/api/IBondingManagement.sol
```diff
@@ -0,0 +1,84 @@
...
+▓▓▓▓▓▓▓▓▓▓ ▓▓▓▓▓▓▓▓▓▓ ▐▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓ ▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓  ▓▓▓▓▓▓▓▓▓▓
+
+                           Trust math, not hardware.
+*/
+
+pragma solidity 0.5.17;
+
+import "@keep-network/sortition-pools/contracts/api/IBonding.sol";
+
+contract IBondingManagement is IBonding {
```
pdyraga, line 19: And this is an abstract contract because interface cannot extend an interface in Solidity 5?

solidity/contracts/AbstractBondedECDSAKeep.sol
```diff
@@ -0,0 +1,643 @@
...
+    }
+
+    /// @notice Initialization function.
+    /// @dev We use clone factory to create new keep. That is why this contract
+    /// doesn't have a constructor. We provide keep parameters for each instance
+    /// function after cloning instances from the master contract.
+    /// @param _owner Address of the keep owner.
+    /// @param _members Addresses of the keep members.
+    /// @param _honestThreshold Minimum number of honest keep members.
+    function initializeBondedECDSAKeep(
```
pdyraga, line 533: What do you think about naming it just `initialize`? It's kinda weird that `initialize` in `BondedECDSAKeep` calls `initializeBondedECDSAKeep` that is in `AbstractBondedECDSAKeep`.

solidity/contracts/AbstractBondedECDSAKeep.sol
```diff
@@ -0,0 +1,643 @@
...
+        address _owner,
+        address[] memory _members,
+        uint256 _honestThreshold,
+        address _bonding
```
pdyraga, line 537: Can we name this parameter `_keepBonding` or rename `_keepBonding` parameter in `BondedECDSAKeep` to `_bonding`.

solidity/contracts/BondedECDSAKeep.sol
```diff
@@ -527,127 +78,48 @@ contract BondedECDSAKeep is IBondedECDSAKeep {
...
+            _keepBonding
+        );
 
-        owner = _owner;
-        members = _members;
-        honestThreshold = _honestThreshold;
         memberStake = _memberStake;
         tokenStaking = TokenStaking(_tokenStaking);
-        keepBonding = KeepBonding(_keepBonding);
         keepFactory = BondedECDSAKeepFactory(_keepFactory);
```
pdyraga, line 90: We don't need `keepFactory` field, we can delete this line. All we need the `keepFactory` for here is to `claimDelegatedAuthority`. It's nothing new, we should clean this up earlier but now we have the opportunity.

solidity/contracts/AbstractBondedECDSAKeep.sol
```diff
@@ -0,0 +1,643 @@
...
+    /// @return Beneficiary address.
+    function beneficiaryOf(address _member)
+        internal
+        view
+        returns (address payable);
+
+    /// @notice Terminates the keep.
+    /// Keep can be marked as terminated only when there is no signing in progress
+    /// or the requested signing process has timed out.
+    function terminateKeep() internal;
```
pdyraga, line 621: `terminateKeep` is declared here but implemented in `BondedECDSAKeep`. The implementation in `BondedECDSAKeep` calls back `markAsTerminated` defined here. We have a similar situation for `closeKeep`. I find it an antipattern. I think we should rather override function in inheriting class and add additional behavior there, where needed.

I propose to implement both `terminateKeep` and `closeKeep` here:
```
    function terminateKeep() internal {
        markAsTerminated();
    }

    function closeKeep() public onlyOwner onlyWhenActive {
        markAsClosed();
        freeMembersBonds();
    }
```

and add the required stake logic specific to KEEP version of keep in `BondedECDSAKeep`:
```
    function terminateKeep() internal {
        super.terminateKeep();
        unlockMemberStakes();
    }

    function closeKeep() public onlyOwner onlyWhenActive {
        super.closeKeep();
        unlockMemberStakes();
    }
```

solidity/contracts/BondedECDSAKeep.sol
```diff
@@ -527,127 +78,48 @@ contract BondedECDSAKeep is IBondedECDSAKeep {
...
-        status = Status.Active;
-        isInitialized = true;
 
         tokenStaking.claimDelegatedAuthority(_keepFactory);
 
         for (uint256 i = 0; i < _members.length; i++) {
             tokenStaking.lockStake(_members[i], _stakeLockDuration);
         }
```
pdyraga, line 96: Can we extract this code to `lockMemberStakes` function and place it next to `unlockMemberStakes`?

solidity/contracts/BondedECDSAKeep.sol
```diff
@@ -14,131 +14,26 @@
...
 
 /// @title Bonded ECDSA Keep
 /// @notice ECDSA keep with additional signer bond requirement.
 /// @dev This contract is used as a master contract for clone factory in
 /// BondedECDSAKeepFactory as per EIP-1167. It should never be removed after
 /// initial deployment as this will break functionality for all created clones.
-contract BondedECDSAKeep is IBondedECDSAKeep {
+contract BondedECDSAKeep is AbstractBondedECDSAKeep {
     using AddressArrayUtils for address[];
     using SafeMath for uint256;
```
pdyraga, line 31: Do we still need them?

solidity/contracts/api/IBondingManagement.sol
nkuba: There are two types of issues when defining it as interface: `Interfaces cannot inherit` and `Functions in interfaces must be declared external`.

pdyraga: _Not great, not terrible_

 Comrade Dyatlov

solidity/contracts/AbstractBondedECDSAKeep.sol
nkuba: I named it this way as `KeepBonding` contract is specific for current bonding with token staking. For ETH only bonding we won't use `KeepBonding` contract, hence the name `_keepBonding` won't be accurate in abstract contract. In `BondedECDSAKeep` contract we are using `KeepBonding` contract, so the parameter name there is `_keepBonding`.

If you want to have both aligned I can agree on using `_bonding`.

solidity/contracts/BondedECDSAKeep.sol
nkuba: Yes, I was thinking about it but wasn't 100% sure about removing. Glad you mentioned it. I'll do that.

solidity/contracts/AbstractBondedECDSAKeep.sol
pdyraga: We name the contract `KeepBonding` because it's Keep-org bonding contract, not because we are bonding KEEP there. I am fine with just `_bonding`.

solidity/contracts/BondedECDSAKeep.sol
nkuba: Nope

solidity/contracts/api/IBondingManagement.sol
nkuba: > _Not great, not terrible_
> 
> Comrade Dyatlov

_It's not bad... It's terrible!_

My Uncle

pdyraga: 😆

solidity/contracts/AbstractBondedECDSAKeep.sol
nkuba: Then we should rename `KeepBonding` contract as it's strictly related to KEEP by token ownership, grants and `TokenStaking`. For ETH only staking we will still be in Keep-org land but `KeepBonding` couldn't be used. A better name for current `KeepBonding` would be `TokenBonding` (because of the relation to `TokenStaking` contract).

pdyraga: Ah... I see what you mean. `KeepBonding` is no longer just about organization prefix but about the fact we refer there to `TokenStaking` and `TokenGrant` and we'll not refer to them for ETH-only version. Cool, I think the original naming convention makes a lot of sense then.

solidity/contracts/BondedECDSAKeep.sol
```diff
@@ -147,361 +37,15 @@ contract BondedECDSAKeep is IBondedECDSAKeep {
...
-
-        emit SignatureSubmitted(digest, _r, _s, _recoveryID);
-    }
 
     /// @notice Closes keep when owner decides that they no longer need it.
     /// Releases bonds to the keep members. Keep can be closed only when
     /// there is no signing in progress or requested signing process has timed out.
     /// @dev The function can be called only by the owner of the keep and only
     /// if the keep has not been already closed.
-    function closeKeep() external onlyOwner onlyWhenActive {
```
pdyraga, line 291: One last comment: can we reorder functions in `BondedECDSAKeep` into:
- `initialize`
- `closeKeep`
- `terminateKeep`
- `slashForSignatureFraud`

(... the rest stay as-is ...)

Issue #524, opened 2020-08-10
Rate-limit of ETH client calls


lukasz-zimnoch:
In case of a significant number of registered keeps (e.g. ~1000), the `keep-ecdsa` client performs a huge number of concurrent calls to the Ethereum client on startup. Those calls are specifically related to awaiting signatures and keep's activity checks. If the external Ethereum client has a rate limit (like Alchemy), it can be easily drained so all requests sent by the `keep-ecdsa` client will fail. Apart from that, there is a high risk the external Ethereum client will drop the connection so the `keep-ecdsa` client will perform reconnection attempts which will worse the situation and increase the overall number of requests. 

To avoid all the problems described above, we should implement a rate-limiting mechanism that will control the number of requests sent to the external Ethereum client.

PR #531, opened 2020-08-17
Enable the Ethereum rate limiter

lukasz-zimnoch:
Closes: #524 

Here we add the possibility of enabling the Ethereum rate limiter introduced by https://github.com/keep-network/keep-common/pull/46

Rate-limiter wraps Ethereum contract backend. It can limit the average number of requests per second using the token bucket algorithm and the request concurrency using a semaphore.


go.mod
```diff
@@ -18,7 +18,7 @@ require (
 	github.com/gogo/protobuf v1.3.1
 	github.com/google/gofuzz v1.1.0
 	github.com/ipfs/go-log v1.0.4
-	github.com/keep-network/keep-common v1.1.1-0.20200703125023-d9872a19ebd1
+	github.com/keep-network/keep-common v1.1.1-0.20200817190406-bcde8d4ccaa4
```
pdyraga, line 21: Let's bump it up to `09eff23`.

configs/config.toml.SAMPLE
```diff
@@ -19,6 +19,19 @@
...
+  # Uncomment to enable Ethereum node rate limiting. Both properties can be
+  # used together or separately.
+  	#
+	# RequestsPerSecondLimit sets the maximum average number of requests
+    	# per second which can be executed against the Ethereum node.
+    	#
+    	# RequestsPerSecondLimit = 150
+    	#
+    	# ConcurrencyLimit sets the maximum number of concurrent requests which
+    	# can be executed against the Ethereum node in the same time.
```
pdyraga, line 32: `s/in the same time/at the same time`

pdyraga: Also, let's be very explicit that all types of requests are affected:

```
# This limit affects all types of Ethereum node requests, including view function calls.
```

configs/config.toml.SAMPLE
```diff
@@ -19,6 +19,19 @@
 	# performed.
 	#
 	# MaxGasPrice = 70000000000 # 70 gwei (default value)
+	#
+  # Uncomment to enable Ethereum node rate limiting. Both properties can be
+  # used together or separately.
+  	#
+	# RequestsPerSecondLimit sets the maximum average number of requests
+    	# per second which can be executed against the Ethereum node.
```
pdyraga, line 27: Let's be very explicit all types of requests are affected:
```
# All types of Ethereum node requests are rate-limited, including view function calls.
```

pdyraga (approved):

pdyraga:
Tested with
```
RequestsPerSecondLimit = 10
ConcurrencyLimit = 10
```
as well as with
```
RequestsPerSecondLimit = 1
ConcurrencyLimit = 50
```

Works as advertised.

PR #532, opened 2020-08-20
Add GoSec workflow

lukasz-zimnoch:
Refs: https://github.com/keep-network/keep-core/issues/2007

Here we introduce the [GoSec](https://github.com/securego/gosec) scan workflow to the Github Actions pipeline and deal with all discovered problems.

Summary of `#nosec` exceptions and their motivation:

1. https://github.com/keep-network/keep-ecdsa/blob/a7348714c621f050ad53b98cbb5967b447ded3c7/internal/config/config.go#L18
Rule G101: Look for hardcoded credentials
Motivation: This line doesn't contain any credentials. It's just the name of the environment variable.

2. https://github.com/keep-network/keep-ecdsa/blob/a7348714c621f050ad53b98cbb5967b447ded3c7/internal/testdata/tss.go#L32
Rule G304: File path provided as taint input
Motivation: We use this line to read a test fixture file. There is no user input.

3. https://github.com/keep-network/keep-ecdsa/blob/a7348714c621f050ad53b98cbb5967b447ded3c7/pkg/chain/local/local.go#L323
Rule G404: Insecure random number source (rand)
Motivation: Local chain implementation doesn't require secure randomness.


pkg/ecdsa/tss/message.go
```diff
@@ -40,13 +40,33 @@ func (m *AnnounceMessage) Type() string {
 }
 
 func RegisterUnmarshalers(broadcastChannel net.BroadcastChannel) {
-	broadcastChannel.RegisterUnmarshaler(func() net.TaggedUnmarshaler {
+	err := broadcastChannel.RegisterUnmarshaler(func() net.TaggedUnmarshaler {
```
pdyraga, line 43: Having the recent `keep-core` version referenced after merging https://github.com/keep-network/keep-ecdsa/pull/535 we should update all those calls.

pkg/client/client.go
```diff
@@ -173,7 +173,7 @@ func Initialize(
 	)
 
 	// Watch for new keeps creation.
-	ethereumChain.OnBondedECDSAKeepCreated(func(event *eth.BondedECDSAKeepCreatedEvent) {
+	_, err := ethereumChain.OnBondedECDSAKeepCreated(func(event *eth.BondedECDSAKeepCreatedEvent) {
```
pdyraga, line 176: Can we deal with the error returned from `OnBondedECDSAKeepCreated` the same way as we did it in `keep-core`?  Explanation: https://github.com/keep-network/keep-core/pull/1996#issuecomment-679062032

pdyraga:
Something is wrong with the generated code, I assume it's about the fact we checked in some code:
```
Step 22/26 : RUN GOOS=linux GOPRIVATE=$GOPRIVATE go build -a -o $APP_NAME ./ && 	mv $APP_NAME $BIN_PATH
 ---> Running in cca52a91b115
# github.com/keep-network/keep-ecdsa/pkg/chain/gen/contract
pkg/chain/gen/contract/BondedECDSAKeepFactory.go:1556:34: becdsakf.contract.MasterBondedECDSAKeepAddress undefined (type *"github.com/keep-network/keep-ecdsa/pkg/chain/gen/abi".BondedECDSAKeepFactory has no field or method MasterBondedECDSAKeepAddress)
The command '/bin/sh -c GOOS=linux GOPRIVATE=$GOPRIVATE go build -a -o $APP_NAME ./ && 	mv $APP_NAME $BIN_PATH' returned a non-zero code: 2
```

pdyraga (approved):

PR #533, opened 2020-08-20
Crytic setup and basic improvements

pdyraga:
Here we configure Crytic as well as doing some small fixes and improvements to make Slither happier.

The few issues left Slither is not happy about are false-positives in my opinion, should be triaged, and probably ignored.

Unit tests are still not working inside Crytic because a bug in Crytic - this was reported and confirmed by their team.


solidity/slither.config.json
```diff
@@ -0,0 +1,4 @@
+{
+    "exclude_dependencies": true,
+    "filter_paths": "test"
```
Shadowfiend, line 3: This should be `contracts/test`, no? Seeing we have a test fix as well hehe.

pdyraga: Yeah, `contracts/test` is better. I started fixing stubs, then realized I can exclude them and because the history was a mess I started rebasing things. So it's not clear from the commit history, but the order of things was different and I added `filter_paths` at the end 😆

solidity/slither.config.json
```diff
@@ -0,0 +1,4 @@
+{
+    "exclude_dependencies": true,
```
Shadowfiend, line 2: Do we want to exclude dependencies? I would almost rather ignore dependency issues so that if a dependency introduces something, we can look at it and make sure we don't have to worry about it.

nkuba (approved):

PR #541, opened 2020-09-02
Sortition pool updates for stateful sets

pdyraga:
Two changes:
- Updated `SANCTIONED_APPLICATIONS` value to `1.1.0-rc.1` `TBTCSystem` address.
- Added `try-catch` around `getSortitionPool` in provisioning script: `getSortitionPool` reverts if pool does not exist for the given application. A similar `try-catch` solution is applied in `lcl-initialize.js`.

This state is already deployed to `keep-ecdsa-2` and `keep-ecdsa-3`.

### DIFF
--- a/infrastructure/kube/templates/keep-ecdsa/initcontainer/provision-keep-ecdsa/provision-keep-ecdsa.js
+++ b/infrastructure/kube/templates/keep-ecdsa/initcontainer/provision-keep-ecdsa/provision-keep-ecdsa.js
@@ -220,7 +220,11 @@ async function createSortitionPool(applicationAddress) {
     return await bondedECDSAKeepFactory.methods.getSortitionPool(applicationAddress).call()
   }
 
-  sortitionPoolContractAddress = await bondedECDSAKeepFactory.methods.getSortitionPool(applicationAddress).call()
+  try {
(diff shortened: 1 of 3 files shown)


lukasz-zimnoch (approved):

infrastructure/kube/templates/keep-ecdsa/initcontainer/provision-keep-ecdsa/provision-keep-ecdsa.js
```diff
@@ -220,7 +220,11 @@ async function createSortitionPool(applicationAddress) {
     return await bondedECDSAKeepFactory.methods.getSortitionPool(applicationAddress).call()
   }
 
-  sortitionPoolContractAddress = await bondedECDSAKeepFactory.methods.getSortitionPool(applicationAddress).call()
+  try {
+    sortitionPoolContractAddress = await bondedECDSAKeepFactory.methods.getSortitionPool(applicationAddress).call()
+  } catch (err) {
+    console.error("failed to get sortition pool", err)
```
Shadowfiend, line 226: Small note: We don’t want this to continue. Sortition pool creation is the responsibility of the tbtc deploy now, not the initcontainer (this is very recent).

PR #549, opened 2020-09-17
Information on the gas cost of operating the client

pdyraga:
Keep ECDSA smart contracts do not reimburse the operator for submitted transactions. It is important that the operators understand the cost of running the client and can estimate the required ETH based on their stake, bond, current gas price, and the expected traffic in the network.


Shadowfiend (approved):
Looks good!

Issue #554, opened 2020-09-23
Signer generation for keep sometimes continues although the public key is submitted


pdyraga:
One of our stakers reported the following warning in logs (time in UTC):

> Sep 23, 2020 @ 07:16:12.824 signer generation for keep [0x9D200A3Ff21F0A1e383e5D78B6b9BC1c6e59b82a]; attempt [90]

That was eventually followed by:

> Sep 23, 2020 @ 07:16:12.952failed to generate signer for keep [0x9D200A3Ff21F0A1e383e5D78B6b9BC1c6e59b82a]: [key generation timeout exceeded]

I inspected it at 07:16 UTC and I see that all member nodes of that keep are up and operating.

What's more, the keep already has the public key generated and published:

> keep address: 0x9D200A3Ff21F0A1e383e5D78B6b9BC1c6e59b82a
> keep index:   96
> pubkey:       0x4ff83f5e701c10c94b6f8f8d93ae5f8bd01eece8dec0cdcab5010f139126754fd88d99f949bea392e2105735aabf22234a71f175eca47f16cf97f1838b4d39ce
> members: 0xA79827813Da962f5A4C95AF40550259EE868d202,0x5CA1F949c75833432d6BC8E3cb6FB23386F63426,0x2eBE08379f4fD866E871A9b9E1d5C695154C6A9F
> isActive:     true
> bond [wei]:   45929146636350000000
> bond [eth]:   45.92914663635

What's even more interesting, that public key was submitted 5 hours earlier by all members around Sep-23-2020 02:16:31 AM UTC, see https://etherscan.io/address/0x9D200A3Ff21F0A1e383e5D78B6b9BC1c6e59b82a.

It seems that in certain circumstances, some nodes may attempt to continue key generation although the key is already there. I confirmed that at least 2 out of 3 nodes still try to generate the key and they fail on the announce protocol, most probably because the 3rd member does correctly recognize the fact the key has been published.

I was able to confirm the key generation for that keep went pretty smooth and nothing weird was observed during that.

pdyraga:
Confirmed at least two out of three nodes have the key share stored locally.

pdyraga:
Logs from the moment of generating keep's public key:

```
Sep 23, 2020 @ 04:18:13.090signer generation for keep [0x9D200A3Ff21F0A1e383e5D78B6b9BC1c6e59b82a]; attempt [2]
[...]

PR #556, opened 2020-09-23
Check signer existence before key generation

lukasz-zimnoch:
Closes: #554 

Added a condition which checks signer existence before triggering a key generation process in response to `BondedECDSAKeepCreatedEvent`.

The problem addressed here concerns duplicated events emerging just after the end of the first key generation process. In that case, the `requestedSigners` cache does no longer contain the original event reference and allows triggering a new key generation in response to the duplicated event. Checking signer existence at this stage will protect against redundant key generation processes which may lead to an unpredictable behavior of the clients.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -191,6 +191,14 @@ func Initialize(
 				}
 				defer requestedSigners.remove(event.KeepAddress)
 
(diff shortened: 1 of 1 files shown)


pdyraga (approved):
Tested locally with a smoke test couple of times.

PR #562, opened 2020-09-26
Event already registered log levels changed from ERROR to WARNING

### LINKED ISSUE #552: signature requested for event [...] and digest [...] already registered
Reported by one of our mainnet stakers, this is happening in client logs after signature has been requested:

image

It seems it could be specific to Ethereum deployment or, we could have a small chain re-org given that two blocks were reported: 10906703 and 10906702. 

Should this think be an ERROR or just a WARNING?

### LINKED ISSUE #560: "keep creation event for keep [...] already registered" should not be logged as an error
`keep creation event for keep [...] already registered` is not an error and it might happen in certain Ethereum client deployments (noded behind a load-balancer). Let's log it as a warning.

pdyraga:
Closes #560 
Closes #552

Event duplicates are happening quite often on mainnet given minor chain
reorgs. Logging event duplicates on ERROR level is too noisy for
operators monitoring their nodes. We are changing event duplicate log
levels for keep creation and signature request from ERROR to WARNING.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -183,7 +183,7 @@ func Initialize(
 		if event.IsMember(ethereumChain.Address()) {
 			go func(event *eth.BondedECDSAKeepCreatedEvent) {
 				if ok := requestedSigners.add(event.KeepAddress); !ok {
-					logger.Errorf(
+					logger.Warningf(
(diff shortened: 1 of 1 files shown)


dimpar (approved):
lgtm!

PR #563, opened 2020-09-26
Sortition pool registration notifications

### LINKED ISSUE #555: Out of bonds client logging improvements
Two suggestions about sortition pool event logging.

When operator is no longer eligible to be in the sortition pool, the client logs an ERROR:

> "failed on signer pool status monitoring: [failed to check operator status for application [0xe20A5C79b39bC8C363f0f49ADcFa82C2a01ab64a]: [got error [execution reverted: Operator not registered for the application] while resolving original error [execution reverted: Operator not registered for the application]]]"

This is fine - operator should be alerted they are removed from the pool. The problem I see with this message is that it's not that clear what's happening. We could improve it a little.

The second suggestion is that we are logging every few seconds when the operator is no longer eligible to be in the pool:
> "operator is not eligible for application [0xe20A5C79b39bC8C363f0f49ADcFa82C2a01ab64a]"

This is pretty aggresive. Some operators could reach the ceiling of their bonds and now they are in the network just to do their signing duty and they are no longer interested in being in the pool until some of their bonds are freed.

pdyraga:
Depends on https://github.com/keep-network/keep-ecdsa/pull/562
Closes  #555

Two improvements included here:

- Increased sortition pool join eligibility check delay to 20 minutes 

  Sortition pool join eligibility check was using the same retry delay as
every other retried operation in the registration logic - 1 sec. This is
pretty aggresive for this particular check as it may happen that the
client is no longer eligible to join the pool because its bonds have
been drained to zero. Here we increase the delay for this check to 20
minutes and in the future we should consider enhancing the logic in  way
allowing the operator to specify that it does no longer want to be in
the sortition pool but it still wants to do their signing duty.

- Refined error message for failed sortition pool status monitoring

  In 99% cases this error is logged when the sortition pool status
monitoring failed because the operator was removed from the sortition
pool because it does no longer have enough unbonded value or stake.

  We do not change the logic in any way here but just improve the error
message to at least give the operator some context on what they should
be looking at.

### DIFF
--- a/pkg/client/registration.go
+++ b/pkg/client/registration.go
@@ -61,7 +61,11 @@ RegistrationLoop:
 			// once the registration is confirmed or if the client is already
 			// registered, we can start to monitor the status
 			if err := monitorSignerPoolStatus(ctx, ethereumChain, application); err != nil {
-				logger.Errorf("failed on signer pool status monitoring: [%v]", err)
+				logger.Errorf(
+					"failed on signer pool status monitoring; please inspect "+
+						"signer's unbonded value and stake: [%v]",
+					err,
+				)
 				time.Sleep(retryDelay) // TODO: #413 Replace with backoff.
 				continue RegistrationLoop
 			}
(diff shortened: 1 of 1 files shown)


dimpar (approved):
looks good

PR #564, opened 2020-09-28
Improvements of the storage layer

lukasz-zimnoch:
Depends on: https://github.com/keep-network/keep-common/pull/49

Here we introduce some improvements to the storage layer:

- Removed the many-to-one relation between signers and keeps in the keep registry. Instead of that, we introduce a one-to-one relationship. The motivation of this change is the fact that one operator can have only one signer for given keep so we can make the registry simpler
- Introduced signer snapshots just after the signer is generated but before the public key is submitted on-chain. This feature allows us to back up the generated key material on disk and recover it in case something goes wrong within a tight time window between public key submission and registration of the signer in the keep registry. 
- Added a return from `generateKeyForKeep` method execution in case the `RegisterSigner` method invocation returns an error. There is no sense to continue and subscribe to keep's events in case the signer hasn't been registered correctly.
- Made a small cleanup of unused things

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -333,8 +333,6 @@ github.com/keep-network/go-libp2p-bootstrap v0.0.0-20200423153828-ed815bc50aec h
 github.com/keep-network/go-libp2p-bootstrap v0.0.0-20200423153828-ed815bc50aec/go.mod h1:xR8jf3/VJAjh3nWu5tFe8Yxnt2HvWsqZHfGef1P5oDk=
 github.com/keep-network/keep-common v1.2.0 h1:hVd2tTd7vL+9CQP5Ntk5kjs+GYvkgrRNBcNvTuhHhVk=
 github.com/keep-network/keep-common v1.2.0/go.mod h1:emxogTbBdey7M3jOzfxZOdfn139kN2mI2b2wA6AHKKo=
-github.com/keep-network/keep-common v1.2.1-0.20200929121054-30cac806a275 h1:UWZ4nQF20LzzLdc1tLzKBa2idd0d8+UsKZApHCEg/Ns=
-github.com/keep-network/keep-common v1.2.1-0.20200929121054-30cac806a275/go.mod h1:emxogTbBdey7M3jOzfxZOdfn139kN2mI2b2wA6AHKKo=
 github.com/keep-network/keep-common v1.2.1-0.20201002105641-e04cc579ff66 h1:x/9fra2wLrBDhS8LOQztFajoM2q13FgBN5jFiqBGBog=
 github.com/keep-network/keep-common v1.2.1-0.20201002105641-e04cc579ff66/go.mod h1:emxogTbBdey7M3jOzfxZOdfn139kN2mI2b2wA6AHKKo=
 github.com/keep-network/keep-core v1.3.0 h1:7Tb33EmO/ntHOEbOiYciRlBhqu5Ln6KemWCaYK0Z6LA=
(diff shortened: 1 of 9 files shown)


pkg/registry/keeps_test.go
```diff
@@ -67,6 +70,67 @@ func TestRegisterSigner(t *testing.T) {
...
+	err = kr.RegisterSigner(keepAddress1, signer2)
+
+	expectedError := fmt.Errorf("signer for keep [%s] already registered", keepAddress1.String())
+	if err == nil || (!reflect.DeepEqual(expectedError, err)) {
+		t.Errorf(
+			"unexpected error\nexpected: [%v]\nactual:   [%v]",
+			expectedError,
+			err,
+		)
+	}
```
pdyraga, line 88: Can we make it a separate test case, e.g. `TestRegisterSignerDuplicate`? We don't need to do any assertions for the first signer there, just confirm we get this error.

pkg/registry/keeps_test.go
```diff
@@ -67,6 +70,67 @@ func TestRegisterSigner(t *testing.T) {
...
+
+	signer2, err := newTestSigner(1)
+	if err != nil {
+		t.Fatalf("failed to get signer: [%v]", err)
+	}
+
+	err = kr.RegisterSigner(keepAddress1, signer2)
+
+	expectedError := fmt.Errorf("signer for keep [%s] already registered", keepAddress1.String())
+	if err == nil || (!reflect.DeepEqual(expectedError, err)) {
```
pdyraga, line 82: Do we need `err == nil` here given that `exectedError` is not `nil`?

go.sum
```diff
@@ -333,6 +333,8 @@ github.com/keep-network/go-libp2p-bootstrap v0.0.0-20200423153828-ed815bc50aec h
 github.com/keep-network/go-libp2p-bootstrap v0.0.0-20200423153828-ed815bc50aec/go.mod h1:xR8jf3/VJAjh3nWu5tFe8Yxnt2HvWsqZHfGef1P5oDk=
 github.com/keep-network/keep-common v1.2.0 h1:hVd2tTd7vL+9CQP5Ntk5kjs+GYvkgrRNBcNvTuhHhVk=
 github.com/keep-network/keep-common v1.2.0/go.mod h1:emxogTbBdey7M3jOzfxZOdfn139kN2mI2b2wA6AHKKo=
+github.com/keep-network/keep-common v1.2.1-0.20200929121054-30cac806a275 h1:UWZ4nQF20LzzLdc1tLzKBa2idd0d8+UsKZApHCEg/Ns=
+github.com/keep-network/keep-common v1.2.1-0.20200929121054-30cac806a275/go.mod h1:emxogTbBdey7M3jOzfxZOdfn139kN2mI2b2wA6AHKKo=
```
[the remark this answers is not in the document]
lukasz-zimnoch: `go mod tidy` doesn't change anything here.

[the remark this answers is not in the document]
pdyraga: Yeah, `keep-core` is still on the old version. OK, let's leave it as-is for now.

pdyraga:
It looks good to me, before testing, I'd like to merge https://github.com/keep-network/keep-common/pull/50 and update the referenced `keep-common` version.

go.sum
```diff
@@ -335,6 +335,8 @@ github.com/keep-network/keep-common v1.2.0 h1:hVd2tTd7vL+9CQP5Ntk5kjs+GYvkgrRNBc
 github.com/keep-network/keep-common v1.2.0/go.mod h1:emxogTbBdey7M3jOzfxZOdfn139kN2mI2b2wA6AHKKo=
 github.com/keep-network/keep-common v1.2.1-0.20200929121054-30cac806a275 h1:UWZ4nQF20LzzLdc1tLzKBa2idd0d8+UsKZApHCEg/Ns=
 github.com/keep-network/keep-common v1.2.1-0.20200929121054-30cac806a275/go.mod h1:emxogTbBdey7M3jOzfxZOdfn139kN2mI2b2wA6AHKKo=
+github.com/keep-network/keep-common v1.2.1-0.20201002105641-e04cc579ff66 h1:x/9fra2wLrBDhS8LOQztFajoM2q13FgBN5jFiqBGBog=
+github.com/keep-network/keep-common v1.2.1-0.20201002105641-e04cc579ff66/go.mod h1:emxogTbBdey7M3jOzfxZOdfn139kN2mI2b2wA6AHKKo=
```
pdyraga, line 339: I think now `go mod tidy` should eliminate the previous `v1.2.1` reference.

pdyraga:
Things I have tested:
- Existing storage with `current` and `archive` directories. After launching the new binary `snapshot` directory is added and the existing ones are unaffected. All active keeps are loaded.
- I was able to sign with existing keeps (`current`) directory.
- I was able to open new keeps, I confirmed `snapshot` is written before `current`.
- I was able to use `snapshot` key shares to produce a signature (copied them to `current` previously deleting original ones from there).

pdyraga (approved):

PR #566, opened 2020-09-30
Signing CLI 

lukasz-zimnoch:
This pull request introduces an out-of-band signing CLI.

That signing CLI contains the following commands:
- `decrypt-key-share`: allows to decrypt a key share for given keep and store it in a file using the client config toml
- `sign-digest`: allows signing a given digest using provided key shares

### DIFF
--- a/cmd/signing.go
+++ b/cmd/signing.go
@@ -1,0 +1,125 @@
+package cmd
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/keep-network/keep-common/pkg/chain/ethereum/ethutil"
+	"github.com/keep-network/keep-common/pkg/persistence"
+	"github.com/keep-network/keep-ecdsa/internal/config"
+	"github.com/keep-network/keep-ecdsa/pkg/registry"
+	"github.com/urfave/cli"
+)
+
+var SigningCommand cli.Command
+
+func init() {
+	SigningCommand = cli.Command{
+		Name:  "signing",
+		Usage: "Provides several tools useful for out-of-band signing",
+		Subcommands: []cli.Command{
+			{
+				Name: "decrypt-key-share",
+				Usage: "Decrypts the key share of the operator for the given " +
+					"keep and stores it in a file",
+				Action: DecryptKeyShare,
+				Flags: []cli.Flag{
+					cli.StringFlag{
+						Name:  "keep-address,k",
+						Usage: "address of the keep",
+					},
+				},
+			},
+		},
+	}
+}
+
+func DecryptKeyShare(c *cli.Context) error {
+	config, err := config.ReadConfig(c.GlobalString("config"))
+	if err != nil {
+		return fmt.Errorf("failed while reading config file: [%v]", err)
+	}
+
+	keyFile, err := ethutil.DecryptKeyFile(
+		config.Ethereum.Account.KeyFile,
+		config.Ethereum.Account.KeyFilePassword,
+	)
+	if err != nil {
+		return fmt.Errorf("failed to decrypt key file: [%v]", err)
+	}
+
+	keepAddressHex := c.String("keep-address")
+	if !common.IsHexAddress(keepAddressHex) {
+		return fmt.Errorf("invalid keep address")
+	}
+
+	keepAddress := common.HexToAddress(keepAddressHex)
+
+	handle, err := persistence.NewDiskHandle(config.Storage.DataDir)
+	if err != nil {
+		return fmt.Errorf(
+			"failed while creating a storage disk handler: [%v]",
+			err,
+		)
+	}
+
+	persistence := persistence.NewEncryptedPersistence(
+		handle,
+		config.Ethereum.Account.KeyFilePassword,
+	)
+
+	keepRegistry := registry.NewKeepsRegistry(persistence)
+
+	keepRegistry.LoadExistingKeeps()
+
+	signers, err := keepRegistry.GetSigners(keepAddress)
+	if err != nil {
+		return fmt.Errorf(
+			"no signers for keep [%s]: [%v]",
+			keepAddress.String(),
+			err,
+		)
+	}
+
+	signer := signers[0]
+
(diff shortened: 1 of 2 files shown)


cmd/signing.go
```diff
@@ -0,0 +1,280 @@
...
+	waitGroup.Wait()
+	close(signingOutcomesChannel)
+
+	for signingOutcome := range signingOutcomesChannel {
+		if signingOutcome.err != nil {
+			logger.Errorf(
+				"[signer:%v] error: [%v]",
+				signingOutcome.signerIndex,
+				signingOutcome.err,
+			)
```
Shadowfiend, line 266: As above, this feels like something we should do as not-a-log, printing to stderr with something solely for human consumption.

cmd/signing.go
```diff
@@ -0,0 +1,280 @@
...
+			continue
+		}
+
+		logger.Infof(
+			"[signer:%v] signature: [%+v]",
```
Shadowfiend, line 271: Is the outcome here a single signature? I don't want to slow us down too much, but I would expect this to:
- Gather all available signatures.
- If there are <`len(signers)`, print an error to stderr and `exit 1`.
- If there are `len(signers)` signatures and they are not all the same, print an error to stderr (possibly listing the three signatures by signer id) and `exit 1`.
- If there are `len(signers)` signatures and they are all the same, print the signature to `stdout` and exit normally (which is to say, return).

I think the error to stderr + exit 1 is doable by just `return`ing an `error`.

Shadowfiend (changes requested):
Left some thoughts on how to make this a little more “command”-y.

lukasz-zimnoch:
Thanks for review @Shadowfiend! Ready for the next look.

cmd/signing.go
```diff
@@ -31,31 +32,29 @@ func init() {
 	SigningCommand = cli.Command{
 		Name:  "signing",
 		Usage: "Provides several tools useful for out-of-band signing",
+		Before: func(c *cli.Context) error {
+			_ = log.SetLogLevel("*", "fatal") // disable the regular logger
```
Shadowfiend, line 36: Thoughts on going through our existing `keep-common` logging abstraction here and doing a `logging.Configure("*", "fatal")`? My thinking is it'll be easier to bridge this if and when we bump our logging dependencies to a different major version or so (e.g. with keep-network/keep-common#30).

cmd/signing.go
```diff
@@ -119,44 +110,45 @@ func DecryptKeyShare(c *cli.Context) error {
...
-
-	if _, err := os.Stat(targetFilePath); !os.IsNotExist(err) {
-		return fmt.Errorf(
-			"could not write shares to file; file [%s] already exists",
-			targetFilePath,
-		)
-	}
-
-	err = ioutil.WriteFile(targetFilePath, signerBytes, 0444) // read-only
+	_, err = os.Stdout.Write(signerBytes)
```
Shadowfiend, line 113: If we do support `-o`, we shouldn't write to stdout if the output file is specified.

cmd/signing.go
```diff
@@ -31,31 +32,29 @@ func init() {
 	SigningCommand = cli.Command{
 		Name:  "signing",
 		Usage: "Provides several tools useful for out-of-band signing",
+		Before: func(c *cli.Context) error {
```
Shadowfiend, line 35: We can also use `Before` here and below to check the `Args` length and make sure it matches what we expect, returning an `error` if it does not.

cmd/signing.go
```diff
@@ -252,24 +244,38 @@ func SignDigest(c *cli.Context) error {
...
+			"signing failed; a single signature should be produced",
 		)
 	}
 
-	logger.Infof("signing completed")
+	for signature, signersCount := range signatures {
+		if signersCount != len(signers) {
+			return fmt.Errorf(
+				"signing failed; all signers should support the signature",
+			)
```
Shadowfiend, line 274: Turns out I was _totally wrong_ about returning an error. It does do us the favor of exiting with a non-zero status, but it prints the error out to a logger. This… Feels like a mistake, but probably not one we need to block this PR on. In particular, it feels like `main.go` should be `fmt.Printf`-ing an error from `app.Run` rather than using a logger.

Shadowfiend (approved):
I think this is ready to try on a real keep :sunglasses:

cmd/signing.go
```diff
@@ -0,0 +1,281 @@
...
+	}
+
+	persistence := persistence.NewEncryptedPersistence(
+		handle,
+		config.Ethereum.Account.KeyFilePassword,
+	)
+
+	keepRegistry := registry.NewKeepsRegistry(persistence)
+
+	keepRegistry.LoadExistingKeeps()
```
Shadowfiend, line 91: Actually a random note I missed making: we will need to make this work more flexibly. Right now you need a full config file and to have your path set up just right to decrypt the keyshare. We should be able to just point this to a keyshare and, if we have the password configured properly, have that keyshare decrypted to stdout. Otherwise if we have a snapshot that didn't get properly persisted, for example, you'd have to move it into the `current` directory to make this work.

For followup work though.

cmd/signing.go
lukasz-zimnoch: Addressed in the follow-up PR:  #568

cmd/signing.go
lukasz-zimnoch: Addressed in the follow-up PR:  #568

cmd/signing.go
lukasz-zimnoch: Addressed in the follow-up PR:  #568

cmd/signing.go
lukasz-zimnoch: Yep, but I feel it's a little bit clearer if we validate the arguments in the subcommand right in the place where we use them.

cmd/signing.go
lukasz-zimnoch: Agree but from the other side, seems we won't be able to use the existing registry code as we do here and we will be forced to copy-paste relevant parts.

Shadowfiend: Yeah I figured structuring it all right will be a bit more complicated.

PR #567, opened 2020-10-02
Backer ECDSA rewards

dimpar:
This PR is a continuation of KEEP Rewards work done in https://github.com/keep-network/keep-ecdsa/pull/513/files and https://github.com/keep-network/keep-core/pull/1863. 

Here we add Rewards distribution for `keep-ecdsa` for the period of time between May13 2020 - Sep14 2020.

`ECDSABackportRewards` will distribute KEEP between involved signers. We count keeps starting from [bonded ECDSA Keep Factory Deployment](https://etherscan.io/address/0x18758f16988E61Cd4B61E6B930694BD9fB07C22F) on May 13 until the new version of this contract that was deployed on [Sep 14](https://etherscan.io/address/0xA7d9E842EFB252389d613dA88EDa3731512e40bD). There were 41 keeps created during that time.

Upon deployment of this contract, keep stakers can claim their rewards calling `receiveReward(keepAddress)`.

### DIFF
--- a/solidity/contracts/ECDSABackportRewards.sol
+++ b/solidity/contracts/ECDSABackportRewards.sol
@@ -23,7 +23,7 @@ import "./api/IBondedECDSAKeep.sol";
 /// the keeps which were created by the BondedECDSAKeepFactory contract:
 /// https://etherscan.io/address/0x18758f16988E61Cd4B61E6B930694BD9fB07C22F
 ///
-/// Keep signers from May release of BondedECDSAKeepFactory contract can claim 
+/// Keep signers from May release of BondedECDSAKeepFactory contract can claim
 /// their rewards at any time.
 contract ECDSABackportRewards is Rewards {
     // BondedECDSAKeepFactory deployment date, May-13-2020 interval started.
@@ -37,7 +37,7 @@ contract ECDSABackportRewards is Rewards {
     // versions of BondedECDSAKeepFactory.
     // Older: https://etherscan.io/address/0x18758f16988E61Cd4B61E6B930694BD9fB07C22F
     // Newer: https://etherscan.io/address/0xA7d9E842EFB252389d613dA88EDa3731512e40bD
-    // The actual value between these 2 contracts deployment time is a little less 
+    // The actual value between these 2 contracts deployment time is a little less
     // than 124 days.
     uint256 internal constant backportECDSATermLength = 124 days;
 
@@ -94,10 +94,7 @@ contract ECDSABackportRewards is Rewards {
         return factory.getKeepOpenedTimestamp(toAddress(_keep)) != 0;
     }
 
-    function _distributeReward(bytes32 _keep, uint256 amount)
-        internal
-        isAddress(_keep)
-    {
+    function _distributeReward(bytes32 _keep, uint256 amount) internal {
         token.approve(toAddress(_keep), amount);
 
         IBondedECDSAKeep(toAddress(_keep)).distributeERC20Reward(
@@ -106,20 +103,11 @@ contract ECDSABackportRewards is Rewards {
         );
     }
 
-    function validAddressBytes(bytes32 keepBytes) internal pure returns (bool) {
-        return fromAddress(toAddress(keepBytes)) == keepBytes;
-    }
-
     function toAddress(bytes32 keepBytes) internal pure returns (address) {
         return address(bytes20(keepBytes));
     }
 
     function fromAddress(address keepAddress) internal pure returns (bytes32) {
         return bytes32(bytes20(keepAddress));
-    }
-
-    modifier isAddress(bytes32 _keep) {
-        require(validAddressBytes(_keep), "Invalid keep address");
-        _;
     }
 }
--- a/solidity/contracts/ECDSABackportRewards.sol
+++ b/solidity/contracts/ECDSABackportRewards.sol
@@ -1,0 +1,61 @@
+/**
+▓▓▌ ▓▓ ▐▓▓ ▓▓▓▓▓▓▓▓▓▓▌▐▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓ ▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓ ▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▄
+▓▓▓▓▓▓▓▓▓▓ ▓▓▓▓▓▓▓▓▓▓▌▐▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓ ▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓ ▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓
+  ▓▓▓▓▓▓    ▓▓▓▓▓▓▓▀    ▐▓▓▓▓▓▓    ▐▓▓▓▓▓   ▓▓▓▓▓▓     ▓▓▓▓▓   ▐▓▓▓▓▓▌   ▐▓▓▓▓▓▓
+  ▓▓▓▓▓▓▄▄▓▓▓▓▓▓▓▀      ▐▓▓▓▓▓▓▄▄▄▄         ▓▓▓▓▓▓▄▄▄▄         ▐▓▓▓▓▓▌   ▐▓▓▓▓▓▓
+  ▓▓▓▓▓▓▓▓▓▓▓▓▓▀        ▐▓▓▓▓▓▓▓▓▓▓▌        ▓▓▓▓▓▓▓▓▓▓▌        ▐▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓
+  ▓▓▓▓▓▓▀▀▓▓▓▓▓▓▄       ▐▓▓▓▓▓▓▀▀▀▀         ▓▓▓▓▓▓▀▀▀▀         ▐▓▓▓▓▓▓▓▓▓▓▓▓▓▓▀
+  ▓▓▓▓▓▓   ▀▓▓▓▓▓▓▄     ▐▓▓▓▓▓▓     ▓▓▓▓▓   ▓▓▓▓▓▓     ▓▓▓▓▓   ▐▓▓▓▓▓▌
+▓▓▓▓▓▓▓▓▓▓ █▓▓▓▓▓▓▓▓▓ ▐▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓ ▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓  ▓▓▓▓▓▓▓▓▓▓
+▓▓▓▓▓▓▓▓▓▓ ▓▓▓▓▓▓▓▓▓▓ ▐▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓ ▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓  ▓▓▓▓▓▓▓▓▓▓
+
+                           Trust math, not hardware.
+*/
+
+pragma solidity 0.5.17;
+
+import "@keep-network/keep-core/contracts/Rewards.sol";
+import "./api/IBondedECDSAKeepFactory.sol";
+
+contract ECDSABackportRewards is Rewards {
+
+    // BondedECDSAKeepFactory deployment date, May-13-2020 interval started.
+    // https://etherscan.io/address/0x18758f16988E61Cd4B61E6B930694BD9fB07C22F
+    uint256 internal constant bondedECDSAKeepFactoryDeployment = 1589408351;
+
+    // We are going to have one interval, with a weight of 100%.
+    uint256[] internal backportECDSAIntervalWeight = [100];
+    uint256 internal constant lastInterval = 0;
+
+    // Interval is the difference in time of creation between older and newer 
+    // versions of BondedECDSAKeepFactory.
+    // Older: https://etherscan.io/address/0x18758f16988E61Cd4B61E6B930694BD9fB07C22F
+    // Newer: https://etherscan.io/address/0xA7d9E842EFB252389d613dA88EDa3731512e40bD
+    uint256 internal constant backportECDSATermLength = 123 days; // 10678946 sec
+
+    // There were 41 Keeps created by BondedECDSAKeepFactory : 0x18758f16988E61Cd4B61E6B930694BD9fB07C22F
+    // The last Keep was opened on May-17-2020
+    // https://etherscan.io/address/0x45A3cACA2F2a78A53607618651C86111c9720AA5
+    uint256 internal constant numberOfCreatedECDSAKeeps = 41;
+
+    // We allocate ecdsa backport rewards to all 41 keeps.
+    uint256 internal constant minimumECDSAKeepsPerInterval = numberOfCreatedECDSAKeeps;
+
+    IBondedECDSAKeepFactory factory;
+
+    constructor (
+        address _token,
+        address _factoryAddress,
+        address _operatorContract,
+        address _stakingContract
+    ) public Rewards(
+        _token,
+        bondedECDSAKeepFactoryDeployment,
+        backportECDSAIntervalWeight,
+        backportECDSATermLength,
+        minimumECDSAKeepsPerInterval
+    ) {
+        factory = IBondedECDSAKeepFactory(_factoryAddress);
+    }
+
+}
--- a/solidity/contracts/ECDSABackportRewards.sol
+++ b/solidity/contracts/ECDSABackportRewards.sol
@@ -19,7 +19,6 @@ import "./BondedECDSAKeepFactory.sol";
 import "./api/IBondedECDSAKeep.sol";
 
 contract ECDSABackportRewards is Rewards {
-
     // BondedECDSAKeepFactory deployment date, May-13-2020 interval started.
     // https://etherscan.io/address/0x18758f16988E61Cd4B61E6B930694BD9fB07C22F
     uint256 internal constant bondedECDSAKeepFactoryDeployment = 1589408351;
@@ -28,7 +27,7 @@ contract ECDSABackportRewards is Rewards {
     uint256[] internal backportECDSAIntervalWeight = [100];
     uint256 internal constant lastInterval = 0;
 
-    // Interval is the difference in time of creation between older and newer 
+    // Interval is the difference in time of creation between older and newer
     // versions of BondedECDSAKeepFactory.
     // Older: https://etherscan.io/address/0x18758f16988E61Cd4B61E6B930694BD9fB07C22F
     // Newer: https://etherscan.io/address/0xA7d9E842EFB252389d613dA88EDa3731512e40bD
@@ -40,20 +39,21 @@ contract ECDSABackportRewards is Rewards {
     uint256 internal constant numberOfCreatedECDSAKeeps = 41;
 
     // We allocate ecdsa backport rewards to all 41 keeps.
-    uint256 internal constant minimumECDSAKeepsPerInterval = numberOfCreatedECDSAKeeps;
+    uint256
+        internal constant minimumECDSAKeepsPerInterval = numberOfCreatedECDSAKeeps;
 
     BondedECDSAKeepFactory factory;
 
-    constructor (
-        address _token,
-        address payable _factoryAddress
-    ) public Rewards(
-        _token,
-        bondedECDSAKeepFactoryDeployment,
-        backportECDSAIntervalWeight,
-        backportECDSATermLength,
-        minimumECDSAKeepsPerInterval
-    ) {
+    constructor(address _token, address payable _factoryAddress)
+        public
+        Rewards(
+            _token,
+            bondedECDSAKeepFactoryDeployment,
+            backportECDSAIntervalWeight,
+            backportECDSATermLength,
+            minimumECDSAKeepsPerInterval
+        )
+    {
         factory = BondedECDSAKeepFactory(_factoryAddress);
     }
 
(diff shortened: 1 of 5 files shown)


solidity/contracts/ECDSABackportRewards.sol
```diff
@@ -0,0 +1,127 @@
...
+    uint256 internal constant backportECDSATermLength = 123 days; // 10678946 sec
+
+    // There were 41 Keeps created by BondedECDSAKeepFactory : 0x18758f16988E61Cd4B61E6B930694BD9fB07C22F
+    // The last Keep was opened on May-17-2020
+    // https://etherscan.io/address/0x45A3cACA2F2a78A53607618651C86111c9720AA5
+    uint256 internal constant numberOfCreatedECDSAKeeps = 41;
+
+    // We allocate ecdsa backport rewards to all 41 keeps.
+    uint256
+        internal constant minimumECDSAKeepsPerInterval = numberOfCreatedECDSAKeeps;
```
pdyraga, line 42: We can set it to `40`, especially that I *think* we should read the count from factory (comment above).

solidity/contracts/ECDSABackportRewards.sol
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    function _getCreationTime(bytes32) internal view returns (uint256) {
+        // Assign each keep to the starting timestamp of its interval.
+        return startOf(0);
+    }
+
+    function _isClosed(bytes32) internal view returns (bool) {
+        // All keeps created between May 13 2020 - Sep 14 2020 are considered closed.
+        return true;
```
pdyraga, line 75: This is a good decision but it took me a while to understand why you did it this way (we have 18 closed keeps and 23 active keeps). Maybe we can link to https://tbtc.network/news/2020-05-21-details-of-the-tbtc-deposit-pause-on-may-18-2020/ and explain closing some of those keeps is not easily achievable given the problem in the previous version of tBTC and do not want to block rewards for good stakers because of that bug.

solidity/contracts/ECDSABackportRewards.sol
```diff
@@ -0,0 +1,127 @@
...
+        factory = BondedECDSAKeepFactory(_factoryAddress);
+    }
+
+    function _getKeepCount() internal view returns (uint256) {
+        // Between May 13 2020 - Sep 14 2020 there were 41 keeps opened.
+        return numberOfCreatedECDSAKeeps;
+    }
+
+    function _getKeepAtIndex(uint256 i) internal view returns (bytes32) {
+        return bytes32(i);
```
pdyraga, line 65: Shouldn't it be:
```
 function _getKeepAtIndex(uint256 index) internal view returns (bytes32) {
        return fromAddress(factory.getKeepAtIndex(index));
 }
```
?

Does it work the current way given that `_recognizedByFactory` operates on address?

solidity/contracts/ECDSABackportRewards.sol
```diff
@@ -0,0 +1,127 @@
...
+    function _recognizedByFactory(bytes32 _keep) internal view returns (bool) {
+        uint256 keepCreatedTimestamp = factory.getKeepOpenedTimestamp(
+            toAddress(_keep)
+        );
+        if (keepCreatedTimestamp == 0) {
+            return false;
+        }
+        return
+            keepCreatedTimestamp <
+            bondedECDSAKeepFactoryDeployment.add(backportECDSATermLength);
```
pdyraga, line 96: Do we need it? Given that the interval has 100% weight, it has already ended, and all keeps are considered happily closed, we should be good without this check. I think it is enough to check whether the creation date is non-zero.

solidity/contracts/ECDSABackportRewards.sol
```diff
@@ -0,0 +1,127 @@
...
+    uint256 internal constant bondedECDSAKeepFactoryDeployment = 1589408351;
+
+    // We are going to have one interval, with a weight of 100%.
+    uint256[] internal backportECDSAIntervalWeight = [100];
+
+    // Interval is the difference in time of creation between older and newer
+    // versions of BondedECDSAKeepFactory.
+    // Older: https://etherscan.io/address/0x18758f16988E61Cd4B61E6B930694BD9fB07C22F
+    // Newer: https://etherscan.io/address/0xA7d9E842EFB252389d613dA88EDa3731512e40bD
+    uint256 internal constant backportECDSATermLength = 123 days; // 10678946 sec
```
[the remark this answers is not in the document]
dimpar: I took older and newer version of deployment `BondedECDSAKeepFactory` timestamps with precision to seconds and the diff was  `10678946 sec` which is `123.59 days`. I wasn't sure if two term length should overlap if I put `124 days`. I think the last keep was opened in May anyway so I decided in favour of `123 days`. I can change it to `124 days` if it's okay for overlapping timestamps (about 9.5h timeframe).

solidity/contracts/ECDSABackportRewards.sol
```diff
@@ -0,0 +1,127 @@
...
+
+    // There were 41 Keeps created by BondedECDSAKeepFactory : 0x18758f16988E61Cd4B61E6B930694BD9fB07C22F
+    // The last Keep was opened on May-17-2020
+    // https://etherscan.io/address/0x45A3cACA2F2a78A53607618651C86111c9720AA5
+    uint256 internal constant numberOfCreatedECDSAKeeps = 41;
```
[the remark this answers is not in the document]
dimpar: Sure, now we read it from the factory contract.

solidity/contracts/ECDSABackportRewards.sol
dimpar: Added additional comment.

solidity/contracts/ECDSABackportRewards.sol
dimpar: Yea, you're right, I think we can remove that additional check.

solidity/contracts/ECDSABackportRewards.sol
dimpar: It should. I had it the way you wrote it, but accidentally replaced it with `bytes32(i)` when debugging tests :/

solidity/contracts/ECDSABackportRewards.sol
```diff
@@ -0,0 +1,118 @@
...
+            minimumECDSAKeepsPerInterval
+        )
+    {
+        factory = BondedECDSAKeepFactory(_factoryAddress);
+    }
+
+    function _getKeepCount() internal view returns (uint256) {
+        // Between May 13 2020 - Sep 14 2020 there were 41 keeps opened.
+        // The last Keep was opened on May-17-2020
+        // https://etherscan.io/address/0x45A3cACA2F2a78A53607618651C86111c9720AA5
```
pdyraga, line 55: I'd remove this comment. We don't care how many of them were created in this contract, the factory tells us about it and we do all the calculations using the number the factory returns.

solidity/contracts/ECDSABackportRewards.sol
```diff
@@ -0,0 +1,118 @@
...
+
+    function _isTerminated(bytes32 groupIndexBytes)
+        internal
+        view
+        returns (bool)
+    {
+        return false;
+    }
+
+    // A keep is recognized if it was created by this factory between May 13 2020 - Sep 14 2020
```
pdyraga, line 85: Nope, a keep is recognized if it was created by this factory no matter when. This is what the code in this function does - we removed the check with the end of interval.

solidity/test/rewards/TestECDSABackportRewards.js
```diff
@@ -0,0 +1,258 @@
...
+
+  const keepSize = 16
+  const numberOfCreatedKeeps = 41
+  const keepCreationTimestamp = 1589408353
+
+  // 1,000,000,000 - total KEEP supply
+  //   200,000,000 - 20% of the total supply goes to staker rewards
+  //   180,000,000 - 90% of staker rewards goes to the ECDSA stakers
+  //     1,800,000 - 1% of ECDSA staker rewards goes to May - Sep keeps
+  const ECDSABackportKEEPRewards = 1800000
```
pdyraga, line 58: To test on real numbers, we need to introduce something like
```
const tokenDecimalMultiplier = web3.utils.toBN(10).pow(web3.utils.toBN(18))
```
and do:
```
const totalRewards = web3.utils.toBN(1800000).mul(tokenDecimalMultiplier)
```

solidity/test/rewards/TestECDSABackportRewards.js
```diff
@@ -0,0 +1,258 @@
...
+  let tokenGrant
+  let keepFactory
+  let bondedSortitionPoolFactory
+  let keepBonding
+  let randomBeacon
+  let keepMembers
+
+  const owner = accounts[0]
+
+  const keepSize = 16
```
pdyraga, line 50: I'd prefer if we were keeping it at `3` simulating the real situation.

solidity/test/rewards/TestECDSABackportRewards.js
```diff
@@ -0,0 +1,258 @@
...
+
+  async function createKeeps() {
+    const members = keepMembers.map((m) => m.operator)
+    for (let i = 0; i < numberOfCreatedKeeps; i++) {
+      await keepFactory.stubOpenKeep(
+        owner,
+        members,
+        tokenStaking.address,
+        keepFactory.address,
+        keepCreationTimestamp
```
pdyraga, line 247: Can we create them at different timestamps? We don't need to equally distribute them between the interval but we can e.g. have them created over the course of 4 days, adding 2 hours to each timestamp in the loop.

solidity/contracts/ECDSABackportRewards.sol
dimpar: Removed

solidity/contracts/ECDSABackportRewards.sol
dimpar: Right, the comment was about broader context, not this specific function. Dates removed.

solidity/test/rewards/TestECDSABackportRewards.js
dimpar: Good point, I missed the fact that we also multiply by 10^18 upon deployment. Added.

solidity/test/rewards/TestECDSABackportRewards.js
dimpar: Sure, done.

solidity/test/rewards/TestECDSABackportRewards.js
dimpar: Done.

solidity/contracts/ECDSABackportRewards.sol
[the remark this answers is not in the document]
pdyraga: I think it's fine to have them overlapping a little. What matters here most is the transparency on how rewards are allocated. Calculating just days gives 124 diff and it's close to unix timestamp difference, especially that it took 12 hours to initialize first stakes on the new contract.

What I'd do is to set it to 124 days and drop the comment with seconds because neither 123 nor 124 days equal 10678946 seconds. I'd rather add a comment on the top of this constant, explaining there were about 124 days between the two events.

solidity/contracts/ECDSABackportRewards.sol
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    // A keep is recognized if it was created by this factory.
+    function _recognizedByFactory(bytes32 _keep) internal view returns (bool) {
+        return factory.getKeepOpenedTimestamp(toAddress(_keep)) != 0;
+    }
+
+    function _distributeReward(bytes32 _keep, uint256 amount)
+        internal
+        isAddress(_keep)
```
pdyraga, line 89: Do we need this check? This function is called from `Rewards` abstract contract only when `_recognizedByFactory` returns `true` (see `factoryMustRecognize` modifier).

If `isAddress` is not needed, then `validAddressBytes` is also not needed.

solidity/contracts/test/BondedECDSAKeepFactoryStub.sol
```diff
@@ -35,4 +35,35 @@ contract BondedECDSAKeepFactoryStub is BondedECDSAKeepFactory {
...
+        BondedECDSAKeep keep = BondedECDSAKeep(keepAddress);
+        keep.initialize(
+            _owner,
+            _members,
+            0,
+            0,
+            0,
+            _tokenStaking,
+            address(0),
+            _keepFactory
```
pdyraga, line 64: Why are we passing it as a constructor parameter instead of doing `address(this)`?

solidity/contracts/test/BondedECDSAKeepFactoryStub.sol
```diff
@@ -35,4 +35,35 @@ contract BondedECDSAKeepFactoryStub is BondedECDSAKeepFactory {
...
+        keepAddress = createClone(masterKeepAddress);
+        BondedECDSAKeep keep = BondedECDSAKeep(keepAddress);
+        keep.initialize(
+            _owner,
+            _members,
+            0,
+            0,
+            0,
+            _tokenStaking,
+            address(0),
```
pdyraga, line 63: `BondedECDSAKeepFactory` has `KeepBonding keepBonding` field so we can pass this address here.

solidity/contracts/test/BondedECDSAKeepFactoryStub.sol
```diff
@@ -35,4 +35,35 @@ contract BondedECDSAKeepFactoryStub is BondedECDSAKeepFactory {
...
+    /// @param _tokenStaking Token staking address.
+    /// @param _keepFactory Keep factory address.
+    /// @param _creationTimestamp Keep creation timestamp.
+    ///
+    /// @return Created keep address.
+    function stubOpenKeep(
+        address _owner,
+        address[] memory _members,
+        address _tokenStaking,
+        address payable _keepFactory,
```
pdyraga, line 51: I don't think we should have them as parameters. `BondedECDSAKeepFactory` has fields:
```
    TokenStaking tokenStaking;
    KeepBonding keepBonding;
```
and the values are passed in stub's constructor.

solidity/contracts/ECDSABackportRewards.sol
```diff
@@ -0,0 +1,115 @@
...
+                           Trust math, not hardware.
+*/
+
+pragma solidity 0.5.17;
+
+import "@keep-network/keep-core/contracts/Rewards.sol";
+import "./BondedECDSAKeepFactory.sol";
+import "./api/IBondedECDSAKeep.sol";
+
+contract ECDSABackportRewards is Rewards {
```
pdyraga, line 21: Can we add docs to this contract explaining what is it used for? Similarly to how `BeaconBackportRewards` contract role is explained.

solidity/test/rewards/TestECDSABackportRewards.js
```diff
@@ -0,0 +1,253 @@
...
+  let bondedSortitionPoolFactory
+  let keepBonding
+  let randomBeacon
+  let keepMembers
+
+  const owner = accounts[0]
+
+  const keepSize = 3
+  const numberOfCreatedKeeps = 41
+  const keepCreationTimestamp = 1589408353
```
pdyraga, line 52: Perhaps `firstKeepCreationTimestamp` given that we are creating them now each 2 hours after another?

solidity/test/rewards/TestECDSABackportRewards.js
```diff
@@ -0,0 +1,253 @@
...
+  const ECDSABackportKEEPRewards = web3.utils
+    .toBN(1800000)
+    .mul(tokenDecimalMultiplier)
+
+  before(async () => {
+    await BondedSortitionPoolFactory.detectNetwork()
+    await BondedSortitionPoolFactory.link(
+      "StackLib",
+      (await StackLib.new({from: owner})).address
+    )
```
pdyraga, line 68: Shouldn't we do it inside `initializeNewFactory` function?

solidity/test/rewards/TestECDSABackportRewards.js
dimpar: I've added a comment above. `firstKeepCreationTimestamp += 7200` and then passing in a loop `firstKeepCreationTimestamp` wouldn't be true, because it's a consecutive keep, not the first one.

solidity/contracts/ECDSABackportRewards.sol
```diff
@@ -0,0 +1,115 @@
...
+        return factory.getKeepCount();
+    }
+
+    function _getKeepAtIndex(uint256 index) internal view returns (bytes32) {
+        return fromAddress(factory.getKeepAtIndex(index));
+    }
+
+    function _getCreationTime(bytes32) internal view returns (uint256) {
+        // Assign each keep to the starting timestamp of its interval.
+        return startOf(0);
```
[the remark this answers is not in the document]
dimpar: Right, done.

solidity/test/rewards/TestECDSABackportRewards.js
```diff
@@ -0,0 +1,254 @@
...
+      // KEEP is added to the signers in every iteration; total 41 times (number of keeps).
+      const expectedSingleReward = new BN(14634)
+
+      for (let i = 0; i < numberOfCreatedKeeps; i++) {
+        const keepAddress = await keepFactory.getKeepAtIndex(i)
+        await rewardsContract.receiveReward(keepAddress)
+
+        await assertKeepBalanceOfBeneficiaries(i, expectedSingleReward)
+      }
+    })
```
pdyraga, line 120: I'd simplify it to checking just the final balances. The logic inside `assertKeepBalanceOfBeneficiaries` with `keepNumber` is harder to grasp and individual distributions are tested in unit tests for parent `Rewards` contract anyway. Finally, we know that in practice we are talking about 3 operators for this backport. 🤷 

For example:
```
    it.only("should correctly distribute rewards between beneficiaries", async () => {
      // 1800000 / 3 = 600000
      const expectedBeneficiaryBalance = web3.utils.toBN(600000)

      for (let i = 0; i < numberOfCreatedKeeps; i++) {
        const keepAddress = await keepFactory.getKeepAtIndex(i)
        await rewardsContract.receiveReward(keepAddress)
      }

      assertKeepBalanceOfBeneficiaries(expectedBeneficiaryBalance) 
    })

    async function assertKeepBalanceOfBeneficiaries(expectedBalance) {
      // solidity is not very good when it comes to floating point precision,
      // we are allowing for +-1 KEEP difference margin between expected and
      // actual value
      const precision = 1

      for (let i = 0; i < keepMembers.length; i++) {
          const balance = await keepToken.balanceOf(keepMembers[i].beneficiary)
          const balanceInKeep = balance.div(tokenDecimalMultiplier)
          expect(balanceInKeep).to.gte.BN(expectedBalance.subn(precision))
          expect(balanceInKeep).to.lte.BN(expectedBalance.addn(precision))
      }
  }
```

solidity/test/rewards/TestECDSABackportRewards.js
```diff
@@ -0,0 +1,254 @@
...
+
+  // 1,000,000,000 - total KEEP supply
+  //   200,000,000 - 20% of the total supply goes to staker rewards
+  //   180,000,000 - 90% of staker rewards goes to the ECDSA stakers
+  //     1,800,000 - 1% of ECDSA staker rewards goes to May - Sep keeps
+  const ECDSABackportKEEPRewards = web3.utils
+    .toBN(1800000)
+    .mul(tokenDecimalMultiplier)
+
+  let keepCreationTimestamp = 1589408353
```
pdyraga, line 62: Why this value is not `1589408351 + 1`?

solidity/contracts/ECDSABackportRewards.sol
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    function _isClosed(bytes32) internal view returns (bool) {
+        // Even though we still have some of the keeps opened, all the keeps
+        // created between May 13 2020 - Sep 14 2020 are considered closed. Because
+        // of the deposits pause https://tbtc.network/news/2020-05-21-details-of-the-tbtc-deposit-pause-on-may-18-2020/
+        // closing all the keeps is not easily achievable. However, we do not want
+        // to block rewards distribution for good stakers caused by the incident on May 18th.
```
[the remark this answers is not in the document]
dimpar: Done.

solidity/contracts/ECDSABackportRewards.sol
dimpar: Added, but wasn't sure if we should include a comment about new functions?

solidity/test/rewards/TestECDSABackportRewards.js
pdyraga: Hmm... I think is is confusing. Maybe we can have:

```  
const keepCreationTimestamp = 1589408352
```

and later:
```
  async function createKeeps() {
    const members = keepMembers.map((m) => m.operator)
    let keepCreationTimestamp = firstKeepCreationTimestamp
    for (let i = 0; i < numberOfCreatedKeeps; i++) {
      await keepFactory.stubOpenKeep(
        owner,
        members,
        tokenStaking.address,
        keepFactory.address,
        keepCreationTimestamp
      )
      keepCreationTimestamp += 7200 // adding 2 hours interval between each opened keep
    }
  }
```
?

solidity/contracts/ECDSABackportRewards.sol
dimpar: I see your point that by the time we call `_distributeReward`, `_keep` will be already verified, however we're calling this from the abstract contract. If we are not worried that potentially in the future during the upgrades we won't miss that check, then I can remove it.

solidity/contracts/ECDSABackportRewards.sol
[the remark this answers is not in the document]
dimpar: Set it to 124 days.

solidity/contracts/ECDSABackportRewards.sol
pdyraga: I'd remove it, it's only increasing gas consumption and this contract needs to trust the parent contract.

solidity/contracts/test/BondedECDSAKeepFactoryStub.sol
dimpar: No reason, just a mistake. I replaced it with `address(this)`

solidity/contracts/test/BondedECDSAKeepFactoryStub.sol
dimpar: Done.

solidity/test/rewards/TestECDSABackportRewards.js
dimpar: Ok, done.

solidity/test/rewards/TestECDSABackportRewards.js
dimpar: Glitch in the matrix. Changed to `1589408352`

solidity/contracts/test/BondedECDSAKeepFactoryStub.sol
dimpar: Right, done.

solidity/test/rewards/TestECDSABackportRewards.js
dimpar: Ok, done.

pdyraga (approved):

PR #569, opened 2020-10-02
Ban members of FullyBackedBondedECDSAKeep in case of signature fraud

nkuba:
In this PR we added functionality of banning keep members in case of a signature fraud for FullyBackedECDSAKeeps.

Once a signature fraud is discovered a keep will call the associated factory to ban its' members. The factory will traverse over all already created sortition pools and ban the keep members.

Depends on: https://github.com/keep-network/keep-ecdsa/pull/483

### DIFF
--- a/solidity/contracts/fully-backed/FullyBackedECDSAKeepFactory.sol
+++ b/solidity/contracts/fully-backed/FullyBackedECDSAKeepFactory.sol
@@ -246,7 +246,7 @@ contract FullyBackedECDSAKeepFactory is
         return bonding.isAuthorizedForOperator(_operator, address(this));
     }
 
-    /// @notice Bans members of a calling keep in an sortition pools associated
+    /// @notice Bans members of a calling keep in a sortition pool associated
     /// with the application for which the keep was created.
     /// @dev The function can be called only by a keep created by this factory.
     function banKeepMembers() public onlyKeep() {
--- a/solidity/contracts/fully-backed/FullyBackedECDSAKeepFactory.sol
+++ b/solidity/contracts/fully-backed/FullyBackedECDSAKeepFactory.sol
@@ -76,6 +76,10 @@ contract FullyBackedECDSAKeepFactory is
     // eligible weight for signer selection.
     uint256 public constant bondWeightDivisor = 1 ether;
 
+    // Maps a keep to an application for which the keep was created.
+    // keep address -> application address
+    mapping(address => address) keepApplication;
+
     // Notification that a new keep has been created.
     event FullyBackedECDSAKeepCreated(
         address indexed keepAddress,
@@ -84,6 +88,10 @@ contract FullyBackedECDSAKeepFactory is
         address indexed application,
         uint256 honestThreshold
     );
+
+    // Notification when an operator gets banned in a sortition pool for
+    // an application.
+    event OperatorBanned(address indexed operator, address indexed application);
 
     constructor(
         address _masterKeepAddress,
@@ -194,6 +202,8 @@ contract FullyBackedECDSAKeepFactory is
             );
         }
 
+        keepApplication[keepAddress] = application;
+
         emit FullyBackedECDSAKeepCreated(
             keepAddress,
             members,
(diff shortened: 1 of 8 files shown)


solidity/contracts/fully-backed/FullyBackedECDSAKeepFactory.sol
```diff
@@ -236,11 +255,32 @@ contract FullyBackedECDSAKeepFactory is
...
+    /// @dev The function can be called only by a keep created by this factory.
+    function banKeepMembers() public onlyKeep() {
+        FullyBackedECDSAKeep keep = FullyBackedECDSAKeep(msg.sender);
+
+        address[] memory members = keep.getMembers();
+
+        for (uint256 i = 0; i < members.length; i++) {
+            address operator = members[i];
+
+            for (uint256 j = 0; j < applications.length; j++) {
```
pdyraga, line 269: We may get into problems here. There is a chance adversary may create a bunch of applications and this call will fail with out-of-gas so reporting frauds will not be possible. Instead of banning from all sortition pools, we should ban from the specific sortition pool.

I believe it should be farily easy as the owner of the keep is the application, right? So we can get the application address from the keep. Worth confirming...

nkuba: This was also my concern, I'm glad you brought it up. Unfortunatelly solving this is not that straightforward, owner of a keep is a Deposit not an Application. I'll try to look for a solution.

pdyraga: Well, even if the owner was an application, assuming owner == application would suck for the API because owner is a parameter and the application may chose anyone as an owner which is actually happening here. I think the easiest option is to add a map `keep -> application` in `FullyBackedECDSAKeepFactory`.

nkuba: > I think the easiest option is to add a map keep -> application in FullyBackedECDSAKeepFactory.

Yep, I'm already working on this.

solidity/contracts/fully-backed/FullyBackedECDSAKeepFactory.sol
```diff
@@ -68,14 +68,20 @@ contract FullyBackedECDSAKeepFactory is
...
+    uint256 public constant defaultMinimumBond = 20 ether;
 
     // Signer candidates in bonded sortition pool are weighted by their eligible
     // stake divided by a constant divisor. The divisor is set to 1 ETH so that
     // all ETHs in available unbonded value matter when calculating operator's
     // eligible weight for signer selection.
     uint256 public constant bondWeightDivisor = 1 ether;
 
+    // List of applications that got sortition pool created.
+    address[] applications;
```
pdyraga, line 80: I think we no longer need this field.

solidity/contracts/fully-backed/FullyBackedECDSAKeepFactory.sol
```diff
@@ -68,14 +68,20 @@ contract FullyBackedECDSAKeepFactory is
...
+
+    // Maps an application for which keep was created to each keep address.
+    mapping(address => address) keepApplication;
```
pdyraga, line 80: I think it's mapping keep to application so instead of 
```
// Maps an application for which keep was created to each keep address.
```

we should have:
```
// Maps a keep to an application for which the keep was created for.
// keep address -> application address
```

solidity/contracts/CandidatesPools.sol
```diff
@@ -75,6 +53,28 @@ contract CandidatesPools {
             .updateOperatorStatus(_operator);
     }
 
+    /// @notice Creates new sortition pool for the application.
+    /// @dev Emits an event after sortition pool creation.
+    /// @param _application Address of the application.
+    /// @return Address of the created sortition pool contract.
+    function createSortitionPool(address _application)
+        public
```
pdyraga, line 61: Should we revert the change and make it back `external`?

solidity/contracts/fully-backed/FullyBackedECDSAKeepFactory.sol
```diff
@@ -236,11 +246,34 @@ contract FullyBackedECDSAKeepFactory is
         return bonding.isAuthorizedForOperator(_operator, address(this));
     }
 
+    /// @notice Bans members of a calling keep in an sortition pools associated
+    /// with the application for which the keep was created.
```
pdyraga, line 250: Documentation is no longer up to date here.

pdyraga (approved):

pdyraga:
Merged, I have also let myself to update the PR description to reflect the final solution we came up with here.

PR #570, opened 2020-10-05
Check operator balance before status update

lukasz-zimnoch:
Closes: #551

This pull request introduces a new `MinimumKeyGenerationBalance` config property which defines a minimum WEI balance below which the operator should not participate in new keeps. The main motivation is to avoid spending funds during the key generation protocol and hold them for signatures requests to mitigate the risk of bonds seizing by not having enough ETH to submit requested signatures to the chain.

Technically, the aforementioned behavior is achieved by checking the current operator's balance against the `MinimumKeyGenerationBalance` value and aborting the operator status update in the sortition pool in case the current balance is below the minimum. Not updating the operator status will force the sortition pool to omit the operator during group selection.

### DIFF
--- a/configs/config.toml.SAMPLE
+++ b/configs/config.toml.SAMPLE
@@ -74,11 +74,11 @@
 #  KeyGenerationTimeout = "3h" 				# optional
 #  SigningTimeout = "2h"					# optional
 #
-# Specifies the minimum operator ETH balance to participate in new keeps.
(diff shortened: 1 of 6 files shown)


lukasz-zimnoch:
After a discussion with @pdyraga and @nkuba the approach presented here is not enough and it will require a lot of changes in the contract side. For now we limit ourselves to the monitoring feature. Closed in favour of https://github.com/keep-network/keep-ecdsa/pull/572.

PR #572, opened 2020-10-06
Operator balance monitoring

lukasz-zimnoch:
Refs: keep-network/keep-ecdsa#551
Depends on: keep-network/keep-common#51
Depends on: https://github.com/keep-network/keep-core/pull/2111
Depends on: https://github.com/keep-network/keep-common/pull/59

Here we implement a simple balance monitor and enable it for the operator account balance. The monitoring routine checks the balance every 10 minutes and display a log error in case the balance falls below a threshold value passed from the config file.


cmd/start.go
```diff
@@ -247,3 +251,34 @@ func initializeDiagnostics(
...
+	alertThreshold := config.Ethereum.BalanceAlertThreshold
+	if alertThreshold == 0 {
+		alertThreshold = 500000000000000000 // 0.5 ETH by default.
+	}
+
+	balanceMonitor.Observe(
+		ctx,
+		ethereumAddress,
+		new(big.Int).SetUint64(alertThreshold),
+		10*time.Minute,
```
dimpar, line 275: Same with this guy, thoughts on declaring this value at the top as const?

cmd/start.go
```diff
@@ -247,3 +251,34 @@ func initializeDiagnostics(
...
+	)
+
+	logger.Infof(
+		"started balance monitoring for address [%v] "+
+			"with alerting below [%v] wei",
```
dimpar, line 280: "with the alert threshold set to [%v] wei"? up to you.

cmd/start.go
```diff
@@ -247,3 +251,34 @@ func initializeDiagnostics(
...
+
+func initializeBalanceMonitoring(
+	ctx context.Context,
+	chainHandle eth.Handle,
+	config *config.Config,
+	ethereumAddress string,
+) {
+	balanceMonitor, err := chainHandle.BalanceMonitor()
+	if err != nil {
+		logger.Errorf("error obtaining balance monitor handle [%v]", err)
```
dimpar, line 263: Don't we want to return just like for stake monitoring? If there's a problem here, then I think it will blow up somewhere down the road, ex. line 271
```
return fmt.Errorf("error obtaining balance monitor handle: [%v]", err)
```

pkg/chain/ethereum/ethereum.go
```diff
@@ -507,3 +509,17 @@ func (ec *EthereumChain) GetOpenedTimestamp(keepAddress common.Address) (time.Ti
 
 	return keepOpenTime, nil
 }
+
+// WeiBalanceOf returns the given address balance in WEI.
```
dimpar, line 513: Maybe we should add to this comment that it is a balance of the latest known block? Something like:
```
WeiBalanceOf returns the wei balance of the given address from the latest known block.
```

cmd/start.go
lukasz-zimnoch: Yep, good catch! I missed a `return` here.

lukasz-zimnoch:
@pdyraga ready for a next look.

pdyraga (approved):

PR #573, opened 2020-10-09
RFC 1: tBTC-specific ECDSA client actions

Shadowfiend:
This was a reasonable first pass, and was used as the basis for the implementation
that has since gone into the client proper. It is not a 1-to-1 description of the current
implementation.

See #574.

### DIFF
--- a/docs/rfc/rfc-1.adoc
+++ b/docs/rfc/rfc-1.adoc
@@ -1,0 +1,230 @@
+:toc: macro
+
+= RFC 1: tBTC-specific ECDSA client actions
+
+:icons: font
+:numbered:
+toc::[]
+
+== Background
+
+The Keep ECDSA client is built as a generic client for ECDSA keeps. In turn,
+ECDSA keeps are designed to be on-chain smart contract representations of a
+signing group that can provide on-demand key generation and signature
+services for a consumer application (represented by another smart contract).
+Currently, the signer responsibilities encapsulated by the ECDSA client are
+exclusively limited to direct responsibilities assigned to ECDSA keeps. However,
+specific applications, like tBTC, have additional expectations or incentivized
+behaviors for the signing parties in their protocols. In these cases, the
+ECDSA client as currently built has no ability to provide these additional
+functions.
+
+This RFC seeks to outline a short-term improvement to the client that can deal
+with the immediate incentivized actions in tBTC's signing group behaviors, so
+that signers are behaving optimally with respect to the underlying protocol.
+
+=== Current Functionality
+
+ECDSA clients currently monitor for three classes of events:
+
+ - Requests for a new keep.
+ - Requests for a signature from a keep the client is already participating in.
+ - Notifications that a keep has closed, either normally or abnormally.
+
+This monitoring is currently closely tied to underlying Ethereum implementation
+details, and has no external dependencies beyond the existing Ethereum
+connection. It is completely application-agnostic; the client's only interest in
+the application is to know whether the current operator account is authorized
+and has available bond for the application in question.
+
+In response to the above requests, the client may do a few things:
+
+New keep requests::
+    Join an ECDSA unicast channel, perform distributed key generation, and
+    publish the key to the ECDSA keep contract.
+Signature requests::
+    Join an ECDSA unicast channel, perform a single-round threshold signature
+    protocol, and publish the resulting signature to the ECDSA keep contract.
+Close notification::
+    Clean up any resources and event handlers, and archive key material for
+    the closed keep.
+
+== Proposal
+
+=== Goal
+
+This proposal is focused on extending both monitoring and reactions for ECDSA
+keeps to include additional actions that are specific to tBTC. In particular,
+these are:
+
+ - Calling `retrievePubkey` on the tBTC deposit that is associated with a given
+   keep once the key is generated and published, if the signature setup timeout
+   is approaching.
+ - Calling `provideRedemptionProof` on a tBTC deposit if it is redeemed,
+   sufficiently confirmed, and the redemption proof timeout is approaching.
+ - Calling `increaseRedemptionFee` on the tBTC deposit that is associated with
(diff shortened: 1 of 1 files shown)


lukasz-zimnoch:
The approach presented here assumes we will add all app-specific behavior directly in the ECDSA client. I understand this is probably the easiest and fastest way for now. But, I have a strong feeling this will lead us towards a hard to maintain code as more applications are added in the future. Also, we will be probably forced to introduce multiple app-specific dependencies in the ECDSA client code (similar to `tbtc.Deposit`) which will complicate things even more. Maybe I'm not aware of all circumstances and security considerations for now, but maybe we should consider some kind of modular approach? For example using https://github.com/hashicorp/go-plugin (or similar) to implement app-specific actions as separate plugins and use them as subprocesses of the ECDSA client.

pdyraga:
My opinion is that we need to design the interfaces to make them losely-coupled and later decide on how we eventually want to separate application-specific behavior from the main client. We don't need to have the final solution now and deciding on the final solution is not possible given that I am sure we will find and learn more during the implmenetation.

Having two processes for client and maintainer is a non-go in my opinion, given the additional complexity of running BTC and Electum node by stakers that we will require anyway.

What I would like to achieve is that it's a tbtc-app-specific module that listenes to keep-ecdsa interface and not the other way. I would also propose to start the work with two independent workstreams:

1. `keep-ecdsa` talks to Electrum and knows how to generated SPV proof.
2. `keep-ecdsa` imports `tbtc` contract Go bindings and can call `retrievePubKey`

Once we have those two we can work on `provideRedemptionProof` and `increaseRedemptionFee`. 

`[tbtc]` should be a separate section in `config.toml` that is optional. It should have their own set of operator addresses to monitor - this could be helpful for staking providers if they want to monitor multiple operators from one client given Electrum and BTC connection is available only on that machine.

For _tbtc-app-specific module that attaches to keep-ecdsa interface and not the other way_ piece, I think we can achieve it by watching keep created events for this node so app-specific interface is listenting to the main client's interface, and reacts by setting up the monitoring on their own. The easiest way to keep them losely-coupled is probably to organize this communication in a form of an event bus.

lukasz-zimnoch:
> Having two processes for client and maintainer is a non-go in my opinion, given the additional complexity of running BTC and Electum node by stakers that we will require anyway.

To clarify, I didn't mean forcing the operator to manage two distinct processes. Giving the example of Hashicorp's `go-plugin` lib, I rather had in mind running app-specific plugins as a subprocesses fully controlled by the main ECDSA client. From the operator's perspective, nothing changes. They just have new sections to manage in the config file. 

Anyway, this was only a proposition for the general approach. I agree with your point to start with small steps but it will be good to know what structure we are trying to achieve eventually.

pdyraga:
> > Having two processes for client and maintainer is a non-go in my opinion, given the additional complexity of running BTC and Electum node by stakers that we will require anyway.
> 
> To clarify, I didn't mean forcing the operator to manage two distinct processes. Giving the example of Hashicorp's `go-plugin` lib, I rather had in mind running app-specific plugins as a subprocesses fully controlled by the main ECDSA client. From the operator's perspective, nothing changes. They just have new sections to manage in the config file.
> 
> Anyway, this was only a proposition for the general approach. I agree with your point to start with small steps but it will be good to know what structure we are trying to achieve eventually.

Absolutely agree. Eventually, this should be designed in a plugin/module-like fashion. For now, let's just keep the interfaces separate and make sure `tbtc` module reads from the main keep process and not vice-versa. The code proposed in this RFC is an early WIP but it's not that bad and is aligned with the idea of modularizatio. I propose to start with point 2 (calling `retrievePubKey`) as it's where we answer most of those questions. Before we start working on point 1 (SPV proofs), we'll have tBTC contracts imported and we'll know how to read from config and interact with the chain.

Shadowfiend:
This was what I was getting at here:

> This proposal is very narrowly focused on shot-term changes. 

My intent was to want to avoid too much refactoring right now, while laying the foundation for the functionality we need.

We're definitely in need of a refactoring in the client. Right now our interfaces in the beacon client make a clear distinction between app-specific and chain/network functionality. In the ECDSA client these are all very tightly coupled---for example, there is no distinct concept of an ECDSA keep. It's just a handful of goroutines managing the lifecycle fired directly from the `client` package. This is deliberate, in that we chose early on to focus on completion and refactor later. My intent with this RFC is to continue that approach for a little longer and push the broader architectural rework (which we need for upcoming Celo work anyway) a little further. I have some ideas on that refactoring as well, which I'll probably try to write up in a separate RFC for feedback… But probably not this week.

 > maybe we should consider some kind of modular approach

Fully agree in terms of modularity in functionality. Using subprocesses vs compiling it into the same executable I think will have to depend on how we observe runtime behavior, how likely folks are to use all applications when there are more applications, how many applications there are, and how we want recovery to work (e.g., child processes can sometimes continue operating when a parent process dies---will that be desirable or not). Probably more factors to consider, but I agree it's something we'll want to keep an eye on and think about.

mhluongo:
Calling out that the two things we see growing on the product aide are

* More host chains
* Other versions of tBTC that will likely have slight different protocol designs

It's unclear to me how the client should map. Do operators want to stake KEEP on Ethereum and Celo?

Often having separate binaries per chain is considered a security improvement lest a less common chain have a vuln that can be chained to a more common (and valuable) chain. We've seen this pattern in exchanges and custody, and I expect it holds here.

Shadowfiend:
@pdyraga @lukasz-zimnoch it feels like we've gone ahead and done some of this---and maybe in the way described in this RFC. Should we update, undraft, and merge?

pdyraga:
I am good to merge it as-is on my side, let's just maybe remove the note about links to come at the bottom. The implementation we have in the client is pretty similar to what's been proposed here and we do not need to keep it 1:1 IMO, the client code is always evolving.

pdyraga (approved):

Issue #446, opened 2020-05-11
ECDSA staker KEEP rewards


pdyraga:
Rewards to distribute KEEP tokens to keep signers based on a predetermined curve separated by intervals.

pdyraga:
There is yet one nice-to-have before we deploy the contract: allow stakers to claim rewards from multiple keeps of their choice in one transaction. It should be up to the caller to determine the cost and make sure the operation does not consume too much gas to fit in one block.

PR #579, opened 2020-10-14
Allowing stakers to receive rewards for multiple keeps

dimpar:
In order to reduce cost of single multiple transactions when claiming rewards, we implement a function that takes an array of keep addresses and process all claims in a single transaction.
A caller is responsible to determine the cost and consumed gas for this transaction.

Refs https://github.com/keep-network/keep-ecdsa/issues/446

### DIFF
--- a/solidity/contracts/ECDSARewards.sol
+++ b/solidity/contracts/ECDSARewards.sol
@@ -102,7 +102,7 @@ contract ECDSARewards is Rewards {
         factory = BondedECDSAKeepFactory(_factoryAddress);
     }
 
-    /// @notice Stakers can receive KEEP rewards from multiple keeps of their choice 
+    /// @notice Stakers can receive KEEP rewards from multiple keeps of their choice
     /// in one transaction to reduce total cost comparing to single calls for rewards.
     /// It is a caller responsibility to determine the cost and consumed gas when
     /// receiving rewards from multiple keeps.
--- a/solidity/contracts/ECDSARewards.sol
+++ b/solidity/contracts/ECDSARewards.sol
@@ -102,6 +102,16 @@ contract ECDSARewards is Rewards {
         factory = BondedECDSAKeepFactory(_factoryAddress);
     }
 
+    // Stakers can receive KEEP rewards from multiple keeps of their choice in
+    // one transaction to reduce total cost comparing to single calls for rewards.
+    // It is a caller responsibility to determine the cost and consumed gas when
+    // receiving rewards from multiple keeps.
+    function receiveReward(bytes32[] memory keepIdentifiers) public {
+        for (uint256 i = 0; i < keepIdentifiers.length; i++) {
+            receiveReward(keepIdentifiers[i]);
+        }
+    }
+
     function _getKeepCount() internal view returns (uint256) {
         return factory.getKeepCount();
     }
--- a/solidity/contracts/ECDSARewards.sol
+++ b/solidity/contracts/ECDSARewards.sol
@@ -102,11 +102,12 @@ contract ECDSARewards is Rewards {
         factory = BondedECDSAKeepFactory(_factoryAddress);
     }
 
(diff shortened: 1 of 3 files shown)


solidity/contracts/ECDSARewards.sol
```diff
@@ -102,6 +102,16 @@ contract ECDSARewards is Rewards {
         factory = BondedECDSAKeepFactory(_factoryAddress);
     }
 
+    // Stakers can receive KEEP rewards from multiple keeps of their choice in
+    // one transaction to reduce total cost comparing to single calls for rewards.
+    // It is a caller responsibility to determine the cost and consumed gas when
+    // receiving rewards from multiple keeps.
```
pdyraga, line 108: Docs should follow the pattern:
```
/// @notice ...
///
/// @param keepIdentifiers ...
```

solidity/test/rewards/TestECDSARewards.js
```diff
@@ -115,20 +115,22 @@ describe("ECDSARewards", () => {
...
       const keepAddress = await keepFactory.getKeepAtIndex(0)
+      const keepAddresses = [keepAddress]
 
       const isEligible = await rewardsContract.eligibleForReward(keepAddress)
       expect(isEligible).to.be.false
 
       await timeJumpToEndOfInterval(0)
       await expectRevert(
-        rewardsContract.receiveReward(keepAddress),
+        rewardsContract.receiveReward(keepAddresses),
```
pdyraga, line 125: Let's rename the new function to `receiveRewards` (plural) and we'll be able to avoid overloading problems.

solidity/test/rewards/TestECDSARewards.js
```diff
@@ -202,14 +206,45 @@ describe("ECDSARewards", () => {
...
+      await keep.publicMarkAsClosed()
+      await keep1.publicMarkAsClosed()
+      await keep2.publicMarkAsClosed()
+
+      await rewardsContract.receiveReward(keepAddresses)
+
+      // reward for the first interval: 7,128,000 KEEP
+      // keeps created: 10 => 712,800 KEEP per keep
+      // member receives: 712,800 / 3 = 237,600 (3 signers per keep)
+      // member was in 3 properly closed keeps: 237,600 * 3 = 712,800
```
pdyraga, line 243: Can we do 2 or 4 properly closed keeps? Just to have a different number than 712,800 that is the same as an allocation for the whole keep.

solidity/contracts/ECDSARewards.sol
dimpar: Added annotations.

solidity/test/rewards/TestECDSARewards.js
dimpar: Done

solidity/test/rewards/TestECDSARewards.js
dimpar: We can, changed to 2 closed keeps.

pdyraga:
Linter does not seem to be particularly happy.

solidity/test/rewards/TestECDSARewards.js
```diff
@@ -214,6 +214,33 @@ describe("ECDSARewards", () => {
...
+      for (let i = 0; i < 10; i++) {
+        await keepFactory.stubOpenKeep(owner, operators, firstIntervalStart)
+      }
+
+      await timeJumpToEndOfIntervalIfApplicable(0)
+
+      const keepAddress = await keepFactory.getKeepAtIndex(0)
+      const keepAddress1 = await keepFactory.getKeepAtIndex(1)
+      const keepAddresses = [keepAddress, keepAddress1]
+      const keep = await BondedECDSAKeepStub.at(keepAddress)
```
pdyraga, line 228: Nitpick, not worth blocking on it but given that we have `keep1` I would name this one `keep0`.

pdyraga (approved):

Issue #574, opened 2020-10-10
Application-specific workflow actions


Shadowfiend:
Certain applications that use ECDSA keeps (e.g., tBTC) benefit strongly from additional actions beyond the base ECDSA keep actions. As a specific example, tBTC deposits have certain state transitions that are *typically* performed by users, but that are *incentivized* to be performed by signers. In these cases, failure to perform the state transition can have consequences on signers, up to and including partial or full bond seizure.

The ECDSA client should have a way to specify, encapsulate, and execute these application-specific behaviors, in a way that can be scaled to support multiple applications and multiple chains.

PR #580, opened 2020-10-14
Initial implementation of app-specific actions

lukasz-zimnoch:
Refs: #574
Depends on: https://github.com/keep-network/tbtc/pull/739

Here we introduce the backbone which allows adding application-specific actions straight into the `keep-ecdsa` client. As its first use, we also add the TBTC extension which allows invoking `retrieveSignerPubkey` action on new deposits which are close to exceeding the signing group formation timeout. Here is a brief summary of major changes:

- Added a new `[Extensions.TBTC]` section in the config which allows configuring the TBTC extension
- Implemented a TBTC wrapper around the existing Ethereum chain implementation which takes advantage of https://github.com/keep-network/tbtc/pull/739 and use these bindings to interact with TBTC contracts
- Developed a structure for app-specific extensions allowing to handle multiple applications and chains
- Implemented a TBTC extension with one `retrieveSignerPubkey` action allowing to add new actions as well
- Created a TBTC wrapper around the local chain implementation and developed several unit test scenarios which leverage that chain


configs/config.toml.SAMPLE
```diff
@@ -108,4 +108,9 @@
 # The port on which the `/diagnostics` endpoint will be available can be
 # customized below.
 # [Diagnostics]
-	# Port = 8081
\ No newline at end of file
+	# Port = 8081
+
+[Actions]
```
pdyraga, line 113: What do you think about naming it `[Applications]`? It's a little less cryptic to me than `[Actions]` and we'll be grouping configurations under application name (`[TBTC]`).

cmd/start.go
```diff
@@ -183,6 +187,23 @@ func Start(c *cli.Context) error {
 	}
 }
 
+func initializeApplicationSpecificActions(
```
pdyraga, line 190: Maybe just `initializeApplicationActions` for simplicity?

pkg/chain/ethereum/tbtc_extensions.go
```diff
@@ -0,0 +1,73 @@
...
+	"math/big"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/keep-network/keep-common/pkg/subscription"
+	"github.com/keep-network/tbtc/bindings/go/contract"
+)
+
+// TBTCEthereumChain represents an Ethereum chain handle with
+// TBTC-specific capabilities.
+type TBTCEthereumChain struct {
```
pdyraga, line 14: Would it be possible to have this guy living in `pkg/actions`? It looks wrong to me that we are polluting main client's ethereum code with application-specific component/

pkg/actions/tbtc_actions.go
```diff
@@ -0,0 +1,49 @@
+package actions
```
pdyraga, line 1: Maybe `applications/tbtc/tbtc.go`?

pkg/chain/ethereum/tbtc_extensions.go
lukasz-zimnoch: This was my first idea but:
- we need private stuff from `EthereumChain` 
- this fits better with the multi-chain approach we have to support.  If that goes to `pkg/actions` we have to mirror the entire structure of the `pkg/chain` here.

configs/config.toml.SAMPLE
lukasz-zimnoch: I think `[Applications]` can be confusing because we have already `[SanctionedApplications]` section. What do you think about `[ApplicationActions]`?

pdyraga: Maybe `ApplicationExtensions`? My reasoning is: we are _extending_ ECDSA client capabilities with whatever application-specific module wants to do. And those extensions will be eventually pluggable.

pkg/chain/ethereum/tbtc_extensions.go
pdyraga: Ideally, I think we should have the following structure:
- core Ethereum functions (keep key gen, signing, sortition pool interactions) defined in `pkg/chain/ethereum/`
- core Celo functions defined in `pkg/chain/celo`
- core Algorand functions defined in `pkg/chain/algorand` 
(...)

All chain functions defined by application extensions should be defined elsewhere, for example:
`pkg/extensions(actions)/tbtc/chain/ethereum`
`pkg/extensions(actions)/tbtc/chain/celo`

And so forth.

This is for the multi-chain approach. Regarding private stuff from `EthereumChain`, what stuff do we need? Maybe those pieces should no longer be private.

configs/config.toml.SAMPLE
lukasz-zimnoch: Yeah makes sense! I'll change it.

pkg/chain/ethereum/tbtc_extensions.go
lukasz-zimnoch: Ok, that structure makes sense. I'll try to align it in that way. Regarding private stuff, we need: `accountKey`, `client`, `nonceManager`, `miningWaiter`, `transactionMutex`. I think these elements should remain private. However, I will try to address that problem somehow.

pkg/applications/tbtc/chain/ethereum/ethereum.go
```diff
@@ -0,0 +1,182 @@
+package ethereum
```
pdyraga, line 1: I was going back and forward on this one in my head for 3 days and unfortunately, I don't think it is a good move and the original idea of keeping it all in `pkg/chain/ethereum` was better IMO.

The fact we need to cast the value returned from `CreateContract` to a specific implementation is the first warning signal. Looking close, we see that `CreateContract` exposes private fields of the base `ethereum` package hoping that the extension will treat them in a way that `ethereum` package expects. There is no difference between this and making those fields public, and we know they should not be public. The extension may not follow the rules and e.g. use a mutex in a way that will block all other calls. That's another warning signal.

Thinking more about it, `go.mod` is importing `tbtc` contract bindings anyway, so I think it's fine to expect - for now - that the client exposes etheruem API for those bindings and the responsibility of an extension is to implement app-specific actions in the correct way, using those bindings. If we decide in the future, we really need those applications to be pluggable, we should refactor base `ethereum` package to allow plugging-in new chain extension without having the base `ethereum` package exposing private properties.

What I would suggest for now - let's do it in a way that is not perfect, but safer and good enough, and not blocking us for future if we decide we really need chain extensions to be pluggable. My proposition - that is actually the same or very similar to what we had here originally:
- Let's keep it logic in a separate file, but in the base ethereum package, e.g. `pkg/chain/ethereum/tbtc.go`.
- Let's have a separate `TBTC` interface with TBTC-specific functions exposed by `Handle` in `pkg/chain/chain.go`, similar to `BlockCounter` or `StakeMonitor`
- The logic of the extension - the code using the chain and defining how to react to certain events - should stay separate, like it's done here now.

lukasz-zimnoch: Ok, agree but I think there is no need to:

> Let's have a separate TBTC interface with TBTC-specific functions exposed by Handle in pkg/chain/chain.go, similar to BlockCounter or StakeMonitor

I think we should rather keep the TBTC interface at the point of use, in the extension file, as we are doing it now. This way we follow the philosophy of Go interfaces and make the code more flexible. Also we don't need to expose the TBTC interface as part of our regular `Handle` interface because the TBTC part is used exclusively by the extension.

pdyraga: I may not see all the parts yet, but if you say it's going to work I like it!

configs/config.toml.SAMPLE
```diff
@@ -108,4 +108,9 @@
 # The port on which the `/diagnostics` endpoint will be available can be
 # customized below.
 # [Diagnostics]
-	# Port = 8081
\ No newline at end of file
+	# Port = 8081
+
+[ApplicationExtensions]
+	[ApplicationExtensions.TBTC]
```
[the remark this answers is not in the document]
lukasz-zimnoch: Nope, it can't be but we can have `[ApplicationExtensions.TBTC]`.

pkg/applications/tbtc/tbtc.go
```diff
@@ -0,0 +1,303 @@
...
+//  2. Incoming events deduplication.
+//  3. Resume extensions executions after client restart.
+func (em *extensionsManager) initializeExtension(
+	extensionName string,
+	triggerEventHandlerSetup depositEventHandlerSetup,
+	stopEventHandlerSetup depositEventHandlerSetup,
+	keepClosedHandlerSetup keepClosedHandlerSetup,
+	transactionSubmitter depositTransactionSubmitter,
+	timeout time.Duration,
+) (subscription.EventSubscription, error) {
```
pdyraga, line 137: To be honest, I liked the previous name (`monitorAndAct`) more, especially that it's hidden inside the package and clearly expresses what we are doing in this function. I'd suggest going back to the previous name and renaming function paramters to clearly divide and express what is *monitor* and what is *act*. It was hard to me to figure out which function does what and put all the pieces together.

My proposition:
```
func (...) monitorAndAct(
  actionName string,
  monitoringStartFn watchDepositEventFn,
  monitoringStopFn watchDepositEventFn,
  keepClosed watchKeepClosedFn,
  actFn submitTxFn,
  timeout time.Duration,
)
```

pkg/applications/tbtc/tbtc.go
```diff
@@ -0,0 +1,303 @@
...
+					"keep closed event occurred for [%v] extension "+
+						"and deposit [%v]",
+					extensionName,
+					deposit,
+				)
+				break monitoring
+			case <-timeoutChan:
+				err := transactionSubmitter(deposit)
+				if err != nil {
+					if transactionAttempt == 3 {
```
pdyraga, line 206: Worth extracting it into a const.

pkg/applications/tbtc/tbtc.go
```diff
@@ -0,0 +1,303 @@
...
+	OnDepositCreated(
+		handler func(depositAddress string),
+	) (subscription.EventSubscription, error)
+
+	// OnDepositRegisteredPubkey installs a callback that is invoked when an
+	// on-chain notification of a deposit's pubkey registration is seen.
+	OnDepositRegisteredPubkey(
+		handler func(depositAddress string),
+	) (subscription.EventSubscription, error)
+}
```
pdyraga, line 53: What do you think about extracting them to `pkg/applications/tbtc/chain.go`?

pkg/applications/tbtc/tbtc.go
```diff
@@ -0,0 +1,300 @@
...
+	actionName string,
+	triggerEventHandlerSetup depositEventHandlerSetup,
+	stopEventHandlerSetup depositEventHandlerSetup,
+	keepClosedHandlerSetup keepClosedHandlerSetup,
+	transactionSubmitter depositTransactionSubmitter,
+	timeout time.Duration,
+) (subscription.EventSubscription, error) {
+	handleTriggerEvent := func(deposit string) {
+		logger.Infof(
+			"triggering [%v] action for deposit [%v]",
```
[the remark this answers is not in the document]
lukasz-zimnoch: I just changed this one to `triggering [%v] extension execution for deposit [%v]`. I think we should treat a partiular extension execution as a whole, not just focus on the action part.

pkg/chain/ethereum/connect.go
```diff
@@ -60,7 +59,7 @@ type EthereumChain struct {
 
 // Connect performs initialization for communication with Ethereum blockchain
 // based on provided config.
-func Connect(accountKey *keystore.Key, config *ethereum.Config) (eth.Handle, error) {
+func Connect(accountKey *keystore.Key, config *ethereum.Config) (*EthereumChain, error) {
```
pdyraga, line 62: Do we need to change it to `EhtereumChain` now?

configs/config.toml.SAMPLE
```diff
@@ -108,4 +108,9 @@
 # The port on which the `/diagnostics` endpoint will be available can be
 # customized below.
 # [Diagnostics]
-	# Port = 8081
\ No newline at end of file
+	# Port = 8081
+
+# Uncomment to enable application-specific extensions.
+# [ApplicationExtensions.TBTC]
```
pdyraga, line 114: Thoughts on making it just `Extensions.TBTC` for simplicity? Looking at the code, we have `InitializeExtensions`, `WithTBTCExtensions` so it seems like a natural thing to do.

pkg/chain/ethereum/connect.go
lukasz-zimnoch: First, we need `EthereumChain` in `start.go` when we wrap it with tbtc extensions. Also, there are no benefits to return an interface here, I'd even say it doesn't fit here. The caller invokes `ethereum.Connect` to obtain a concrete ethereum chain instance, not a generic chain type. So that change is also in accordance with the Go `accept interfaces return structs` principle.

cmd/start.go
```diff
@@ -183,6 +187,35 @@ func Start(c *cli.Context) error {
 	}
 }
 
+func initializeExtensions(
+	ctx context.Context,
+	config config.Extensions,
+	ethereumChain *ethereum.EthereumChain,
+) {
+	if len(config.TBTC.TBTCSystem) > 0 {
```
pdyraga, line 195: If `[Extensions.TBTC]` is not defined in the config file, is `config.TBTC` not `nil`? I wonder if we perhaps can reduce it to `config.TBTC != nil`.

pkg/extensions/tbtc/tbtc.go
```diff
@@ -0,0 +1,269 @@
...
+			)
+			return
+		}
+		defer stopEventSubscription.Unsubscribe()
+
+		keepClosedChan, keepClosedSubscriptionCancel, err := keepClosedFn(deposit)
+		if err != nil {
+			logger.Errorf(
+				"could not setup keep closed handler for [%v] "+
+					"extension execution and deposit [%v]: [%v]",
```
pdyraga, line 137: `could not setup keep closed handler for [%v] monitoring for deposit [%v]` perhaps?

pdyraga (commented):
Some comments on general naming and logs.

cmd/start.go
lukasz-zimnoch: It's a struct, not a pointer so it won't be `nil`.

pkg/extensions/tbtc/tbtc.go
```diff
@@ -0,0 +1,268 @@
...
+	"github.com/keep-network/keep-common/pkg/subscription"
+	eth "github.com/keep-network/keep-ecdsa/pkg/chain"
+)
+
+var logger = log.Logger("extensions-tbtc")
+
+const maxTransactionAttempts = 3
+
+// InitializeExtensions initializes extensions specific to the TBTC application.
+func InitializeExtensions(ctx context.Context, handle Handle) error {
```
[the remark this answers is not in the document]
lukasz-zimnoch: My understanding was the opposite 😀. But, your point seems to have more sense. I will stick to your proposition.

lukasz-zimnoch:
@pdyraga all comments addressed! I'm starting tests on my local environment but I don't expect any big changes in the code. Also, I think those TODOs listed in `pkg/extensions/tbtc/tbtc.go` should be addressed in separate PRs.

pkg/extensions/tbtc/tbtc.go
```diff
@@ -0,0 +1,286 @@
...
+			case <-keepClosedChan:
+				logger.Infof(
+					"keep closed event occurred for [%v] "+
+						"monitoring for deposit [%v]",
+					monitoringName,
+					deposit,
+				)
+				break monitoring
+			case <-timeoutChan:
+				err := actFn(deposit)
```
pdyraga, line 187: Should we log on info that we are going to act?

pkg/extensions/tbtc/tbtc.go
```diff
@@ -0,0 +1,286 @@
...
+				if err != nil {
+					if actionAttempt == maxActionAttempts {
+						logger.Errorf(
+							"could not perform action "+
+								"for [%v] monitoring for deposit [%v]: [%v]; "+
+								"last attempt failed",
```
pdyraga, line 193: Maybe `the maximum number of attempts reached`?

pkg/extensions/tbtc/tbtc.go
```diff
@@ -0,0 +1,286 @@
...
+
+					timeoutChan = time.After(backoff)
+					actionAttempt++
+				} else {
+					logger.Infof(
+						"action for [%v] monitoring for "+
+							"deposit [%v] has been performed successfully",
+						monitoringName,
+						deposit,
+					)
```
pdyraga, line 221: I would rather log that we are going to do something (my comment above) and then log error in case the action failed instead of logging this. We are not sure if the action succeeded - we submitted it to the chain, it's still waiting to be mined and there is still a small chance something can co wrong.

pkg/extensions/tbtc/tbtc.go
```diff
@@ -0,0 +1,286 @@
...
+					break monitoring
+				}
+			}
+		}
+
+		logger.Infof(
+			"[%v] monitoring for deposit [%v] "+
+				"has been completed",
```
pdyraga, line 229: I am not sure about `has been completed`, IMO it suggests it's a happy exit while we could reach the maximum number of retries. Let's maybe say `stopped [%v] monitoring for deposit [%v]`. WDYT?

pkg/extensions/tbtc/tbtc.go
```diff
@@ -0,0 +1,286 @@
...
+
+	unsubscribe := func() {
+		keepClosedSubscription.Unsubscribe()
+		keepTerminatedSubscription.Unsubscribe()
+	}
+
+	return signalChan, unsubscribe, nil
+}
+
+func exponentialBackoff(iteration int) time.Duration {
```
pdyraga, line 280: Can we add a comment explaining how the first backoffs will look like? My understanding is:
```
// 2000-2100ms,  4000-4100ms, 8000-8100ms,...
```

pdyraga (commented):
Some additional comments, mostly about naming, docs, and logs.

Reviewing tests now.

pkg/chain/local/local.go
```diff
@@ -53,25 +54,18 @@ func Connect() Chain {
...
-	lc.handlerMutex.Lock()
-	defer lc.handlerMutex.Unlock()
-
-	lc.keeps[keepAddress] = &localKeep{
-		members: members,
-	}
-	lc.keepAddresses = append(lc.keepAddresses, keepAddress)
+	// #nosec G104 (errors unhandled)
+	// No need to handle that error for now as it is triggered only
+	// in case the keep already exists.
```
pdyraga, line 59: So maybe `panic`?

pkg/extensions/tbtc/tbtc_test.go
```diff
@@ -0,0 +1,428 @@
...
+	keepPubkey, err := submitKeepPublicKey(depositAddress, tbtcChain)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// wait a while before triggering the stop event because the
+	// extension must have time to handle the start event
+	time.Sleep(100 * time.Millisecond)
+
+	err = tbtcChain.RetrieveSignerPubkey(depositAddress)
```
pdyraga, line 99: Let's add a comment that this call is a "stop piece" in this test. It took me a while to figure it out why we are calling it and not leaving it up to `monitorRetrievePubKey` 😆

pkg/extensions/tbtc/tbtc_test.go
```diff
@@ -0,0 +1,428 @@
...
+	expectedError := fmt.Errorf("no pubkey for deposit [%v]", depositAddress)
+	if !reflect.DeepEqual(expectedError, err) {
+		t.Errorf(
+			"unexpected error\n"+
+				"expected: [%v]\n"+
+				"actual:   [%v]",
+			expectedError,
+			err,
+		)
+	}
```
pdyraga, line 193: Should we check there is no `DepositPubkey` if we already asserted `RetrieveSigner` was not called? It looks to me that we are testing the logic of `local` chain implementation this way 🤔

pkg/chain/local/local.go
lukasz-zimnoch: I don't think so. I'd rather ignore as actually nothing happens if the keep aleady exists.

pdyraga: But it should never happen, right? I suggested `panic` as a kind of a guard condition for an invariant so that we can avoid hours of test debugging just to discover something in `local` is wrong and there are two keeps wih the same address. Maybe I don't have the full picture, though.

pkg/extensions/tbtc/tbtc_test.go
lukasz-zimnoch: I think having this check shows better what outcomes are we expecting in real world. I see that as follows:
1. Keep has been closed
2. The monitoring of retrieve pubkey has been stopped in result
3. Because the monitoring has been stopped, `RetrieveSignerPubkey` was not called
4. Because `RetrieveSignerPubkey` was not called, the deposit pubkey is not set

pkg/chain/local/local.go
lukasz-zimnoch: With current use, it can't happen. I see your point now and I think we could add the `panic` here for now.

pkg/extensions/tbtc/tbtc.go
```diff
@@ -0,0 +1,292 @@
...
+					"keep closed event occurred for [%v] "+
+						"monitoring for deposit [%v]",
+					monitoringName,
+					deposit,
+				)
+				break monitoring
+			case <-timeoutChan:
+				logger.Infof(
+					"timeout occurred for [%v] monitoring for "+
+						"deposit [%v]; performing action",
```
pdyraga, line 189: Nitpick: When I see `timeout occurred` in logs, I always assume something went wrong while in this case it's a happy path. I think we should avoid logging *timeout* here. Maybe:
```
`[%v] not performed by depositor in the expected time frame for deposit [%v], acting on their behalf`
```

lukasz-zimnoch: I will change the `not performed by depositor` to `not performed by externally` in case of other actions this will not be always a depositor I think.

pdyraga: Or let's just have `not performed` to not suggest anything. It's good enough.

pkg/extensions/tbtc/tbtc_test.go
pdyraga: Yeah, I see your point. From the perspective of defining the full expectations of how the system should behave it's indeed better.

pdyraga (approved):
🚀

PR #581, opened 2020-10-14
Fixing tests when future intervals are over

dimpar:
In this PR we are changing the way of keeps creation. All the tests were passing before, because we were in the interval 0. When the time comes for future intervals, some tests would've failed.

Refs https://github.com/keep-network/keep-ecdsa/issues/446

### DIFF
--- a/solidity/test/rewards/TestECDSARewards.js
+++ b/solidity/test/rewards/TestECDSARewards.js
@@ -63,7 +63,7 @@ describe("ECDSARewards", () => {
     it("should equal expected allocation in first interval", async () => {
       await keepFactory.stubBatchOpenFakeKeeps(100, firstIntervalStart)
 
-      await timeJumpToEndOfInterval(0)
+      await timeJumpToEndOfIntervalIfApplicable(0)
       await rewardsContract.allocateRewards(0)
       const allocated = await rewardsContract.getAllocatedRewards(0)
       const allocatedKeep = allocated.div(tokenDecimalMultiplier)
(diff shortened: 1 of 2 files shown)


solidity/test/rewards/TestECDSARewards.js
```diff
@@ -77,9 +77,20 @@ describe("ECDSARewards", () => {
...
+      const endOf = await rewardsContract.endOf(0)
+      let now = await time.latest()
+      // create 50 keeps in the second interval
+      if (now.lt(endOf)) {
+        await time.increaseTo(endOf.addn(1))
+        now = await time.latest()
+        await keepFactory.stubBatchOpenFakeKeeps(50, now)
+      } else {
+        await keepFactory.stubBatchOpenFakeKeeps(50, endOf.addn(1))
+      }
```
pdyraga, line 91: Can we simplify it to:
```
const firstIntervalEnd = await rewardsContract.endOf(0)
await keepFactory.stubBatchOpenFakeKeeps(50, firstIntervalEnd.addn(1))
```
?

solidity/test/rewards/TestECDSARewards.js
```diff
@@ -94,9 +105,17 @@ describe("ECDSARewards", () => {
...
-      await timeJumpToEndOfInterval(2)
+      const endOf = await rewardsContract.endOf(1)
+      let now = await time.latest()
+      if (now.lt(endOf)) {
+        await time.increaseTo(endOf.addn(1))
+        now = await time.latest()
+        await keepFactory.stubBatchOpenFakeKeeps(40, now)
+      } else {
+        await keepFactory.stubBatchOpenFakeKeeps(40, endOf.addn(1))
+      }
```
pdyraga, line 116: I think it can be simplified to:
```
const secondIntervalEnd = await rewardsContract.endOf(1)
await keepFactory.stubBatchOpenFakeKeeps(40, secondIntervalEnd.addn(1))
```

solidity/test/rewards/TestECDSARewards.js
dimpar: Right, interval 0 is already over.

solidity/test/rewards/TestECDSARewards.js
dimpar: Right, done

pdyraga (approved):

PR #584, opened 2020-10-20
Cap the maximum reward per ECDSA keep operator per interval

eth-r:
Refs #446

Set a limit on the amount of KEEP a single ECDSA keep operator can receive in a single reward interval.

Use the withdrawal pattern to track allocated and withdrawn rewards. Rewards exceeding the cap will be returned to the unallocated pool.

### DIFF
--- a/solidity/contracts/ECDSARewards.sol
+++ b/solidity/contracts/ECDSARewards.sol
@@ -123,8 +123,8 @@ contract ECDSARewards is Rewards {
     /// The rewards will be paid to the beneficiary
     /// of the operator requesting the withdrawal.
     /// @param interval The interval
-    function withdrawRewards(uint256 interval) external {
-        address beneficiary = tokenStaking.beneficiaryOf(msg.sender);
+    function withdrawRewards(uint256 interval, address operator) external {
+        address beneficiary = tokenStaking.beneficiaryOf(operator);
 
 
             uint256 allocatedForBeneficiary
--- a/solidity/contracts/ECDSARewards.sol
+++ b/solidity/contracts/ECDSARewards.sol
@@ -206,7 +206,7 @@ contract ECDSARewards is Rewards {
             if (newAllocation > beneficiaryRewardCap) {
                 uint256 deallocatedAmount = newAllocation.sub(beneficiaryRewardCap);
                 newAllocation = beneficiaryRewardCap;
-                sendToNewRewards(deallocatedAmount);
+                deallocate(deallocatedAmount);
             }
             allocatedRewards[beneficiary][interval] = newAllocation;
         }
--- a/solidity/contracts/ECDSARewards.sol
+++ b/solidity/contracts/ECDSARewards.sol
@@ -125,7 +125,10 @@ contract ECDSARewards is Rewards {
     /// @param operator The operator
     /// @return The amount allocated
     function getAllocatedRewards(uint256 interval, address operator)
-        external view returns (uint256) {
+        external
+        view
+        returns (uint256)
+    {
         address beneficiary = tokenStaking.beneficiaryOf(operator);
         return allocatedRewards[beneficiary][interval];
     }
@@ -136,7 +139,10 @@ contract ECDSARewards is Rewards {
     /// @param operator The operator
     /// @return The amount already withdrawn
     function getWithdrawnRewards(uint256 interval, address operator)
-        external view returns (uint256) {
+        external
+        view
+        returns (uint256)
+    {
         address beneficiary = tokenStaking.beneficiaryOf(operator);
         return withdrawnRewards[beneficiary][interval];
     }
@@ -147,7 +153,10 @@ contract ECDSARewards is Rewards {
     /// @param operator The operator
     /// @return The amount withdrawable
     function getWithdrawableRewards(uint256 interval, address operator)
-        external view returns (uint256) {
+        external
+        view
+        returns (uint256)
+    {
         address beneficiary = tokenStaking.beneficiaryOf(operator);
         uint256 allocated = allocatedRewards[beneficiary][interval];
         uint256 withdrawn = withdrawnRewards[beneficiary][interval];
--- a/solidity/contracts/ECDSARewards.sol
+++ b/solidity/contracts/ECDSARewards.sol
@@ -204,7 +204,7 @@ contract ECDSARewards is Rewards {
             if (newAllocation > operatorRewardCap) {
                 uint256 deallocatedAmount = newAllocation.sub(operatorRewardCap);
                 newAllocation = operatorRewardCap;
-                unallocatedRewards = unallocatedRewards.add(deallocatedAmount);
+                sendToNewRewards(deallocatedAmount);
             }
             allocatedRewards[member][interval] = newAllocation;
         }
--- a/solidity/contracts/ECDSARewards.sol
+++ b/solidity/contracts/ECDSARewards.sol
@@ -126,9 +126,8 @@ contract ECDSARewards is Rewards {
     function withdrawRewards(uint256 interval, address operator) external {
         address beneficiary = tokenStaking.beneficiaryOf(operator);
 
-
-            uint256 allocatedForBeneficiary
-         = allocatedRewards[beneficiary][interval];
+        uint256 allocatedForBeneficiary
+            = allocatedRewards[beneficiary][interval];
         uint256 alreadyWithdrawn = withdrawnRewards[beneficiary][interval];
 
         require(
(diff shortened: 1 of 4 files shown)


solidity/contracts/ECDSARewards.sol
```diff
@@ -158,16 +173,41 @@ contract ECDSARewards is Rewards {
...
+
+        for (uint256 i = 0; i < memberCount; i++) {
+            address member = members[i];
+            uint256 addedAllocation = allocations[i];
+            uint256 prevAllocated = allocatedRewards[member][interval];
+            uint256 newAllocation = prevAllocated.add(addedAllocation);
+            if (newAllocation > operatorRewardCap) {
+                uint256 deallocatedAmount = newAllocation.sub(operatorRewardCap);
+                newAllocation = operatorRewardCap;
+                unallocatedRewards = unallocatedRewards.add(deallocatedAmount);
```
pdyraga, line 207: We may have a little problem with upgrades. When finalizing the upgrade, we are allocating all past intervals and then, tokens are moved to the new contract. Rewards for past intervals are distributed later and we may find ourselves in a situtation when there is a lot of unallocated rewards on the previous rewards contract and there is an incetive to stay there.

What do you think? One option could be to disable any allocations after finalizing the upgrade and having a way to send all unallocated rewards to the new contract.

eth-r: Good catch, the unallocated rewards line should be using `sendToNewRewards` instead, which takes care of this correctly.

eth-r: Derp, I was looking at the wrong version of the rewards upgrade. We should do this yes:

> One option could be to disable any allocations after finalizing the upgrade and having a way to send all unallocated rewards to the new contract.

solidity/contracts/ECDSARewards.sol
```diff
@@ -88,8 +88,22 @@ contract ECDSARewards is Rewards {
     uint256 internal constant minimumECDSAKeepsPerInterval = 1000;
 
     BondedECDSAKeepFactory factory;
-
-    constructor(address _token, address payable _factoryAddress)
+    TokenStaking tokenStaking;
+
+    // TODO: set actual value
+    uint256 internal constant operatorRewardCap = 400000 * 10**18;
```
benlongstaff, line 94: should the cap be for the beneficiary address rather than the operator address? 

0xABF7335F324873Ccb1960C9cDb7278E627dd39eF
received the highest rewards out of all the beneficiaries (i believe 882,885 KEEP) by running 4 nodes and only staking 416,668 KEEP across them by juicing large BTC lot sizes. They would be unaffected by the change

0xBF62A4cf0e78b8799dcA0B481E4669a760C0f99F
will receive 669,583 KEEP from running one node, staking 9,293,734 KEEP and bonding 3,750 ETH and not juicing the keeps. they have the highest number of keeps still open. They would be affected by the change

solidity/contracts/ECDSARewards.sol
pdyraga: Addressed in https://github.com/keep-network/keep-core/pull/2136 and later in 3616caf58f750839b45bf9a268d12912ef9cba61.

solidity/contracts/ECDSARewards.sol
pdyraga: Addressed in https://github.com/keep-network/keep-ecdsa/pull/584/commits/35c47b5dfd6fdcc0572050b117b554899f95de84.

solidity/contracts/ECDSARewards.sol
```diff
@@ -100,6 +116,33 @@ contract ECDSARewards is Rewards {
...
         factory = BondedECDSAKeepFactory(_factoryAddress);
+        tokenStaking = TokenStaking(_tokenStakingAddress);
+    }
+
+    /// @notice Withdraw all available rewards for the given interval.
+    /// The rewards will be paid to the beneficiary
+    /// of the operator requesting the withdrawal.
+    /// @param interval The interval
+    function withdrawRewards(uint256 interval) external {
+        address beneficiary = tokenStaking.beneficiaryOf(msg.sender);
```
pdyraga, line 127: If we do it this way, only operator can withdraw. I think we should rather allow to pass the operator/beneficiary as a parameter.

solidity/contracts/ECDSARewards.sol
```diff
@@ -88,8 +88,24 @@ contract ECDSARewards is Rewards {
...
+    // The amount of tokens each individual beneficiary address
+    // can receive in a single interval is capped.
+    // TODO: set actual value
+    uint256 internal constant beneficiaryRewardCap = 400000 * 10**18;
+    // The total amount of rewards allocated to the given beneficiary address,
+    // in the given interval.
+    // `allocatedRewards[beneficiary][interval] -> amount`
+    mapping(address => mapping(uint256 => uint256)) allocatedRewards;
+    // The amount of interval rewards withdrawn to the given beneficiary.
+    mapping(address => mapping(uint256 => uint256)) withdrawnRewards;
```
pdyraga, line 102: Should we add two public functions allowing to fetch this information?

pdyraga:
I fixed a problem in tests in https://github.com/keep-network/keep-ecdsa/pull/584/commits/7dc61ccae4c373b5340dd448c1cea32cf8744f72, now they are failing as expected.

pdyraga:
... And added calls to `withdrawRewards` where applicable so they are now passing as expected.

solidity/contracts/ECDSARewards.sol
eth-r: Added, plus another one for fetching the withdrawable amount

solidity/contracts/ECDSARewards.sol
eth-r: Operator added as parameter; we can also switch to beneficiary if we like that better.

dimpar:
Code looks good to me, but tests will be added in a separate PR.

solidity/contracts/ECDSARewards.sol
pdyraga: I think the operator makes more sense 👌

solidity/contracts/ECDSARewards.sol
```diff
@@ -100,6 +116,67 @@ contract ECDSARewards is Rewards {
...
+    }
+
+    /// @notice Withdraw all available rewards for the given interval.
+    /// The rewards will be paid to the beneficiary of the specified operator.
+    /// @param interval The interval
+    /// @param operator The operator
+    function withdrawRewards(uint256 interval, address operator) external {
+        address beneficiary = tokenStaking.beneficiaryOf(operator);
+
+        uint256 allocatedForBeneficiary
```
pdyraga, line 164: We may consider renaming this variable to just `allocated` so that the linter is not getting crazy about tabulating the line.

pdyraga:
The last three comments and then I am ready to merge this one and https://github.com/keep-network/keep-ecdsa/pull/587 soon after.

solidity/contracts/ECDSARewards.sol
```diff
@@ -88,8 +88,24 @@ contract ECDSARewards is Rewards {
     uint256 internal constant minimumECDSAKeepsPerInterval = 1000;
 
     BondedECDSAKeepFactory factory;
+    TokenStaking tokenStaking;
 
-    constructor(address _token, address payable _factoryAddress)
+    // The amount of tokens each individual beneficiary address
+    // can receive in a single interval is capped.
+    // TODO: set actual value
+    uint256 internal constant beneficiaryRewardCap = 400000 * 10**18;
```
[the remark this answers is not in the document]
pdyraga: Well... we can actually leave it as-is for consistency. We'll verify this contract anyway.

solidity/contracts/ECDSARewards.sol
pdyraga: Done in https://github.com/keep-network/keep-ecdsa/pull/584/commits/873cb9a177301a3529a317d79127131c37a991de

solidity/contracts/ECDSARewards.sol
```diff
@@ -158,16 +235,44 @@ contract ECDSARewards is Rewards {
         return factory.getKeepOpenedTimestamp(toAddress(_keep)) != 0;
     }
 
+    // Get the members of the specified keep,
+    // and distribute the reward amount between them.
+    // The reward isn't paid out immediately,
+    // but is instead kept in the reward contract
+    // until each operator individually requests to withdraw the rewards.
```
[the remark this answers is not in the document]
pdyraga: Done in https://github.com/keep-network/keep-ecdsa/pull/584/commits/8da20610ac4291e61a20d53a98e4fcb13856eda7

pdyraga (approved):
Need approval from either @eth-r or @dimpar before the merge since I was also committing here.

dimpar (approved):
@pdyraga your last changes look good

PR #585, opened 2020-10-22
Redemption flow actions

lukasz-zimnoch:
Refs: #574 
Depends on: #580 
Depends on: https://github.com/keep-network/tbtc/pull/740

This pull request expands the TBTC extension by adding two new actions specific to the deposit redemption flow.
Here is a brief summary of both actions:

**`ProvideRedemptionSignature` action**
- Starts the monitoring on new `RedemptionRequested` event. That event is emitted either by [`requestRedemption`](https://github.com/keep-network/tbtc/blob/135f6c58e585033b8a48c8edc4448a323cbf0d04/solidity/contracts/deposit/DepositRedemption.sol#L81) or [`increaseRedemptionFee`](https://github.com/keep-network/tbtc/blob/135f6c58e585033b8a48c8edc4448a323cbf0d04/solidity/contracts/deposit/DepositRedemption.sol#L216) contract methods.
- Stops the monitoring if `GotRedemptionSignature` or `Redeemed` event occurs for the observed deposit. This means either [`provideRedemptionSignature`](https://github.com/keep-network/tbtc/blob/135f6c58e585033b8a48c8edc4448a323cbf0d04/solidity/contracts/deposit/DepositRedemption.sol#L172) or [`provideRedemptionProof`](https://github.com/keep-network/tbtc/blob/135f6c58e585033b8a48c8edc4448a323cbf0d04/solidity/contracts/deposit/DepositRedemption.sol#L294) has been called by someone else.
- Performs the `provideRedemptionSignature` action once the deposit is close to exceeding the [redemption signature timeout](https://github.com/keep-network/tbtc/blob/135f6c58e585033b8a48c8edc4448a323cbf0d04/solidity/contracts/system/TBTCConstants.sol#L14). 

**`ProvideRedemptionProof` action**
- Starts the monitoring on new `GotRedemptionSignature` event. This event occurs once `provideRedemptionSignature` contract method is called.
- Stops the monitoring if `RedemptionRequested` or `Redeemed` event occurs for the observed deposit. These events can be emitted by `increaseRedemptionFee` or `provideRedemptionProof` respectively.
- Performs the `increaseRedemptionFee` action once the deposit is close to exceeding the [redemption proof timeout](https://github.com/keep-network/tbtc/blob/135f6c58e585033b8a48c8edc4448a323cbf0d04/solidity/contracts/system/TBTCConstants.sol#L16). This part will be extended soon by adding the ability of submitting the redemption proof directy if the BTC transaction exists.


pkg/extensions/tbtc/events_log.go
```diff
@@ -0,0 +1,49 @@
+package tbtc
+
+import (
+	"sync"
+)
+
+type tbtcEventsLog struct {
```
pdyraga, line 7: We have a similar structure in `pkg/client/events_tracking.go` as well as in `keep-core`. Even though I like `eventsLog` name better, I think we should stick to `*Track`/`*_tracking` naming for structures and file for consistency. I think this structure is going to provide the same features as the one in `events_tracking.go` except that it also offers to get all events.

pkg/extensions/tbtc/tbtc.go
```diff
@@ -19,15 +21,15 @@ var logger = log.Logger("tbtc-extension")
...
 	logger.Infof("initializing tbtc extension")
 
-	tbtc := &tbtc{handle}
+	tbtc := newTBTC(chain)
 
 	err := tbtc.monitorRetrievePubKey(
 		ctx,
 		exponentialBackoff,
-		150*time.Minute,
+		150*time.Minute, // 30 minutes before the 3 hours on-chain timeout
```
pdyraga, line 32: Why not 15 minutes before just like for `monitorProvideRedemptionSignature`? `maxActionAttemps = 3` and we have backoff of 2, 4, and 8 seconds. We also have the transaction monitoring process increasing gas price for transaction in the pool.

pkg/chain/ethereum/ethereum.go
```diff
@@ -507,3 +510,48 @@ func (ec *EthereumChain) GetOpenedTimestamp(keepAddress common.Address) (time.Ti
...
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	result := make([]*eth.SignatureSubmittedEvent, 0)
+
+	hasNext := true
+	for hasNext {
+		hasNext = iterator.Next()
```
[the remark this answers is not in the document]
lukasz-zimnoch: `Next` function docs are not specific about the behavior for the first element. However, if we look at `FilterSignatureSubmitted`:
```
func (_BondedECDSAKeep *BondedECDSAKeepFilterer) FilterSignatureSubmitted(opts *bind.FilterOpts, digest [][32]byte) (*BondedECDSAKeepSignatureSubmittedIterator, error) {

	var digestRule []interface{}
	for _, digestItem := range digest {
		digestRule = append(digestRule, digestItem)
	}

	logs, sub, err := _BondedECDSAKeep.contract.FilterLogs(opts, "SignatureSubmitted", digestRule)
	if err != nil {
		return nil, err
	}
	return &BondedECDSAKeepSignatureSubmittedIterator{contract: _BondedECDSAKeep.contract, event: "SignatureSubmitted", logs: logs, sub: sub}, nil
}
```
We can see the `Event` pointer is not set to point to the first element by default. All mutations of the `Event` field are done by the `Next` method.

pkg/extensions/tbtc/tbtc.go
```diff
@@ -36,29 +38,73 @@ func Initialize(ctx context.Context, handle Handle) error {
...
 )
 
-type submitDepositTxFn func(deposit string) error
+type depositMonitoringContext struct {
+	startEvent depositEvent
+}
+
+type submitDepositTxFn func(*depositMonitoringContext) error
 
 type backoffFn func(iteration int) time.Duration
```
pdyraga, line 96: We have more and more abstractions added here and it's easy to lose track. Let's keep the definitions used in `monitorAndAct` right before `monitorAndAct`:
```
type watchDepositEventFn func(
	handler depositEventHandler,
) (subscription.EventSubscription, error)

type watchKeepClosedFn func(depositAddress string) (
	keepClosedChan chan struct{},
	unsubscribe func(),
	err error,
)

type submitDepositTxFn func(*depositMonitoringContext) error

type backoffFn func(iteration int) time.Duration

// TODO:
//  1. Filter incoming events by operator interest.
//  2. Incoming events deduplication.
//  3. Resume monitoring after client restart.
func (t *tbtc) monitorAndAct(
	ctx context.Context,
	monitoringName string,
	monitoringStartFn watchDepositEventFn,
	monitoringStopFn watchDepositEventFn,
	keepClosedFn watchKeepClosedFn,
	actFn submitDepositTxFn,
	actBackoffFn backoffFn,
	timeout time.Duration,
) (subscription.EventSubscription, error) {
```

pkg/extensions/tbtc/tbtc.go
```diff
@@ -36,29 +38,73 @@ func Initialize(ctx context.Context, handle Handle) error {
...
+}
+
+type depositRedemptionRequestedEvent struct {
+	*baseDepositEvent
+
+	digest      [32]uint8
+	blockNumber uint64
+}
+
+type depositEventHandler func(depositEvent)
```
pdyraga, line 78: I prefer the previous version: `type depositEventHandler func(depositAddress string)`. The additional indirection makes it harder to understand what's happening and the logic in app-specific actions is already quite complex.

pkg/extensions/tbtc/tbtc.go
```diff
@@ -36,29 +38,73 @@ func Initialize(ctx context.Context, handle Handle) error {
...
+func (bdd *baseDepositEvent) getDepositAddress() string {
+	return bdd.depositAddress
+}
+
+type depositRedemptionRequestedEvent struct {
+	*baseDepositEvent
+
+	digest      [32]uint8
+	blockNumber uint64
+}
```
pdyraga, line 76: I am not sure about it. I see your point but as a person who did not implement this code, it is hard for me to figure out what goes where. Maybe we can update the interface of `tbtcEventsLog`  to be similar to the one in `events_tracking.go` where we have a separate struct for each type of event and then, the function adding an event to this struct can accept any parameters that are needed? We then don't need to introduce additional abstractions but just operate on simple types.

lukasz-zimnoch: Those abstractions are not for `tbtcEventsLog`. Maybe it's not yet clear but this is just a side use. The most important use case is that we need some kind of context in the action function. For example, look at the provide redemption signature action. Once the timeout is exceeded, we have to provide the signature but to do that we need the digest and starting block from the event which started the monitoring action. I considered several possibilites to achieve that but this one is definitely the best and safest one despite it introduces additional complexity.

pkg/extensions/tbtc/tbtc.go
lukasz-zimnoch: So am I but I explained why we have to introduce them [here](https://github.com/keep-network/keep-ecdsa/pull/585#discussion_r510775802)

pkg/extensions/tbtc/tbtc.go
```diff
@@ -97,6 +149,144 @@ func (t *tbtc) monitorRetrievePubKey(
...
+					depositAddress,
+					v,
+					signatureSubmittedEvent.R,
+					signatureSubmittedEvent.S,
+				)
+			}
+		}
+
+		return fmt.Errorf(
+			"could not find signature for digest: [%v]",
```
pdyraga, line 260: Worth confirming at some point if the digest is logged in a human-readable form this way.

pkg/extensions/tbtc/events_log.go
lukasz-zimnoch: I got rid of the events log. It is not needed.

pkg/chain/ethereum/tbtc.go
```diff
@@ -3,10 +3,14 @@ package ethereum
 import (
 	"fmt"
 	"math/big"
+	"sort"
+
+	eth "github.com/keep-network/keep-ecdsa/pkg/chain"
```
pdyraga, line 8: Let's keep it as a `chain`.

pkg/chain/ethereum/tbtc.go
```diff
@@ -3,10 +3,14 @@ package ethereum
...
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/keep-network/keep-common/pkg/subscription"
 	"github.com/keep-network/tbtc/pkg/chain/ethereum/gen/contract"
+	"github.com/keep-network/tbtc/pkg/chain/ethereum/gen/filterer"
```
pdyraga, line 13: "filerer" in this context sounds weird IMO. It suggests we are filtering something from tBTC while we only do an event lookup. Looking at https://github.com/keep-network/tbtc/pull/740/, I'd probably go with "event log"

pkg/extensions/tbtc/tbtc.go
pdyraga: Even with the current version of the code where we got rid of the additional abstractions, I still opt for moving those functions to live next to `monitorAndAct`.

pkg/extensions/tbtc/tbtc.go
```diff
@@ -97,6 +134,270 @@ func (t *tbtc) monitorRetrievePubKey(
...
+	ctx context.Context,
+	actBackoffFn backoffFn,
+	timeout time.Duration,
+) error {
+	monitoringStartFn := func(
+		handler depositEventHandler,
+	) (subscription.EventSubscription, error) {
+		// Start right after a redemption has been requested or the redemption
+		// fee has been increased.
+		return t.chain.OnDepositRedemptionRequested(handler)
```
pdyraga, line 147: The fact the same event is logged by both `increaseRedemptionFee` and `_requestRedemption` is really helpful here. 🙏

pkg/extensions/tbtc/tbtc.go
```diff
@@ -97,6 +134,270 @@ func (t *tbtc) monitorRetrievePubKey(
...
+		if err != nil {
+			return nil, err
+		}
+
+		// Stop in case the redemption proof has been provided by someone else.
+		redeemedSubscription, err := t.chain.OnDepositRedeemed(handler)
+		if err != nil {
+			return nil, err
+		}
+
```
pdyraga, line 164: Hate to say that but I think we need to block this function here for some time (6-12 blocks) and check the chain before we signal the monitoring goroutine to stop. In case of a chain reorganizaion, we may get an event, stop monitoring, but the final chain state may not include the transaction with the redemption signature provided. In this case, our monitoring goroutine should act. I am not sure yet how it impacts the timeouts and exponential backoff for retries.

The same comment applies to all `monitoringStopFn` implementations.

If this suggestion does not lead to changing the code structure significantly - and I think it does not - I am fine with just leaving a TODO here and addressing this and all other `monitoringStopFn` implementations in a separate PR. Leaving the decision to you.

pdyraga, line 164: It is tricky but shouldn't we also close the current monitoring goroutine in case of `RedemptionRequested` event? It means that someone fee-bumped and we'll have another instance of this goroutine created.

Alternatively - and I am not sure if it is not a better solution - we shouldn't create another monitoring goroutine for `provideRedemptionSignature` in case there is already one working. With this approach we can address the scenario described above but also event duplicates.

pkg/extensions/tbtc/tbtc.go
```diff
@@ -97,6 +134,270 @@ func (t *tbtc) monitorRetrievePubKey(
...
+	actFn := func(depositAddress string) error {
+		keepAddress, err := t.chain.KeepAddress(depositAddress)
+		if err != nil {
+			return err
+		}
+
+		redemptionRequestedEvents, err := t.chain.PastDepositRedemptionRequestedEvents(
+			depositAddress,
+			0, // TODO: Should be something better that 0
```
pdyraga, line 181: This could make all clients deployed against Infura unhappy. As far as I know, they allow to reach no more than 1000 blocks in the past and assuming each block takes 15sec on average to mine, this is just 250 minutes back.

If there is no workaround, without seriously sacrificing code quality, we are fine. Just raising a concern.

pkg/extensions/tbtc/tbtc.go
```diff
@@ -97,6 +134,270 @@ func (t *tbtc) monitorRetrievePubKey(
...
+			latestRedemptionRequestedEvent.BlockNumber,
+		)
+		if err != nil {
+			return err
+		}
+
+		depositDigest := latestRedemptionRequestedEvent.Digest
+
+		for _, signatureSubmittedEvent := range signatureSubmittedEvents {
+			if bytes.Equal(signatureSubmittedEvent.Digest[:], depositDigest[:]) {
```
pdyraga, line 208: Wouldn't it be enough to take the last one, compare digest, and if it matches the expected one, submit? Othwerwise, we give up because we can't find the expected signature.

pkg/extensions/tbtc/tbtc.go
```diff
@@ -97,6 +134,270 @@ func (t *tbtc) monitorRetrievePubKey(
...
+	) (subscription.EventSubscription, error) {
+		// Start right after a redemption signature has been provided.
+		return t.chain.OnDepositGotRedemptionSignature(handler)
+	}
+
+	monitoringStopFn := func(
+		handler depositEventHandler,
+	) (subscription.EventSubscription, error) {
+		// Stop in case the redemption fee has been increased by someone else.
+		redemptionRequestedSubscription, err := t.chain.OnDepositRedemptionRequested(
```
pdyraga, line 272: Here we not only need to confirm the redemption fee increase TX has enough confirmations (see previous comment) but also if the one we are looking at happened after TX in which the redemption signature has been provided.

pkg/extensions/tbtc/tbtc.go
```diff
@@ -36,31 +39,65 @@ func Initialize(ctx context.Context, handle Handle) error {
...
 	unsubscribe func(),
 	err error,
 )
 
-type submitDepositTxFn func(deposit string) error
+type submitDepositTxFn func(depositAddress string) error
 
 type backoffFn func(iteration int) time.Duration
 
+type timeoutFn func(depositAddress string) (time.Duration, error)
```
pdyraga, line 89: Maybe `timeoutShiftFn`? This function is shifting the `timeout` parameter if needed.

pkg/extensions/tbtc/tbtc.go
lukasz-zimnoch: NVM, I've changed the approach so we have the previous version.

pkg/extensions/tbtc/tbtc.go
lukasz-zimnoch: NVM, I've changed the approach and got rid of the events log and those abstractions.

pkg/extensions/tbtc/tbtc.go
```diff
@@ -97,6 +149,144 @@ func (t *tbtc) monitorRetrievePubKey(
...
+					digest,
+					blockNumber,
+				}
+
+				// Log the event to make it available for future provide
+				// redemption proof monitoring process.
+				t.eventsLog.logDepositRedemptionRequestedEvent(
+					depositAddress,
+					event,
+				)
```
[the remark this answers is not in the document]
lukasz-zimnoch: Events log has been removed.

pkg/chain/ethereum/tbtc.go
lukasz-zimnoch: The `filterer` naming is borrowed from `go-ethereum`. I wanted to be consistent with the naming of contract backends generated by `abigen`. But agree, `event log` sounds much better.

pkg/extensions/tbtc/tbtc.go
lukasz-zimnoch: I'd vote for a separate PR. This one is already big enough.

lukasz-zimnoch: This goroutine works only when a deposit is in the `AWAITING_WITHFRAWAL_SIGNATURE` state. Once the signature is submitted this goroutine stops and the deposit switches to `AWAITING_WITHDRAWAL_PROOF` state. Fee bumps are only possible in the latter state.

pkg/extensions/tbtc/tbtc.go
lukasz-zimnoch: Unfortunately, the approach presented here is the safest and simplest one. We can cache incoming events as an alternative but given event duplications, chain reorgs and stuff, I want to avoid caching as it strongly complicates the code.

pkg/extensions/tbtc/tbtc.go
lukasz-zimnoch: I wanted to avoid relying on events order here. Maybe we could just start iterating from the end? This way we combine the advantages of both approaches.

pkg/extensions/tbtc/tbtc.go
lukasz-zimnoch: As I responded in the previous comment, I've added a TODO and I will handle this in a separate PR.

pkg/extensions/tbtc/tbtc.go
lukasz-zimnoch: It's shifting only for one case. For the rest, it returns a constant timeout. I look a this as a function that returns the timeout value so I think the `timeoutFn` covers more possible use cases.

lukasz-zimnoch:
@pdyraga added the unit tests and addressed all comments! I'm ready for the next look.

go.mod
```diff
@@ -20,7 +20,7 @@ require (
 	github.com/ipfs/go-log v1.0.4
 	github.com/keep-network/keep-common v1.2.1-0.20201020114759-19c123cbd4f4
 	github.com/keep-network/keep-core v1.3.0
-	github.com/keep-network/tbtc v1.1.1-0.20201020115551-5f9077c74826
+	github.com/keep-network/tbtc v1.1.1-0.20201026093513-cb9246987718
```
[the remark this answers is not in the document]
lukasz-zimnoch: Nope, we don't have to do anything in `keep-common`.

pkg/chain/chain.go
```diff
@@ -20,6 +20,8 @@ type Handle interface {
 	StakeMonitor() (chain.StakeMonitor, error)
 	// BlockCounter returns a block counter.
 	BlockCounter() chain.BlockCounter
+	// BlockTimestamp returns given block's timestamp.
+	BlockTimestamp(blockNumber *big.Int) (uint64, error)
```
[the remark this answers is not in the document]
lukasz-zimnoch: Yeah, that method should return an error. I'll add a comment.

go.mod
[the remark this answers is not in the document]
pdyraga: Don't we need generators for eventlog? From the description of the referenced PR:

> Added a TBTCSystemEventLog allowing to access past events emitted by the TBTCSystem contract. This is a temporary workaround because contract bindings generated by the keep-common generator don't support this feature yet.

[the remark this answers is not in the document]
lukasz-zimnoch: We just need the `abi` file which is already generated and tracked by the VCS.

pkg/chain/ethereum/connect.go
```diff
@@ -149,3 +152,19 @@ func addClientWrappers(
...
+		defer cancelCtx()
+
+		header, err := client.HeaderByNumber(ctx, blockNumber)
+		if err != nil {
+			return 0, err
+		}
+
+		return header.Time, nil
+	}
+}
```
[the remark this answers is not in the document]
lukasz-zimnoch: Agree but can we do all this work in a separate PR? I will add a TODO and work on it as my next thing. This work needs to introduce some changes in `keep-common` and `keep-core` as well and I really don't want to add more complexity to this PR.

pkg/chain/event.go
```diff
@@ -51,3 +62,19 @@ func (e *BondedECDSAKeepCreatedEvent) IsMember(address common.Address) bool {
 	}
 	return false
 }
+
+// DepositRedemptionRequestedEvent is an event emitted when a deposit
+// redemption has been requested or the redemption fee has been increased.
+// FIXME: This event should be placed in `extensions/tbtc` package but
+//  this cause an import cycle. For now, there are no better option
+//  than place it right here.
```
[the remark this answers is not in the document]
lukasz-zimnoch: Yeah sounds reasonable.

pkg/extensions/tbtc/tbtc.go
```diff
@@ -16,18 +19,21 @@ import (
 
 var logger = log.Logger("tbtc-extension")
 
-const maxActAttempts = 3
+const (
+	maxActAttempts           = 3
+	pastEventsLookbackBlocks = 10000
```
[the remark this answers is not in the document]
lukasz-zimnoch: This is possible but I'm afraid this can lead to some problems in the future. I'm not sure if we can safely rely on blocks numbers taken from events due to some weird things which may happen like event duplications or chain reorgs. Also, this will work for all actions but `provideRedemptionProof` which should look in the past beyond the monitoring start block. Taking those concerns in account, I think the safest and simplest way is just start from block `current - 10k`

pkg/extensions/tbtc/tbtc.go
```diff
@@ -97,10 +119,296 @@ func (t *tbtc) monitorRetrievePubKey(
...
+			latestRedemptionRequestedEvent.BlockNumber,
+		)
+		if err != nil {
+			return err
+		}
+
+		depositDigest := latestRedemptionRequestedEvent.Digest
+
+		// Start iterating from the latest event.
+		for i := len(signatureSubmittedEvents) - 1; i >= 0; i-- {
```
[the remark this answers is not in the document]
lukasz-zimnoch: Yep now we force the order. But, I'd leave the reversed loop just in case.

pkg/extensions/tbtc/tbtc.go
lukasz-zimnoch: I've added the link to this discussion to the [TODOs section](https://github.com/keep-network/keep-ecdsa/pull/585#discussion_r513447505) where I listed all the things related to events deduplication, reorgs etc.

pkg/chain/ethereum/connect.go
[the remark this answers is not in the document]
pdyraga: I'll leave this discussion open here until https://github.com/keep-network/keep-common/pull/55 is reviewed and merged.

pkg/extensions/tbtc/tbtc.go
pdyraga: Handled in https://github.com/keep-network/keep-ecdsa/pull/591 but I'll this discussion open here until that PR is reviewed and merged.

pkg/extensions/tbtc/tbtc.go
pdyraga: Is this one already addressed in one of follow-up PRs?

pkg/extensions/tbtc/tbtc.go
pdyraga: Leaving this discussion open until all follow-up PRs addressing those issues are reviewed and merged.

pkg/extensions/tbtc/tbtc.go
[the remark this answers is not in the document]
pdyraga: I think it's misleading. If we are leaving a loop here because there is an edge case where we can not just take the last event, we should also do something with `redemptionRequestedEvents`. Happy to be proven wrong but I do not think there is such an edge case for tBTC extension.

go.mod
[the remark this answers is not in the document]
pdyraga: We do need to implement a change to the generator in `keep-common` so that it generates `TBTCSystemEventLog`, correct?

Let me put it this way:

We build tBTC go package from scratch:
- we remove all generated files from ` pkg/chain/ethereum/gen`
- we do `go generate ./...` there
- we push a new version

Then we build `keep-ecdsa`:
- we update a dependency to `tbtc`
- we run `go build`

My understanding is that the compilation will fail because `TBTCSystemEventLog` will not be there.

In other words, we are depending on a file that *should be* generated but it's not generated.

pkg/extensions/tbtc/tbtc.go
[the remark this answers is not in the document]
pdyraga: We can leave it like that for now. A little concerned about how much it costs to scan 10k previous blocks for operators using 3rd party Ethereum providers who charge by compute unit but we can revise in the future.

pkg/extensions/tbtc/tbtc_test.go
```diff
@@ -367,27 +391,1294 @@ func TestRetrievePubkey_ContextCancelled_WithWorkingMonitoring(t *testing.T) {
...
+	)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// wait a bit longer than the monitoring timeout
+	// to make sure the potential transaction completes
+	time.Sleep(2 * timeout)
+
+	expectedIncreaseRedemptionFeeCalls := 1
```
pdyraga, line 1095: Could be worth adding an explanatory comment that this call comes from `IncreaseRedemptionFee` in the test and the monitoring script is not expected to do this call. At least that's my understanding of this test.

pkg/chain/local/local.go
```diff
@@ -32,6 +36,10 @@ type Chain interface {
 type localChain struct {
 	handlerMutex sync.Mutex
 
+	blockCounter          chain.BlockCounter
+	blocksTimestamps      map[uint64]uint64
+	blocksTimestampsMutex sync.RWMutex
```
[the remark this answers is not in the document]
pdyraga: (please look at [this comment](https://github.com/keep-network/keep-ecdsa/pull/585#discussion_r518715714) before making a decision here)

pkg/extensions/tbtc/tbtc.go
lukasz-zimnoch: It's addressed in https://github.com/keep-network/keep-ecdsa/pull/592 where we check the status of deposit after `12` blocks.

go.mod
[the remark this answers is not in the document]
lukasz-zimnoch: Not quite. Eventually, we should extend the existing `keep-common` generator in that way it generates the `Filter*` methods straight into the `TBTCSystem` binding struct. `TBTCSystemEventLog` is just a temporary workaround created manually without using the generator. It is also tracked by VCS similarly as the regular bindings. So, when we build tBTC package from scratch we shouldn't delete anything. We should only run `go generate ./...` to capture possible contract changes.

pkg/extensions/tbtc/tbtc.go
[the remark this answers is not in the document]
lukasz-zimnoch: IIRC, it's always one `eth_getLogs` call if we are talking about the number of requests.

go.mod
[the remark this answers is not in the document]
pdyraga: Right. So just to be clear - I am absolutely fine with merging this PR having `TBTCSystemEventLog` created manually just like now but I do not want to release a new version before we have this file generated or having `Filter*` methods generated into `TBTCSystem` struct.

pkg/extensions/tbtc/tbtc.go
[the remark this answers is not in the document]
lukasz-zimnoch: Ok, I'll remove that loop.

pkg/chain/local/local.go
```diff
@@ -145,6 +179,9 @@ func (lc *localChain) SubmitKeepPublicKey(
 	keepAddress common.Address,
 	publicKey [64]byte,
 ) error {
+	lc.handlerMutex.Lock()
+	defer lc.handlerMutex.Unlock()
```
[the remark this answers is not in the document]
lukasz-zimnoch: Ok sure. But, I can't rename it to `mutex` because there is also a `mutex` instance in the `TBTCLocalChain` which extends the `localChain`. I will do the following:
- rename `handlerMutex` in `localChain` to `localChainMutex`
- rename `mutex` in `TBTCLocalChain` to `tbtcLocalChainMutex`

pkg/chain/local/local.go
```diff
@@ -45,12 +53,38 @@ type localChain struct {
...
+	localChain := &localChain{
+		blockCounter:        blockCounter,
+		blocksTimestamps:    map[uint64]uint64{0: uint64(time.Now().Unix())},
 		keeps:               make(map[common.Address]*localKeep),
 		keepCreatedHandlers: make(map[int]func(event *eth.BondedECDSAKeepCreatedEvent)),
 		clientAddress:       common.HexToAddress("6299496199d99941193Fdd2d717ef585F431eA05"),
 		authorizations:      make(map[common.Address]bool),
 	}
+
+	go localChain.observeBlocksTimestamps(context.Background())
```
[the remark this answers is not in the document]
lukasz-zimnoch: > I think we are not clearing up properly this way and we'll have this goroutine running even after tests are done.

I'll pass a context here but as I observed the `go test` command kills the process once tests are done. Am I missing something?

pkg/extensions/tbtc/tbtc.go
```diff
@@ -36,29 +42,45 @@ func Initialize(ctx context.Context, handle Handle) error {
...
 type tbtc struct {
-	chain Handle
+	chain chain.TBTCHandle
+}
+
+func newTBTC(chain chain.TBTCHandle) *tbtc {
+	return &tbtc{
+		chain: chain,
+	}
 }
```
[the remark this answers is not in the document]
lukasz-zimnoch: I'd rather vote to keep the constructor close to the struct it creates. And it's not used in tests only but also in `Initialize` function.

pkg/extensions/tbtc/tbtc_test.go
lukasz-zimnoch: But we are expecting exactly one call in this test.

pkg/extensions/tbtc/tbtc.go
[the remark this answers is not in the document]
pdyraga: Yeah, though it sometimes does not matter if it's just one call but how expensive was to compute a respose for this call.

pkg/chain/local/local.go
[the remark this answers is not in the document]
pdyraga: Having the mutex for entire struct we could leave this map as `map[uint64]uint64` but that's fine.

go.mod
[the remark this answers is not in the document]
pdyraga: I'll leave this discussion open until the proper solution is implemented. I do not want to block this PR, though.

pkg/chain/local/local.go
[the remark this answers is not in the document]
pdyraga: Correct, goroutines should not be dangling forever but given that a new instance of local chain is created for every single test, I think it's worth cleaning up.

pkg/extensions/tbtc/tbtc_test.go
pdyraga: Right, and my understanding is that this single call comes from `tbtcChain.IncreaseRedemptionFee` called at line 1116 in this test. I am not saying we should expect more than one call to be recorded here. It just took me a while to understand where this single call comes from so I think it makes sense to leave a comment for future-us that this call does not come from the monitoring goroutine but from this test. Unless I missed something?

pdyraga: Non-blocking of course, we can address it in any other PR.

pdyraga (approved):

PR #587, opened 2020-10-27
Rewards cap tests

dimpar:
Depends on https://github.com/keep-network/keep-ecdsa/pull/584
Refs #446

Adding tests around the maximum cap rewards per ECDSA keep operator.

### DIFF
--- a/solidity/contracts/test/ECDSARewardsStub.sol
+++ b/solidity/contracts/test/ECDSARewardsStub.sol
@@ -19,9 +19,4 @@ contract ECDSARewardsStub is ECDSARewards {
     {
         beneficiaryRewardCap = 10000 * 10**18;
     }
-
-    function getBeneficiaryRewardCap() public view returns (uint256) {
-        return beneficiaryRewardCap;
-    }
-
 }
--- a/solidity/contracts/test/ECDSARewardsStub.sol
+++ b/solidity/contracts/test/ECDSARewardsStub.sol
@@ -1,0 +1,27 @@
+pragma solidity 0.5.17;
+
+import "../../contracts/ECDSARewards.sol";
+
+/// @title ECDSA Rewards Stub for ecdsa rewards testing
+/// @dev This contract is for testing purposes only.
+contract ECDSARewardsStub is ECDSARewards {
+    constructor(
+        address _token,
+        address payable _factoryAddress,
+        address _tokenStakingAddress
+    )
+        public
+        ECDSARewards(
+            _token,
+            _factoryAddress,
+            _tokenStakingAddress
+        )
+    {
+        beneficiaryRewardCap = 10000 * 10**18;
+    }
+
+    function getBeneficiaryRewardCap() public view returns (uint256) {
+        return beneficiaryRewardCap;
+    }
+
+}
--- a/solidity/contracts/test/ECDSARewardsStub.sol
+++ b/solidity/contracts/test/ECDSARewardsStub.sol
@@ -5,6 +5,9 @@ import "../../contracts/ECDSARewards.sol";
 /// @title ECDSA Rewards Stub for ecdsa rewards testing
 /// @dev This contract is for testing purposes only.
 contract ECDSARewardsStub is ECDSARewards {
+
+    uint256 internal tokenDecimalMultiplier = 10**18;
+
     constructor(
         address _token,
         address payable _factoryAddress,
@@ -16,7 +19,9 @@ contract ECDSARewardsStub is ECDSARewards {
             _factoryAddress,
             _tokenStakingAddress
         )
-    {
-        beneficiaryRewardCap = 10000 * 10**18;
+    {}
+
+    function setBeneficiaryRewardCap(uint256 _beneficiaryRewardCap) public {
+        beneficiaryRewardCap = _beneficiaryRewardCap.mul(tokenDecimalMultiplier);
     }
 }
--- a/solidity/test/rewards/TestECDSARewards.js
+++ b/solidity/test/rewards/TestECDSARewards.js
@@ -272,11 +272,11 @@ describe("ECDSARewards", () => {
       await timeJumpToEndOfIntervalIfApplicable(0)
 
       // mark first keep as closed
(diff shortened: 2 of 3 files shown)


solidity/test/rewards/TestECDSARewards.js
```diff
@@ -230,10 +230,130 @@ describe("ECDSARewards", () => {
...
+
+      for (let i = 0; i < operators.length; i++) {
+        await expectRevert(
+          rewardsContract.withdrawRewards(0, operators[i], {
+            from: operators[i],
+          }),
+          "No rewards to withdraw"
+        )
+      }
+    })
```
pdyraga, line 264: It's worth adding yet one test that will cover `receiveReward + withdrawRewards + receiveRewards + withdrawRewards` path for the same interval.

solidity/test/rewards/TestECDSARewards.js
```diff
@@ -230,10 +230,130 @@ describe("ECDSARewards", () => {
...
+
+      for (let i = 0; i < operators.length; i++) {
+        await rewardsContract.withdrawRewards(0, operators[i], {
+          from: operators[i],
+        })
+      }
+
+      await assertAllocatedRewards(expectedBalanceFromTwoKeeps, 0, operators)
+      await assertWithdrawableRewards(new BN(0), 0, operators)
+      await assertWithdrawnRewards(expectedBalanceFromTwoKeeps, 0, operators)
```
pdyraga, line 308: I think it's time to reorganize tests a little and separate allocation from withdrawal:
```
    rewards allocation
      should not be possible when a keep is not closed 
      should not be possible when a keep is terminated
      should exclude terminated groups
      should cap the maximum reward per beneficiary
    rewards withdrawal
      should correctly distribute rewards between beneficiaries
      should revert when rewards have been withdrawn already 
      should correctly receive rewards from multiple keeps
      should correctly track withdrawable and withdrawn amount
```

solidity/test/rewards/TestECDSARewards.js
dimpar: I've added for: `receiveReward + withdrawRewards + receiveRewards` since the second `receiveRewards` will trigger revert: `Rewards already claimed`

solidity/test/rewards/TestECDSARewards.js
dimpar: Yes, agree. Reorganized :)

solidity/contracts/test/ECDSARewardsStub.sol
```diff
@@ -0,0 +1,27 @@
...
+            _tokenStakingAddress
+        )
+    {
+        beneficiaryRewardCap = 10000 * 10**18;
+    }
+
+    function getBeneficiaryRewardCap() public view returns (uint256) {
+        return beneficiaryRewardCap;
+    }
+
```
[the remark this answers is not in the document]
dimpar: Right, removed

solidity/test/rewards/TestECDSARewards.js
pdyraga: That's very good but what I suggested is yet another scenario. The scenario I have in mind is when you claim even more rewards by executing the second `receiveReward` for another keep. Real-world example is when that second keep gets closed later, once you have already received and withdrawn reward for the first keep.

dimpar: Okay, added in 09453fd

solidity/contracts/test/ECDSARewardsStub.sol
```diff
@@ -0,0 +1,22 @@
...
+    )
+        public
+        ECDSARewards(
+            _token,
+            _factoryAddress,
+            _tokenStakingAddress
+        )
+    {
+        beneficiaryRewardCap = 10000 * 10**18;
+    }
```
pdyraga, line 21: Wouldn't it be cleaner to provide a setter on the stub that would be called with a value provided by test?

dimpar: Hm, I don't know if it is cleaner, because it is a "constant" value that should probably not be changed in tests just like in a real `ECDSARewards` contract.  I believe we have a similar example in `TokenStakingStub` where we declare `minimumStake` without a setter. 
But if you think we should have a setter, I can certainly add it.

solidity/test/rewards/TestECDSARewards.js
```diff
@@ -268,11 +342,88 @@ describe("ECDSARewards", () => {
...
+      await assertWithdrawnRewards(new BN(0), 0, operators)
+
+      await withdrawRewards(0, operators)
+
+      await assertWithdrawableRewards(new BN(0), 0, operators)
+      await assertWithdrawnRewards(expectedBalanceFromTwoKeeps, 0, operators)
+    })
   })
 
   async function assertKeepBalanceOfBeneficiaries(expectedBalance) {
```
pdyraga, line 424: Can we add `operators` as a parameter here? Just to stay consistent with the rest of `assert*` functions.

solidity/contracts/test/ECDSARewardsStub.sol
pdyraga: I mean... we update this value for tests and we have a good reason for doing that. I think we should either have a setter and set the value on a stub from tests or don't have a stub at all and test against the real value. Latter is not possible with the test harness we have so I'd go with option one.

pdyraga:
Two last comments and I am good here.

solidity/test/rewards/TestECDSARewards.js
dimpar: Hm, `operators` are actually visible by all the functions in tests which is set in `before`. Maybe let's do the other way around and remove this parameter from `assert*` functions?

pdyraga: Yeah, we can do that.

solidity/contracts/test/ECDSARewardsStub.sol
dimpar: Done

pdyraga (approved):

PR #589, opened 2020-10-30
Unregister fully backed keep's members on termination

nkuba:
The owner can seize signer bonds. This happens when keep members misbehave or are inactive. In such a case, a keep gets terminated. 
To prevent a situation that inactive or misbehaving operators are selected once again to a new keep we unregister them from any pools associated with a keep factory.
We expect the operators to re-register to if they are working correctly or once they solve their problems.

Depends on: https://github.com/keep-network/sortition-pools/pull/98
Depends on: https://github.com/keep-network/keep-ecdsa/pull/569

### DIFF
--- a/cmd/signing.go
+++ b/cmd/signing.go
@@ -1,0 +1,287 @@
+package cmd
+
+import (
+	"context"
+	"encoding/hex"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"sync"
+	"time"
+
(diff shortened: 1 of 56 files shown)


pdyraga:
We decided to not implement #271 natively in the protocol but instead introduce a KEEP-only pool. At the same time, there are efforts in Keep community to implement community staking pools.

PR #590, opened 2020-11-02
Filter TBTC monitoring by operator interest

lukasz-zimnoch:
Refs: #574
Depends on: #585

All deposit monitoring events within the TBTC extension should be filtered by operator interest. Specifically, an operator should check its membership in the underlying keep before starting the monitoring routine in response to an incoming deposit event.

### DIFF
--- a/pkg/chain/local/tbtc.go
+++ b/pkg/chain/local/tbtc.go
@@ -102,16 +102,15 @@ func NewTBTCLocalChain(ctx context.Context) *TBTCLocalChain {
 	}
 }
 
-func (tlc *TBTCLocalChain) CreateDeposit(depositAddress string) {
+func (tlc *TBTCLocalChain) CreateDeposit(
+	depositAddress string,
+	signers []common.Address,
+) {
 	tlc.tbtcLocalChainMutex.Lock()
 	defer tlc.tbtcLocalChainMutex.Unlock()
 
 	keepAddress := generateAddress()
-	tlc.OpenKeep(keepAddress, []common.Address{
-		generateAddress(),
-		generateAddress(),
-		generateAddress(),
-	})
+	tlc.OpenKeep(keepAddress, signers)
 
 	tlc.deposits[depositAddress] = &localDeposit{
 		keepAddress:               keepAddress.Hex(),
@@ -625,6 +624,16 @@ func (tlc *TBTCLocalChain) SetAlwaysFailingTransactions(transactions ...string) 
 
 func (tlc *TBTCLocalChain) Logger() *localChainLogger {
 	return tlc.logger
+}
+
+func (tlc *TBTCLocalChain) RandomSigningGroup(size int) []common.Address {
+	signers := make([]common.Address, size)
+
+	for i := range signers {
+		signers[i] = generateAddress()
+	}
+
+	return signers
 }
 
 func fromLittleEndianBytes(bytes [8]byte) *big.Int {
--- a/pkg/chain/local/tbtc.go
+++ b/pkg/chain/local/tbtc.go
@@ -2,6 +2,7 @@ package local
 
 import (
 	"bytes"
+	"context"
 	"encoding/binary"
 	"fmt"
 	"math/big"
@@ -73,7 +74,7 @@ func (lcl *localChainLogger) IncreaseRedemptionFeeCalls() int {
 type TBTCLocalChain struct {
 	*localChain
 
-	mutex sync.Mutex
+	tbtcLocalChainMutex sync.Mutex
 
 	logger *localChainLogger
 
@@ -87,9 +88,9 @@ type TBTCLocalChain struct {
 	depositRedeemedHandlers               map[int]func(depositAddress string)
 }
 
-func NewTBTCLocalChain() *TBTCLocalChain {
+func NewTBTCLocalChain(ctx context.Context) *TBTCLocalChain {
 	return &TBTCLocalChain{
-		localChain:                            Connect().(*localChain),
+		localChain:                            Connect(ctx).(*localChain),
 		logger:                                &localChainLogger{},
 		alwaysFailingTransactions:             make(map[string]bool),
 		deposits:                              make(map[string]*localDeposit),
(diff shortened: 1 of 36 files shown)


pkg/extensions/tbtc/tbtc.go
```diff
@@ -420,6 +435,10 @@ func (t *tbtc) monitorAndAct(
 	timeoutFn timeoutFn,
 ) (subscription.EventSubscription, error) {
 	handleStartEvent := func(depositAddress string) {
+		if !monitoringFilterFn(depositAddress) {
```
pdyraga, line 438: Maybe `shouldMonitorFn`? I see `monitoringFilterFn` is more consistent with other `monitoring*` functions but `!monitoringFilterFn(...)` reads weird. My first thought was that this is the case for a situation when this function is not defined.

pdyraga: Also... why do we pass it as a parameter, can't we just call `shouldMonitorDeposit` directly? It's always the same function that is passed here.

pkg/extensions/tbtc/tbtc_test.go
```diff
@@ -1718,3 +1718,20 @@ func terminateKeep(
 func constantBackoff(_ int) time.Duration {
 	return time.Millisecond
 }
```
pdyraga, line 1720: Can we add at least one test that proves monitoring is not triggered for deposit the client is not a signer of?

pkg/extensions/tbtc/tbtc.go
lukasz-zimnoch: Ok, I'll rename. However, I'd vote to leave it as a parameter. It's a more flexible approach and fits better with the generic spirit of the `monitorAndAct` function.

pkg/extensions/tbtc/tbtc.go
```diff
@@ -604,6 +616,28 @@ func (t *tbtc) watchKeepClosed(
...
+		return false
+	}
+
+	for _, member := range members {
+		if member == t.chain.Address() {
+			return true
+		}
+	}
+
+	return false
```
pdyraga, line 638: This function executes two calls for every state transition we are monitoring. It means we add 3x2 calls for every new deposit now and we'll have 4x2 calls once we implement provide redemption proof. So 8 calls in total for every new deposit in the system. It's not the end of the world, but it can increase operational costs.

I think we have two options to improve it: 
- use `Registry` for all monitorings but retrieve pubkey,
- introduce time cache in a similar fashion as in the firewall.

I like the second idea better because we do not depend on the order of events and when keep is added to the registry. No logic is leaking from other modules.

For the second option, we would have two time caches - one for deposits we are tracking and another for deposits we are not tracking. Two calls to the chain would be executed for every new deposit but only one time.

pkg/extensions/tbtc/tbtc.go
```diff
@@ -391,6 +397,8 @@ func (t *tbtc) monitorProvideRedemptionProof(
 	return nil
 }
 
+type depositFilterFn func(depositAddress string) bool
```
pdyraga, line 400: Maybe `shouldMonitorDepositFn`? We have `submitDepositTxFn` so it's aligned.

pkg/extensions/tbtc/tbtc.go
```diff
@@ -604,6 +616,28 @@ func (t *tbtc) watchKeepClosed(
 	return signalChan, unsubscribe, nil
 }
 
+func (t *tbtc) shouldMonitorDeposit(
+	depositAddress string,
+) bool {
```
pdyraga, line 621: I think this one as well as `Initialize` in this file and `CreateDeposit` in `pkg/chain/local/tbtc.go` should fit in one line. We don't need to break them.

pkg/chain/local/tbtc.go
```diff
@@ -627,6 +626,16 @@ func (tlc *TBTCLocalChain) Logger() *localChainLogger {
...
+func (tlc *TBTCLocalChain) RandomSigningGroup(size int) []common.Address {
+	signers := make([]common.Address, size)
+
+	for i := range signers {
+		signers[i] = generateAddress()
+	}
+
+	return signers
+}
+
```
pdyraga, line 638: Thoughts on keeping it in `local.go` next to  `generateAddress` and not as a function of `TBTCLocalChain`?

pkg/extensions/tbtc/tbtc.go
lukasz-zimnoch: I'm not sure a time cache will improve the situation. The major drawback is the fact we don't know when the deposit state will change. It may take months until the deposit redempton will be started. So, we can't determine the optimal caching timespan. Also it's not true we will have 4x2 once we implement proofs. Proofs will be in the same monitoring routine as increase redemption fee.

pkg/extensions/tbtc/tbtc.go
lukasz-zimnoch: Agree with the first two but `CreateDeposit` doesn't fit.

pkg/extensions/tbtc/tbtc.go
pdyraga: Correct. Time cache would not work between deposit and redemption but we can reduce the number of calls during them separately. 

Looking at the next PR, #591, in `monitorAndAct` we do:
- `shouldMonitorFn`?
- `acquireMonitoringLock`

In this setup, we'll execute keep memebrs check for every deposit state change in the system. And we'll do it not only one time but for every event duplicate. 

One option could be to swap those calls, but then we'll be still executing keep members check for:
- all deposits we are not a part of, for which event duplicates occured and we already released the monitoring lock,
- all increase fee events for the deposit we are already monitoring.

We can set the lifetime of this cache to the maximum time of redemption with 3 x increased fee (because why not).

lukasz-zimnoch: Ok. Can we do it in a separate PR to not block here? I'm currently entering the optimization land so it's a good candidate.

pdyraga: Sure, I'll just keep this thread open for now.

pdyraga (approved):
We decided to look into the [optimizations with a time cache](https://github.com/keep-network/keep-ecdsa/pull/590#pullrequestreview-527308719) in a separate PR.

pkg/extensions/tbtc/tbtc.go
lukasz-zimnoch: Addressed in the follow-up PR: https://github.com/keep-network/keep-ecdsa/pull/600

PR #591, opened 2020-11-02
Prevent TBTC monitoring duplication

lukasz-zimnoch:
Refs: #574
Depends on: #590 

Implemented a monitoring lock mechanism that should prevent triggering duplicated monitoring instances. Those redundant instances can be started as a response to duplicated events that are sometimes delivered by third-party Ethereum nodes.


pkg/extensions/tbtc/monitoring_lock.go
```diff
@@ -0,0 +1,55 @@
...
+	monitoringName string,
+) monitoringLockKey {
+	return monitoringLockKey(fmt.Sprintf(
+		"%v-%v",
+		depositAddress,
+		strings.ReplaceAll(monitoringName, " ", ""),
+	))
+}
+
+type monitoringLock struct {
```
Shadowfiend, line 22: Could thus be better served by directly using [`sync.Map`](https://golang.org/pkg/sync/#Map 
)?

lukasz-zimnoch: Yeah, good idea!

pdyraga:
> duplicated events that are sometimes delivered by third-party Ethereum nodes.

Heads up: even if the operator is not using Infura or Alchemy, and they are running their own Ethereum client, duplicate events are still happening for small chain re-orgs.

pkg/extensions/tbtc/tbtc.go
```diff
@@ -439,6 +440,17 @@ func (t *tbtc) monitorAndAct(
 			return
 		}
 
+		if !t.monitoringLock.tryLock(depositAddress, monitoringName) {
+			logger.Warningf(
+				"lock for [%v] monitoring for deposit [%v] has been "+
+					"already acquired; could not start the monitoring",
```
pdyraga, line 446: I'd rather say the given deposit is already monitored for this event. If I were reading logs and saw a message about already acquired lock and monitoring that could not be started because of that, I'd think it's something wrong with synchronization. While here, it's just a normal thing with duplicate events.

pkg/extensions/tbtc/tbtc.go
```diff
@@ -439,6 +440,17 @@ func (t *tbtc) monitorAndAct(
 			return
 		}
 
+		if !t.monitoringLock.tryLock(depositAddress, monitoringName) {
```
pdyraga, line 443: Can we cover this check in `tbtc_test.go`?

pkg/extensions/tbtc/monitoring_lock.go
```diff
@@ -0,0 +1,41 @@
...
+	locks sync.Map
```
pdyraga, line 23: Wouldn't it be enough to have `monitoringLocks sync.Map` in `type tbtc struct` and have `tryLock` and `releaseLock` defined on this struct? We'd be able to have exactly the same tests as in `monitoring_lock_test.go`, just one level of indirection less.

pkg/extensions/tbtc/tbtc.go
```diff
@@ -437,6 +439,17 @@ func (t *tbtc) monitorAndAct(
 			return
 		}
 
+		if !t.acquireMonitoringLock(depositAddress, monitoringName) {
+			logger.Warningf(
+				"[%v] monitoring for deposit [%v] is already running; "+
+					"could not start new monitoring instance",
```
pdyraga, line 445: Maybe just the first part `[%v] monitoring for deposit [%v] is already running`? We do not even bother to start new monitoring because it's already running so it's not that we *could not* start it. 😉

pkg/extensions/tbtc/tbtc_test.go
```diff
@@ -1851,6 +1853,153 @@ func TestProvideRedemptionProof_OperatorNotInSigningGroup(
...
+		return subscription.NewEventSubscription(func() {}), nil
+	}
+
+	keepClosedFn := func(depositAddress string) (chan struct{}, func(), error) {
+		return make(chan struct{}), func() {}, nil
+	}
+
+	actCounter := 0
+	actFn := func(depositAddress string) error {
+		actCounter++
```
pdyraga, line 1891: Shouldn't we use atomic counter? https://golang.org/pkg/sync/atomic/#AddUint64

pdyraga (commented):
Two small comments. Otherwise, once https://github.com/keep-network/keep-ecdsa/pull/590 is merged I am good to merge this PR as well.

pdyraga (approved):

PR #592, opened 2020-11-03
Add TBTC monitoring confirmations

lukasz-zimnoch:
Refs: #574
Depends on: #591 
Depends on: https://github.com/keep-network/keep-common/pull/56

Implemented chain confirmations upon monitoring stop events and after executing monitoring actions. This should prevent problems that may occur due to chain reorganizations.


pkg/chain/confirmation.go
```diff
@@ -0,0 +1,32 @@
...
+import (
+	"fmt"
+)
+
+// WaitForChainConfirmation ensures that after receiving specific number of block
+// confirmations the state of the chain is actually as expected. It waits for
+// predefined number of blocks since the start block number provided. After the
+// required block number is reached it performs a check of the chain state with
+// a provided function returning a boolean value.
+func WaitForChainConfirmation(
```
pdyraga, line 12: Can we put this guy in `keep-common`? We already have `MiningWaiter` there and at some point, we'll want to integrate this check with `keep-core` as well.

pkg/extensions/tbtc/tbtc_test.go
```diff
@@ -1538,6 +1561,7 @@ func TestProvideRedemptionProof_ContextCancelled_WithWorkingMonitoring(
 		ctx,
 		constantBackoff,
 		timeout,
+		defaultLocalBlockConfirmations,
 	)
 	if err != nil {
 		t.Fatal(err)
```
pdyraga, line 1567: It would be good to have tests validating monitoring goroutines are really confirming the state on-chain. We may need to do it in another file with some short block times to do not hang on those tests too long.

lukasz-zimnoch: Those tests are really confirming the state. They are just doing this after `0` blocks.

pkg/chain/confirmation.go
lukasz-zimnoch: Can be a little bit problematic as it depends on `chain.Handle`. From the other side, passing the block counter alone seems a bit weird.

pdyraga: Does it? I think it is fine - this function does not need anything else from the chain than the acces to the util counting blocks. We'll pass exactly what this function needs.

What feels a bit weird to me is this function in this package next to interface declarations, I think it's a perfect candidate for `keep-common` chain utilities. Also, what triggers a warning to me is `logger` declaration we put in `chain.go` next to interface declarations. Logger should live next to the implementation, not the interface. 

Yeah, let's please make it a `keep-common` code.

lukasz-zimnoch: Ok, done.

pkg/client/client.go
```diff
@@ -3,11 +3,12 @@ package client
 
 import (
 	"context"
-	"fmt"
 	"math/big"
 	"sync"
 	"time"
 
+	"github.com/keep-network/keep-common/pkg/chain/ethereum/ethutil"
```
pdyraga, line 10: Oh, I see my suggestion about leaving confirmations in `ethutil` for now does not play well here. This code should not import anything ethereum-specific. True, it's already importing `go-ethereum/common` for `common.Address` but I don't think it's right. Let me fix it.

pdyraga: Please see https://github.com/keep-network/keep-common/pull/57

pkg/extensions/tbtc/tbtc_test.go
```diff
@@ -17,20 +17,30 @@ import (
...
+	defaultLocalBlockConfirmations = 0
 )
 
+func newTestTBTC(chain chain.TBTCHandle) *tbtc {
+	tbtc := newTBTC(chain)
+
+	tbtc.blockConfirmations = defaultLocalBlockConfirmations
+
+	return tbtc
+}
```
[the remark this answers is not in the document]
lukasz-zimnoch: This PR https://github.com/keep-network/keep-ecdsa/pull/600 will give more sense for the `newTBTC`. Also, the `tbtc` struct will probably grow over time.

pkg/extensions/tbtc/tbtc.go
```diff
@@ -91,18 +96,51 @@ func (t *tbtc) monitorRetrievePubKey(
...
+		handler depositEventHandler,
+	) (subscription.EventSubscription, error) {
+		return t.chain.OnDepositCreated(handler)
+	}
+
+	monitoringStopFn := func(
+		handler depositEventHandler,
+	) (subscription.EventSubscription, error) {
+		return t.chain.OnDepositRegisteredPubkey(func(depositAddress string) {
+			if t.waitDepositStateChangeConfirmation(
```
pdyraga, line 111: Should we emit a warning if this is not true? I agree that error is good for `actFn`, here error could be too much but warning feels good.

pkg/extensions/tbtc/tbtc_test.go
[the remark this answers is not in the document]
pdyraga: Cool, I'll leave this discussion open for now but let's don't care about it for this PR and revisit once we merge #600.

pkg/extensions/tbtc/tbtc.go
```diff
@@ -128,6 +166,8 @@ func (t *tbtc) monitorProvideRedemptionSignature(
 	actBackoffFn backoffFn,
 	timeout time.Duration,
 ) error {
+	initialDepositState := chain.AwaitingWithdrawalSignature
+
 	monitoringStartFn := func(
```
pdyraga, line 178: Not for this PR and probably not for this release but I wonder whether we shouldn't wait e.g. 6 blocks before we start the monitoring. If we do, we can avoid dangling goroutines. On the other hand, even if the transaction was dropped, it should soon be included in another block as the application should keep tracking deposit progress. Just something to note and think about.

pkg/extensions/tbtc/tbtc.go
```diff
@@ -417,9 +515,6 @@ type backoffFn func(iteration int) time.Duration
 
 type timeoutFn func(depositAddress string) (time.Duration, error)
 
-// TODO:
-//  1. Handle chain reorgs (keep-ecdsa/pull/585#discussion_r511760283 and keep-ecdsa/pull/585#discussion_r513447505)
-//  2. Resume monitoring after client restart.
```
pdyraga, line 422: Point 2 should stay, correct?

pkg/extensions/tbtc/tbtc.go
```diff
@@ -645,6 +744,84 @@ func (t *tbtc) shouldMonitorDeposit(depositAddress string) bool {
...
+			depositAddress,
+			err,
+		)
+		return false
+	}
+
+	return confirmed
+}
+
+func (t *tbtc) waitKeepInactivityConfirmation(
```
pdyraga, line 826: Can we rename it to `waitKeepNotActiveConfirmation`? *Inactivity* could suggest keep has not provided the requested signature and it's been terminated. *NotActive* relates more to the keep state.
(Still not perfect but I do not have any better idea).

pkg/extensions/tbtc/tbtc.go
```diff
@@ -645,6 +744,84 @@ func (t *tbtc) shouldMonitorDeposit(depositAddress string) bool {
...
+	keepAddress string,
+) bool {
+	currentBlock, err := t.chain.BlockCounter().CurrentBlock()
+	if err != nil {
+		logger.Errorf(
+			"could not get current block while confirming "+
+				"inactivity for keep [%v]: [%v]",
```
pdyraga, line 833: `...while confirming keep [%v] is not active: [%v]`?

pkg/extensions/tbtc/tbtc.go
```diff
@@ -645,6 +744,84 @@ func (t *tbtc) shouldMonitorDeposit(depositAddress string) bool {
...
+		t.chain.BlockCounter(),
+		currentBlock,
+		t.blockConfirmations,
+		func() (bool, error) {
+			return t.chain.IsActive(common.HexToAddress(keepAddress))
+		},
+	)
+	if err != nil {
+		logger.Errorf(
+			"could not confirm inactivity for keep [%v]: [%v]",
```
pdyraga, line 850: `could not confirm if keep [%v] is not active: [%v]`?

pkg/extensions/tbtc/tbtc.go
lukasz-zimnoch: I removed this because I think we should first discuss if we really need restart handling and if so, how we should approach it. Because of that, I think we should rather create a separate GH issue than leave a TODO.

pdyraga: Oh, there will be an issue today, that's for sure :) This TODO was just like a comment for someone else browsing this code and wondering how restarts are handled.

pkg/extensions/tbtc/tbtc_test.go
pdyraga: Correct but if I completely remove the code calling the confirmation function, tests will still pass, correct?

lukasz-zimnoch: I understand your point. Those tests will still pass in that case but from the other side, confirmations make sense only when used with Ethereum. Writing separate tests for them using the local chain will say us the local chain is correctly implemented and nothing more.

pkg/extensions/tbtc/tbtc.go
lukasz-zimnoch: Sure I can leave that TODO then.

pkg/extensions/tbtc/tbtc_test.go
pdyraga: Unless we increase the block time for a local chain and... assert on time? OK, I do agree it's one of those rare cases where it's probably not worth it.

pkg/client/client.go
pdyraga: We still need to bump `keep-common` dependency and update calls.

lukasz-zimnoch: Done.

pdyraga (approved):

PR #593, opened 2020-11-03
ECDSA staker rewards cap set to 3M keep tokens per interval

pdyraga:
Refs #446

An individual beneficiary can not receive more than 3M KEEP tokens in a single interval. This amount is about 20% of the second interval full allocation.

I have made `beneficiaryRewardCap` a public constant. It's a different situation to the rest of the constants defined in `ECDSARewards` as they are only used as constructor parameters for the parent contract.

Added test to make sure this value is set at `3M * 10^18` just to prevent accidental change and ensure the order of operators is correct.


dimpar (approved):
Lgtm!

PR #594, opened 2020-11-03
Improve Ethereum client wrappers adoption

lukasz-zimnoch:
Refs: #574
Depends on: #585
Depends on: https://github.com/keep-network/keep-common/pull/55

Changed the way Ethereum client is used in the Ethereum chain implementation. Specifically, got rid of the bare client
usages. Instead of that, the wrapped client which provides logging and rate-limiting capabilities is used everywhere.

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -335,6 +335,8 @@ github.com/keep-network/keep-common v1.2.0 h1:hVd2tTd7vL+9CQP5Ntk5kjs+GYvkgrRNBc
 github.com/keep-network/keep-common v1.2.0/go.mod h1:emxogTbBdey7M3jOzfxZOdfn139kN2mI2b2wA6AHKKo=
 github.com/keep-network/keep-common v1.2.1-0.20201020114759-19c123cbd4f4 h1:CivupPSFswHACua5xZGKdeYxsCQ2cmRomTIBh8kfk70=
 github.com/keep-network/keep-common v1.2.1-0.20201020114759-19c123cbd4f4/go.mod h1:emxogTbBdey7M3jOzfxZOdfn139kN2mI2b2wA6AHKKo=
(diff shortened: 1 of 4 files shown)


go.sum
```diff
@@ -335,6 +335,8 @@ github.com/keep-network/keep-common v1.2.0 h1:hVd2tTd7vL+9CQP5Ntk5kjs+GYvkgrRNBc
 github.com/keep-network/keep-common v1.2.0/go.mod h1:emxogTbBdey7M3jOzfxZOdfn139kN2mI2b2wA6AHKKo=
 github.com/keep-network/keep-common v1.2.1-0.20201020114759-19c123cbd4f4 h1:CivupPSFswHACua5xZGKdeYxsCQ2cmRomTIBh8kfk70=
 github.com/keep-network/keep-common v1.2.1-0.20201020114759-19c123cbd4f4/go.mod h1:emxogTbBdey7M3jOzfxZOdfn139kN2mI2b2wA6AHKKo=
+github.com/keep-network/keep-common v1.2.1-0.20201103151750-c52f5d480c10 h1:KuZP1ArRAK1X4PunfteJX+QnP/XSfvfmi/ZR2a6kXTU=
+github.com/keep-network/keep-common v1.2.1-0.20201103151750-c52f5d480c10/go.mod h1:emxogTbBdey7M3jOzfxZOdfn139kN2mI2b2wA6AHKKo=
```
pdyraga, line 339: Let's remember to `go mod tidy` before we merge.

pdyraga (commented):
Code looks good. Holding with merge until all previous PRs are merged and this one can be un-drafted and tested.

pdyraga:
@lukasz-zimnoch Can you please update this branch now having all 3 previous PRs merged? I'd love to have the fresh version of the code here before doing a final pass.

go.sum
lukasz-zimnoch: Done that but nothing changes.

pdyraga: This is weird, tried it as well. Let's leave it as-is and we'll revisit once we switch `keep-core` to the new version of `keep-common`.

pdyraga (approved):

Issue #597, opened 2020-11-05
ECDSA staker rewards adjustments for ETH-only delegation


pdyraga:
As currently implemented, ECDSA staker reward contracts support only one ECDSA keep factory. To fully implement ETH-only delegations for ECDSA operators we need to find a way to include ETH-only keeps in staker rewards. One option is updating reward contracts to support two factories. Another option - if math would make sense - could be to deploy another rewards contract for ETH-only stakers.

pdyraga:
We decided to not implement this natively in the protocol but instead introduce a KEEP-only pool. At the same time, there are efforts in Keep community to implement community staking pools.

PR #598, opened 2020-11-09
Make rewards contract able to handle multiple concurrent factories

eth-r:
Refs: #597
Depends on https://github.com/keep-network/keep-core/pull/2158


pdyraga:
Closing without merging as we changed the approach for ECDSA staker rewards. The new approach is described in https://blog.keep.network/a-new-rewards-mechanism-deef3412c3e1 and work can be tracked in https://github.com/keep-network/keep-ecdsa/issues/602.

PR #599, opened 2020-11-10
Adjust ECDSA keep rewards by lifespan

eth-r:
Refs: #602

Reward includes a constant fraction that is paid to all happily closed keeps, and a variable fraction which is paid based on the lifespan of the keep.

Keep lifespans are estimated by having users report closed keeps. If the closure of a keep has not been reported, assume it lasted long enough to be eligible for maximum reward.

### DIFF
--- a/solidity/contracts/ECDSARewards.sol
+++ b/solidity/contracts/ECDSARewards.sol
@@ -196,12 +196,18 @@ contract ECDSARewards is Rewards {
     function sendHeartbeat(bytes32 _keep) public {
         require(!_isClosed(_keep),"Keep is closed");
         require(!_isTerminated(_keep), "Keep is terminated");
+        address keepAddress = toAddress(_keep);
 
         // Assumes bonds don't change while the keep is active
-        BondedECDSAKeep keep = BondedECDSAKeep(toAddress(_keep));
-        uint128 keepBond = uint128(keep.checkBondAmount());
+        Heartbeat previousHeartbeat = keepHeartbeats[keepAddress];
+        uint128 keepBond = previousHeartbeat.bondValue;
 
-        keepHeartbeats[address(keep)] = Heartbeat(
+        if (keepBond == 0) {
+            BondedECDSAKeep keep = BondedECDSAKeep(keepAddress);
+            keepBond = uint128(keep.checkBondAmount());
+        }
+
+        keepHeartbeats[keepAddress] = Heartbeat(
             uint128(block.timestamp),
             keepBond
         );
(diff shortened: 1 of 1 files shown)


solidity/contracts/ECDSARewards.sol
```diff
@@ -182,6 +193,20 @@ contract ECDSARewards is Rewards {
         token.safeTransfer(beneficiary, withdrawableRewards);
     }
 
+    function sendHeartbeat(bytes32 _keep) public {
```
pdyraga, line 196: What's the gas cost of this function? Assuming it is called for every keep, we should *probably* ensure the minimum reward >> heartbeat cost.

solidity/contracts/ECDSARewards.sol
```diff
@@ -266,6 +294,32 @@ contract ECDSARewards is Rewards {
...
+        internal returns (uint256) {
+        uint256 createdAt = _getCreationTime(_keep);
+
+        Heartbeat lastHeartbeat = keepHeartbeats[toAddress(_keep)];
+
+        uint256 lastTimestamp = uint256(lastHeartbeat.timestamp);
+        uint256 keepBond = uint256(lastHeartbeat.bondAmount);
+
+        uint256 rewardLifespan = lastTimestamp == 0
+            ? assumedMinimumLifetime
```
pdyraga, line 307: In this setup, do I need to send a heartbeat before the funding timeout? My understanding is that I don't have to and the only heartbeat needed is once the redemption has been requested.

solidity/contracts/ECDSARewards.sol
```diff
@@ -266,6 +294,32 @@ contract ECDSARewards is Rewards {
...
+            : lastTimestamp.sub(createdAt);
+
+        uint256 rewardBond = keepBond == 0
+            ? assumedMinimumBond
```
pdyraga, line 311: OK, and here I see the first heartbeat is needed because otherwise the minimum bond can be assumed. Still, I think both `assumedMiniumumLifetime` and `assumedMinimumBond` make sense.

solidity/contracts/ECDSARewards.sol
eth-r: For the first call, it should be a few thousand in the necessary queries + 20,000 for the storage. For later updates the overhead goes down by 15,000 because we aren't initializing a previously empty storage slot.

pdyraga:
Closing without merging as we changed the approach for ECDSA staker rewards. The new approach is described in https://blog.keep.network/a-new-rewards-mechanism-deef3412c3e1 and work can be tracked in https://github.com/keep-network/keep-ecdsa/issues/602.

PR #600, opened 2020-11-12
TBTC extension optimizations

lukasz-zimnoch:
Refs: #574
Depends on: https://github.com/keep-network/keep-ecdsa/pull/592

Here we add some TBTC extension optimizations that should reduce costs related to Ethereum integration. Specifically:

- We introduce a time cache for the monitoring filter. Determining whether a deposit should be tracked requires two ETH calls. This check is performed for each deposit multiple times as a response to monitored state transition events and all duplicates of those events. This can result in a significant number of unnecessary ETH calls. Introducing a time cache here should greatly reduce the number of calls.
- We implement a signer-based action delay factor. Currently, all monitoring signers act at the same time. As result, only the first transaction is always successful while the rest revert and burn some amount of gas by the way. The change introduced here forces each signer to wait an additional time, according to its signer index, before acting. This should prevent other signers to act if a successful action has been already made.

### DIFF
--- a/pkg/extensions/tbtc/tbtc.go
+++ b/pkg/extensions/tbtc/tbtc.go
@@ -26,10 +26,23 @@ import (
 var logger = log.Logger("tbtc-extension")
 
 const (
-	maxActAttempts               = 3
-	pastEventsLookbackBlocks     = 10000
-	defaultBlockConfirmations    = 12
-	monitoringCachePeriod        = 24 * time.Hour
+	// Maximum number of action attempts before giving up and returning
+	// a monitoring error.
+	maxActAttempts = 3
+
+	// Determines how many blocks from the past should be included
+	// during the past events lookup.
+	pastEventsLookbackBlocks = 10000
+
+	// Number of blocks which should elapse before confirming
+	// the given chain state expectations.
+	defaultBlockConfirmations = 12
+
+	// Determines how long the monitoring cache will maintain
+	// its entries.
+	monitoringCachePeriod = 24 * time.Hour
+
+	// Used to calculate the action delay factor for the given signer index.
 	defaultSignerActionDelayStep = 5 * time.Minute
 )
 
--- a/pkg/extensions/tbtc/tbtc.go
+++ b/pkg/extensions/tbtc/tbtc.go
@@ -26,10 +26,11 @@ import (
 var logger = log.Logger("tbtc-extension")
 
 const (
-	maxActAttempts            = 3
-	pastEventsLookbackBlocks  = 10000
-	defaultBlockConfirmations = 12
-	monitoringCachePeriod     = 24 * time.Hour
+	maxActAttempts               = 3
+	pastEventsLookbackBlocks     = 10000
+	defaultBlockConfirmations    = 12
+	monitoringCachePeriod        = 24 * time.Hour
+	defaultSignerActionDelayStep = 5 * time.Minute
 )
 
 // Initialize initializes extension specific to the TBTC application.
@@ -87,6 +88,7 @@ type tbtc struct {
 	blockConfirmations        uint64
 	monitoredDepositsCache    *cache.TimeCache
 	notMonitoredDepositsCache *cache.TimeCache
+	signerActionDelayStep     time.Duration
 }
 
 func newTBTC(chain chain.TBTCHandle) *tbtc {
@@ -95,6 +97,7 @@ func newTBTC(chain chain.TBTCHandle) *tbtc {
 		blockConfirmations:        defaultBlockConfirmations,
 		monitoredDepositsCache:    cache.NewTimeCache(monitoringCachePeriod),
 		notMonitoredDepositsCache: cache.NewTimeCache(monitoringCachePeriod),
+		signerActionDelayStep:     defaultSignerActionDelayStep,
 	}
 }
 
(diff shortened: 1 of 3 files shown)


pkg/extensions/tbtc/tbtc.go
```diff
@@ -24,9 +26,11 @@ import (
...
 
 const (
-	maxActAttempts            = 3
-	pastEventsLookbackBlocks  = 10000
-	defaultBlockConfirmations = 12
+	maxActAttempts               = 3
+	pastEventsLookbackBlocks     = 10000
+	defaultBlockConfirmations    = 12
+	monitoringCachePeriod        = 24 * time.Hour
+	defaultSignerActionDelayStep = 5 * time.Minute
```
pdyraga, line 33: Non-blocking but we should document those constants.

pkg/extensions/tbtc/tbtc.go
```diff
@@ -761,23 +790,58 @@ func (t *tbtc) watchKeepClosed(
...
-	for _, member := range members {
+	for index, member := range members {
 		if member == t.chain.Address() {
-			return true
+			return index, nil
 		}
 	}
 
-	return false
+	return 0, fmt.Errorf(
```
pdyraga, line 830: Instead of returning an error, I'd probably return `-1, nil`. Then, inside `shouldMonitorDeposit` we can decide whether we should not monitor the deposit or if we do not want we should be doing that because of an error during the check.

pkg/extensions/tbtc/tbtc.go
```diff
@@ -761,23 +790,58 @@ func (t *tbtc) watchKeepClosed(
...
+	}
+
+	if t.notMonitoredDepositsCache.Has(depositAddress) {
+		return false
+	}
+
+	if _, err := t.getSignerIndex(depositAddress); err != nil {
+		t.notMonitoredDepositsCache.Add(depositAddress)
+		return false
+	}
```
pdyraga, line 807: We don't really know here whether we shouldn't monitor or whether we failed to check if we should monitor. Please see my comment to `getSignerIndex` function.

pkg/extensions/tbtc/tbtc.go
```diff
@@ -761,23 +790,58 @@ func (t *tbtc) watchKeepClosed(
 }
 
 func (t *tbtc) shouldMonitorDeposit(depositAddress string) bool {
```
pdyraga, line 792: What do you think about returning `(bool, int)` from this function and having `timeoutFn` accept `memberIndex int` instead of `depositAddress string`? We can spare two Ethereum client calls this way.

pdyraga, line 792: Can we add unit tests for this function covering caching? Just like we have for the firewall make sure that depending on `getSignerIndex` result, for the next `shouldMonitorDeposit` invocation, chain call is not executed and appropriate value is returned?

pkg/extensions/tbtc/tbtc.go
```diff
@@ -761,23 +790,58 @@ func (t *tbtc) watchKeepClosed(
...
 	}
 
-	return false
+	return 0, fmt.Errorf(
+		"operator is not a signer of deposit [%v]",
+		depositAddress,
+	)
+}
+
+func (t *tbtc) getSignerActionDelay(
```
pdyraga, line 836: We should have a simple unit test for this function, we don't necessarily need to test the full path with timeout but at least the calculations here.

pkg/extensions/tbtc/tbtc.go
lukasz-zimnoch: What in the case when we get the result from the cache?

pdyraga: Ah yes. OK, it's more complex. Let's leave it as is, we can always revisit later.

pkg/extensions/tbtc/tbtc.go
```diff
@@ -24,9 +26,24 @@ import (
...
+
+	// Number of blocks which should elapse before confirming
+	// the given chain state expectations.
 	defaultBlockConfirmations = 12
+
+	// Determines how long the monitoring cache will maintain
+	// its entries.
+	monitoringCachePeriod = 24 * time.Hour
+
+	// Used to calculate the action delay factor for the given signer index.
```
pdyraga, line 45: `... to avoid all signers executing the same action for deposit at the same time` ?

pkg/extensions/tbtc/tbtc.go
```diff
@@ -24,9 +26,24 @@ import (
...
+	// Determines how many blocks from the past should be included
+	// during the past events lookup.
+	pastEventsLookbackBlocks = 10000
+
+	// Number of blocks which should elapse before confirming
+	// the given chain state expectations.
 	defaultBlockConfirmations = 12
+
+	// Determines how long the monitoring cache will maintain
+	// its entries.
```
pdyraga, line 42: `... about which deposits should be monitored by this client instance`?

pkg/extensions/tbtc/tbtc.go
```diff
@@ -761,23 +803,65 @@ func (t *tbtc) watchKeepClosed(
...
+	}
+
+	signerIndex, err := t.getSignerIndex(depositAddress)
 	if err != nil {
+		logger.Errorf(
+			"could not get signer index for deposit [%v]: [%v]",
+			depositAddress,
+			err,
+		)
+		return false // return false but don't cache the result in case of error
```
pdyraga, line 824: This is a good decision but I'd add some more context to the error we log about what are the implications of it. My proposition:
```
could not check if deposit [%v] should be monitored: failed to get signer index: [%v]`
```

pdyraga (approved):
🚀

Issue #603, opened 2020-11-13
Confirm chain state after submitting keep's public key and signature


pdyraga:
During the work on application-specific actions, we abstracted utility allowing to wait for a certain number of blocks to confirm the deposit state, and if the state is not the one we expect, trigger the action again. A similar mechanism can be used to confirm a public key has been submitted to the keep or that the requested signature has been submitted to improve handling of minor-to-mid chain reorgs.

Please see [`WaitForBlockConfirmations`](https://github.com/keep-network/keep-common/blob/d53a6658b425f3baa306b1ed08aa2a6159347daf/pkg/chain/chainutil/chainutil.go) in `keep-common`.

PR #610, opened 2020-11-16
Add chain confirmations for keygen and signing 

lukasz-zimnoch:
Closes: #603 

This pull request introduces chain confirmations for keygen and signing. This allows confirming the public key or signature have actually reached the chain so the client can retry in case of failure. This change should also help in the case of small chain reorgs.


pkg/node/node.go
```diff
@@ -250,9 +254,102 @@ func (n *Node) publishSignerPublicKey(
...
+
+	for iteration := 1; iteration <= maxIterations; iteration++ {
+		logger.Infof(
+			"starting iteration [%v] of member public key "+
+				"submission monitoring for keep: [%v]",
+			iteration,
+			keepAddress.Hex(),
+		)
+
+		currentBlock, err := n.ethereumChain.BlockCounter().CurrentBlock()
```
pdyraga, line 283: Why don't we use `WaitForBlockConfirmations` from `keep-common` for all of this? The action is public key resubmission.

pdyraga (commented):
Why don't we use `WaitForBlockConfirmations` from `keep-common` for all of this? The action is public key resubmission.

pkg/node/node.go
```diff
@@ -433,8 +481,8 @@ func (n *Node) publishSignature(
 // conflicting public key is published or until keep established public key
 // or until key generation timed out.
```
pdyraga, line 482: I think the documentation is no longer up to date.

pkg/node/node.go
```diff
@@ -476,32 +524,105 @@ func (n *Node) monitorKeepPublicKeySubmission(
...
+	}
+
+	for {
+		select {
+		case event := <-publicKeyPublished:
+			logger.Infof(
+				"public key [%x] has been accepted by keep: [%s]",
+				event.PublicKey,
+				keepAddress.String(),
+			)
```
pdyraga, line 558: That's tricky but I think the 12 blocks confirmation that will potentially result in `return`ing from this `for`, should be triggered here. Otherwise, we are waiting for 12 blocks from the moment we submitted the transaction to the mempool and not from the moment it's been mined.

pdyraga: We can use `WaitForBlockConfirmations` for that - if the function returns `true`, we can send a kill signal to a channel that will close this loop, e.g.

```
case event := <-publicKeyConfirmed:
  logger.Info("public key confirmed for keep [%v] [%v] blocks later, quitting monitoring", ...)
  return
case event := <- publicKeyPublished
  logger.Info("public key published for keep [%v], waiting [%v] blocks to confirm the state",...)

  isSignatureConfirmed, err := chainutil.WaitForBlockConfirmations()
  // if err, give up
  if (isSignatureConfirmed) {
    publicKeyConfirmed <-...
  }
case pubkeyCheckBlock := <-pubkeyCheckTrigger:
  // just like now, except that we do not return but just try to re-submit
```

pkg/node/node.go
```diff
@@ -424,6 +425,53 @@ func (n *Node) publishSignature(
...
+				keepAddress.String(),
+				err,
+			)
+			time.Sleep(retryDelay) // TODO: #413 Replace with backoff.
+			continue
+		}
+
+		if !isSignatureConfirmed {
+			logger.Errorf(
+				"signature for keep [%s] is not confirmed",
```
pdyraga, line 468: `"signature submission for keep [%s] not confirmed after [%v] blocks, trying to submit again"` ?

pkg/node/node.go
pdyraga: My proposition:
```
// monitorKeepPublicKeySubmission observes the chain until either the first
// conflicting public key is published or until keep established public key
// or until the key generation timed out. It also tries to re-submit the public
// key if it has not been submitted in the given time frame as a way to protect
// against chain forks (i.e. transaction mined in a fork that has been dropped).
// 
// When event informing about conflicting public key is emitted from the chain,
// monitoring stops immediately. Even if this event was emitted in a minority
// fork, it is clear that the operator who submitted the conflicting key is
// dishonest and it is better to abandon this keep.
//
// Function checks the chain periodically and inspects whether the public key
// has been registered on the chain. If the key is there, the function waits
// for the required number of confirmations and repeats the check. If the public
// key for the keep is still there, the monitoring exits successfully.
// In the case when public key for the keep is not yet established during the
// periodic check or it is gone after waiting for a certain number of
// confirmations (chain reorganization), this function will attempt to submit
// the public key again.
```

pkg/node/node.go
```diff
@@ -474,34 +507,126 @@ func (n *Node) monitorKeepPublicKeySubmission(
...
+		select {
+		case event := <-conflictingPublicKey:
+			logger.Errorf(
+				"member [%x] has submitted conflicting public key for keep [%s]: [%x]",
+				event.SubmittingMember,
+				keepAddress.String(),
+				event.ConflictingPublicKey,
+			)
+			return
+		case pubkeyCheckBlock := <-pubkeyCheckTrigger:
```
pdyraga, line 545: Let's explain why do we execute the check periodically instead of waiting for an event.

pkg/node/node.go
```diff
@@ -474,34 +507,126 @@ func (n *Node) monitorKeepPublicKeySubmission(
...
+					"keep [%s] still does not have public key "+
+						"and resubmission by this member failed with: [%v]",
+					keepAddress.String(),
+					err,
+				)
+			}
+
+			pubkeyCheckTrigger, err = n.ethereumChain.BlockCounter().BlockHeightWaiter(
+				pubkeyCheckBlock + blockConfirmations,
+			)
```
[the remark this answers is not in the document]
lukasz-zimnoch: Maybe I'm missing something but can you clarify? We have the same code in `registration.go` which works well.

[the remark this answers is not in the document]
pdyraga: Nevermind, my brain was in Scala land.

pkg/node/node.go
```diff
@@ -474,34 +524,138 @@ func (n *Node) monitorKeepPublicKeySubmission(
...
+			// trusted because they may come from a forked chain or can
+			// the same event can be delivered multiple times.
+			keepPublicKey, err := n.ethereumChain.GetPublicKey(keepAddress)
+			if err != nil {
+				logger.Errorf(
+					"failed to get keep public key during "+
+						"public key submission monitoring for keep [%s]: [%v]",
+					keepAddress.String(),
+					err,
+				)
```
pdyraga, line 589: Shouldn't we have a `continue` here? If we failed to get this key, let's wait for another tick instead of trying to resubmit IMO.

pkg/node/node.go
```diff
@@ -474,34 +524,138 @@ func (n *Node) monitorKeepPublicKeySubmission(
...
+					)
+					if err != nil {
+						logger.Errorf(
+							"failed to perform keep public key "+
+								"confirmation during public key submission "+
+								"monitoring for keep [%s]: [%v]",
+							keepAddress.String(),
+							err,
+						)
+					}
```
pdyraga, line 615: Same here, I'd probably `continue` instead of submitting the key again 🤔

pkg/node/node.go
```diff
@@ -474,34 +524,138 @@ func (n *Node) monitorKeepPublicKeySubmission(
...
+								"has been confirmed",
+							keepPublicKey,
+							keepAddress.String(),
+						)
+						return
+					}
+				}
+			}
+
+			err = n.ethereumChain.SubmitKeepPublicKey(keepAddress, publicKey)
```
pdyraga, line 629: Could be worth logging in INFO that public key could not be confirmed after 12 blocks and we are attempting to re-submit.

pkg/node/node.go
lukasz-zimnoch: We can't just `continue` because we have to set the new `pubkeyCheckTrigger` at the end of this case.

pkg/node/node.go
lukasz-zimnoch: We can't just `continue` because we have to set the new `pubkeyCheckTrigger` at the end of this case.

pkg/node/node.go
```diff
@@ -374,7 +375,7 @@ func (n *Node) publishSignature(
 
 		// Someone submitted the signature and it was accepted by the keep.
 		// We are fine, leaving.
```
pdyraga, line 377:
```
// Someone submitted the signature, it was accepted by the keep,
// and there are enough confirmations from the chain.
// We are fine, leaving.
```
?

pkg/node/node.go
```diff
@@ -403,7 +404,7 @@ func (n *Node) publishSignature(
 
 			// Check if we failed because someone else submitted in the meantime
 			// or because something wrong happened with our transaction.
```
pdyraga, line 406: Maybe:
```
// Check if we failed because someone else submitted in the meantime
// or because something wrong happened with our transaction.
// If someone else submitted in the meantime, wait for enough confirmations
// from the chain before making a decision about leaving the submission process.
```
?

pkg/node/node.go
pdyraga: Yeah, I see. Well, it's not the end of the world and I don't think it's worth making this code even more complex.

pkg/node/node.go
pdyraga: Yeah, I see. Well, it's not the end of the world and I don't think it's worth making this code even more complex.

pkg/node/node.go
```diff
@@ -424,41 +429,109 @@ func (n *Node) publishSignature(
...
+// the public key again.
+func (n *Node) monitorKeepPublicKeySubmission(
+	keepAddress common.Address,
+	publicKey [64]byte,
+) {
+	monitoringCtx, monitoringCancel := context.WithTimeout(
+		context.Background(),
+		monitorKeepPublicKeySubmissionTimeout,
+	)
+	defer monitoringCancel()
```
pdyraga, line 531: I am afraid there is one thing we did not take into account: retried `SubmitPublicKey` transactions are not failing on gas estimate but they are getting mined and are burning gas on a failed transaction. For `SubmitPublicKey`, the gas estimate is hardcoded in `ethereum` implementation. If it is good or not, that's another discussion but it becomes very clear now than in this code we should not make assumptions about how the implementation works.

The scenario I was testing:
Three nodes in the group and one node exited just before submitting the public key but this can be also a fatal Ethereum client failure or any other issue with it. Two other nodes kept trying to submit their public keys, burning `29447` gas on each attempt:
```
2020-11-18T16:58:47.712+0100	INFO	keep-ecdsa	keep [0x8914a0983C0C9803Bcf70B13C132A8f9B41F3D4C] still does not have a confirmed public key; re-submitting public key [c4d5a7e8702776e2da4ac1c9873f9c07ef8c99af1c22da27e2c7b0ae868c78384fd7b6b947a627aaa40950372692e4850d1f021457809c4bcacb528d081dd335]
2020-11-18T16:58:48.320+0100	INFO	keep-contract-BondedECDSAKeep	submitted transaction submitPublicKey with id: [0x2c4e0881f5bd9561fa3fa0332e0c4cc6c3368c294315b5ac6a3f09b907d7ac66] and nonce [170]
2020-11-18T16:58:51.325+0100	INFO	keep-ethutil	transaction [2c4e08…d7ac66] mined with status [0] at block [1220]
```

Depending on the gas price and block times, those clients can burn more ether on those failed transactions than it would be seized from them for not publishing a keep public key.

I was considering whether we can:
- remove `monitoringCtx` altogether,
- change `pubkeyCheckTrigger` to tick on time, say every 20 minutes,
- give up after two executions of `case pubkeyCheckBlock := <-pubkeyCheckTrigger`.

What I like about this solution is that we have a fixed number of retries so there is a limit of potentially failed transactions and we use time as a ticker, just like in the monitoring of signature submission, so we are consistent.

WDYT?

pkg/node/node.go
```diff
@@ -424,41 +429,109 @@ func (n *Node) publishSignature(
 			continue
 		}
 
+		if !n.confirmSignature(keepAddress, digest) {
```
pdyraga, line 432: I think we should perhaps add a blocking function `waitForSignature` that will block for e.g. 10 minutes and check every one minute if the signature has been submitted:
```
if !(n.waitForSignature(keepAddress, digest) && n.confirmSignature(keepAddress.digest))
```

The scenario I was testing:
Three group members submitted signature transactions at the same time. None of them has failed on gas estimation and all three transactions landed in the mempool. They were underpriced compared to other transactions and it took 10 blocks to mine first of them. In this scenario, `confirmSignature` verified only 2 confirmations.

It seems to me that this check is the only one affected. WDYT?

pkg/node/node.go
```diff
@@ -424,41 +429,109 @@ func (n *Node) publishSignature(
...
+			time.Sleep(retryDelay) // TODO: #413 Replace with backoff.
+			continue
+		}
+
 		logger.Infof("signature submitted for keep [%s]", keepAddress.String())
```
pdyraga, line 437: There is a little weird effect in the logs now:
```
2020-11-19T11:45:34.694+0100	INFO	keep-ecdsa	signature submission for keep [0xe175585397E143636920390ecfFc11828ec80E4e] has been confirmed
2020-11-19T11:45:34.694+0100	INFO	keep-ecdsa	signature submitted for keep [0xe175585397E143636920390ecfFc11828ec80E4e]
```

What do you think about removing this log, and letting `confirmSignature` print something like:
```
"signature for keep [%s] successfully submitted and confirmed"
```
?

pkg/node/node.go
```diff
@@ -241,15 +244,13 @@ func (n *Node) publishSignerPublicKey(
 		publicKey,
 	)
 
-	monitoringAbort := make(chan interface{})
-	go n.monitorKeepPublicKeySubmission(monitoringAbort, keepAddress)
-
 	err := n.ethereumChain.SubmitKeepPublicKey(keepAddress, publicKey)
 	if err != nil {
-		close(monitoringAbort)
 		return fmt.Errorf("failed to submit public key: [%v]", err)
```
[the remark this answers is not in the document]
lukasz-zimnoch: I don't think so. `SubmitKeepPublicKey` retries on error under the hood. If we have an error here I don't see any sense to retry from `monitorKeepPublicKeySubmission`. Also, we can't just log the error and rely on asynchronous `monitorKeepPublicKeySubmission` because we have to return the `signer` (or not) and properly handle this keep in the client.

pkg/node/node.go
lukasz-zimnoch: Yep, sounds good. I'll change that in the way you proposed.

pkg/node/node.go
lukasz-zimnoch: Yep, you're right. I'll add the `waitForSignature` function.

pkg/node/node.go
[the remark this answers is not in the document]
pdyraga: If this `SubmitKeepPublicKey` fails, we do not return an error (my proposition), and `monitorKeepPublicKeySubmission` does not eventually succeed with publishing the key to the chain, we will return a `signer` that will be properly handled in the client but will never be asked to sign anything *unless* the transaction will be manually re-submitted from outside of the client by one of the operators. Something we have already seen once or twice by the way.

Now, assuming we leave this code as-is and the first `SubmitKeepPublicKey` does not fail but the transaction will not be mined for whatever the reason - keep in mind no error returned from this function does not guarantee the key to end on the chain - and `monitorKeepPublicKeySubmission` does not eventually succeed with publishing the key to the chain, we are going to have exactly the same situation, right?

What feels wrong to me: the fact this function does not return an error does not guarantee the signature will land on the chain and that the `signer` will ever be asked for providing a signature. Also, we cancel the process on error returned from `SubmitKeepPublicKey` but only for the first invocation. We do not do it for `SubmigKeepPublicKey` inside `monitorKeepPublicKeySubmission`. Maybe there is a valid reason for that but I do not see it yet. Why treat those two identical invocations differently? 

> If we have an error here I don't see any sense to retry from monitorKeepPublicKeySubmission

Let me reverse this statement: if we have an error in `SubmigKeepPublicKey` in `monitorKeepPublicKeySubmission` what is a sense of continuing this loop instead of just returning? 

> `SubmitKeepPublicKey` retries on error under the hood.

That's a very good point. So maybe instead of modifying anything here, we should exit on error returned from `SubmitKeepPublicKey` in `monitorPublicKeySubmission`?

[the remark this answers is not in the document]
pdyraga: More thoughts on this matter: retries in `SubmitKeepPublicKey` helps to overcome infrastructure-level issues like problem with Ethereum client not fully synced with the most recent state.

What we are trying to protect against here are reorgs. And protection about a reorg is all about noting the time the state has changed, then waiting for N blocks, confirming the state, and either repeating an action or leaving. Having this in mind, if feels to me that returning from `monitorKeepPublicKeySubmission` on `SubmitKeepPublicKey`  error is the cleanest option. Thoughts?

pkg/node/node.go
lukasz-zimnoch: I changed but I introduced a constant 10 minutes trigger with a give up after six executions. This is simpler than juggling with two different tickers or weird counters. Also, this change opened the possibility to add `continue` in places you mentioned in this comment https://github.com/keep-network/keep-ecdsa/pull/610#discussion_r526021630

pkg/node/node.go
```diff
@@ -590,12 +567,22 @@ func (n *Node) monitorKeepPublicKeySubmission(
...
+						"has been reached",
+					keepAddress.String(),
+					maxPubkeyChecksCount,
+				)
+			}
+
 			logger.Infof(
-				"checking public key submission for keep [%s] "+
-					"at block [%v]",
+				"checking public key submission for keep [%s]",
```
pdyraga, line 638: `s/checking/confirming`?

pkg/node/node.go
pdyraga: The reason I proposed 5/20/20 checks is to have the first check executed pretty soon so that in the optimistic case - and I assume this is the case for 95% of keeps - the log about the public key confirmed is closer to logs about generating this key. 

10 minutes is also fine but I would consider reducing the number of retries to 3. As you previously mentioned, `SubmitPublicKey` has retries implemented under the hood and here we are just protecting against reorgs. In the case of one client in the group not operating properly, reducing the number of retries from 6 to 3 should let two other clients save gas. And 3 retries protecting against reorgs should be more than enough given that we are currently not retrying and in the vast majority of cases we are fine.

pkg/node/node.go
```diff
@@ -424,41 +429,157 @@ func (n *Node) publishSignature(
...
+		}
+	}
+}
+
+func (n *Node) confirmSignature(
+	keepAddress common.Address,
+	digest [32]byte,
+) bool {
+	logger.Infof(
+		"starting confirming signature submission for keep [%s]",
```
pdyraga, line 501: Can we change it to:
```
confirming on-chain signature submission for keep [%s]
```
?

pkg/node/node.go
```diff
@@ -424,41 +429,157 @@ func (n *Node) publishSignature(
...
+		logger.Errorf(
+			"signature submission for keep [%s] not confirmed; "+
+				"trying to submit the signature again",
+			keepAddress.String(),
+		)
+		return false
 	}
 
+	logger.Infof(
+		"signature for keep [%s] successfully submitted and confirmed",
```
pdyraga, line 551: Can we change it to:
```
signature for keep [%s] successfully submitted and confirmed on-chain
```
?

pdyraga (commented):
Some suggestions about clarifying logs. I'd like to go from:
```
2020-11-20T11:42:00.396+0100	INFO	keep-ecdsa	publishing signature for keep [0x435eaf9C3F0E78FA4f8c330e261cFD4D9F096fe3]; attempt [1]
2020-11-20T11:42:01.390+0100	INFO	keep-contract-BondedECDSAKeep	submitted transaction submitSignature with id: [0x55126c68481fa44a0c81183a5885ef6d9ae0c5bb114c386d8f22c7755bee2053] and nonce [224]
2020-11-20T11:42:01.390+0100	INFO	keep-ecdsa	starting waiting for signature for keep [0x435eaf9C3F0E78FA4f8c330e261cFD4D9F096fe3]
2020-11-20T11:42:04.395+0100	INFO	keep-ethutil	transaction [55126c…ee2053] mined with status [0] at block [9652]
2020-11-20T11:43:01.399+0100	INFO	keep-ecdsa	signature waiter for keep [0x435eaf9C3F0E78FA4f8c330e261cFD4D9F096fe3] has detected the signature
2020-11-20T11:43:01.399+0100	INFO	keep-ecdsa	starting confirming signature submission for keep [0x435eaf9C3F0E78FA4f8c330e261cFD4D9F096fe3]
2020-11-20T11:43:01.400+0100	INFO	keep-chainutil	waiting for block [9667] to confirm chain state
2020-11-20T11:45:25.556+0100	INFO	keep-ecdsa	signature for keep [0x435eaf9C3F0E78FA4f8c330e261cFD4D9F096fe3] successfully submitted and confirmed
```

to:
```
2020-11-20T11:42:00.396+0100	INFO	keep-ecdsa	publishing signature for keep [0x435eaf9C3F0E78FA4f8c330e261cFD4D9F096fe3]; attempt [1]
2020-11-20T11:42:01.390+0100	INFO	keep-contract-BondedECDSAKeep	submitted transaction submitSignature with id: [0x55126c68481fa44a0c81183a5885ef6d9ae0c5bb114c386d8f22c7755bee2053] and nonce [224]
2020-11-20T11:42:01.390+0100	INFO	keep-ecdsa	waiting for signature for keep [0x435eaf9C3F0E78FA4f8c330e261cFD4D9F096fe3] to appear on-chain
2020-11-20T11:42:04.395+0100	INFO	keep-ethutil	transaction [55126c…ee2053] mined with status [0] at block [9652]
2020-11-20T11:43:01.399+0100	INFO	keep-ecdsa	signature for keep [0x435eaf9C3F0E78FA4f8c330e261cFD4D9F096fe3] appeared on-chain
2020-11-20T11:43:01.399+0100	INFO	keep-ecdsa	confirming on-chain signature submission for keep [0x435eaf9C3F0E78FA4f8c330e261cFD4D9F096fe3]
2020-11-20T11:43:01.400+0100	INFO	keep-chainutil	waiting for block [9667] to confirm chain state
2020-11-20T11:45:25.556+0100	INFO	keep-ecdsa	signature for keep [0x435eaf9C3F0E78FA4f8c330e261cFD4D9F096fe3] successfully submitted and confirmed on-chain
```

pkg/node/node.go
```diff
@@ -474,34 +595,132 @@ func (n *Node) monitorKeepPublicKeySubmission(
...
+					"monitoring of public key submission for keep [%s] "+
+						"has been cancelled because maximum checks count [%v] "+
+						"has been reached",
+					keepAddress.String(),
+					maxPubkeyChecksCount,
+				)
+			}
+
+			logger.Infof(
+				"confirming public key submission for keep [%s]",
```
pdyraga, line 638: Can we change it to:
```
confirming on-chain public key submission for keep [%s]
```
? I'd love to keep it aligned with signature part.

pkg/node/node.go
```diff
@@ -474,34 +595,132 @@ func (n *Node) monitorKeepPublicKeySubmission(
...
+							keepAddress.String(),
+							err,
+						)
+						continue
+					}
+
+					if isConfirmed {
+						logger.Infof(
+							"public key [%x] for keep [%s] "+
+								"has been confirmed",
```
pdyraga, line 698: Can we change it to:
```
public key [%x] for keep [%s] successfully submitted and confirmed on-chain
```
to keep it aligned with signature part?

pkg/node/node.go
```diff
@@ -424,41 +429,157 @@ func (n *Node) publishSignature(
...
+		keepAddress.String(),
 	)
-	defer monitoringCancel()
 
-	publicKeyPublished := make(chan *eth.PublicKeyPublishedEvent)
-	conflictingPublicKey := make(chan *eth.ConflictingPublicKeySubmittedEvent)
+	for {
+		select {
+		case <-checkTicker.C:
+			isAwaitingSignature, err := n.ethereumChain.IsAwaitingSignature(
```
pdyraga, line 462: Can we add a comment - same as for public key submission - why do we prefer not to use events in this case?
```
// We check the state periodically instead of relying on
// incoming events. The main motivation is that events could not be
// trusted here because they may come from a forked chain or can
// the same event can be delivered multiple times.
```
?

pkg/node/node.go
```diff
@@ -474,34 +595,132 @@ func (n *Node) monitorKeepPublicKeySubmission(
...
-		)
-	case <-abort:
-		logger.Warningf(
-			"monitoring of public key submission for keep [%s] "+
-				"has been aborted",
-			keepAddress.String(),
-		)
+	pubkeyChecksCounter := 0
+	const maxPubkeyChecksCount = 3
+	const pubkeyCheckTick = 10 * time.Minute
```
pdyraga, line 601: I'd note why we are doing it less aggressive than for signature submission:
```
	// All three operators need to submit public key to the chain so we are
	// less aggressive with check ticks than in case of signature submission
	// where only one signature is enough.
```

pkg/node/node.go
```diff
@@ -474,34 +595,132 @@ func (n *Node) monitorKeepPublicKeySubmission(
...
+			// operator that published the conflicting key and it is safer to
+			// abandon this keep.
+			logger.Errorf(
+				"member [%x] has submitted conflicting public key for keep [%s]: [%x]",
+				event.SubmittingMember,
+				keepAddress.String(),
+				event.ConflictingPublicKey,
+			)
+			return
+		case <-pubkeyCheckTicker.C:
```
pdyraga, line 624: `const monitorKeepPublicKeySubmissionTimeout = 30 * time.Minute` on the top of the file is no longer used.

pkg/node/node.go
```diff
@@ -241,15 +244,13 @@ func (n *Node) publishSignerPublicKey(
...
-
 	err := n.ethereumChain.SubmitKeepPublicKey(keepAddress, publicKey)
 	if err != nil {
-		close(monitoringAbort)
 		return fmt.Errorf("failed to submit public key: [%v]", err)
 	}
 
+	go n.monitorKeepPublicKeySubmission(keepAddress, publicKey)
+
 	return nil
```
pdyraga, line 254: There is very little happening in this function now - can we inline this code in `GenerateSignerForKeep` above? We can probably even eliminate the `Debugf` log as we have logs on Info level informing about the same.

pkg/node/node.go
lukasz-zimnoch: Already addressed in the next PR: https://github.com/keep-network/keep-ecdsa/pull/616

pkg/node/node.go
```diff
@@ -424,41 +411,161 @@ func (n *Node) publishSignature(
...
+				)
+				return true
+			}
+		case <-ctx.Done():
+			logger.Error(
+				"stop waiting for signature for keep [%s] because "+
+					"[%v] timeout has been exceeded",
+				keepAddress.String(),
+				waitTimeout,
+			)
```
pdyraga, line 475: `Error->Errorf` - it's currently rendered as:
```
2020-11-20T14:42:25.031+0100	ERROR	keep-ecdsa	stop waiting for signature for keep [%s] because [%v] timeout has been exceeded0xe1C664a63BD1D017270aaE9bda45A776bA86a07810m0s
```

Let's also rework the message a little to be less confusing - we are stopping only the current iteration not the whole process. My proposition:
```
signature for keep [%s] has not appeared on the chain after [%v] from submitting it
```

pdyraga:
Scenarios tested:
- happy path, everyone submits public key and signature on time, no transactions are dropped
- signature transactions are dropped from all three clients during the first broadcast attempt
- public key transaction from one of the clients is dropped during the first broadcast attempt

In all those cases clients were able to retry as advertised and confirm transactions on-chain after 12 blocks.

pdyraga (approved):

Issue #612, opened 2020-11-18
Add CLI tool to sign a message with operator key


pdyraga:
Signers may work together to recover BTC from liquidated deposits. One part of the process is to sign a BTC address where their share should go to. Either operator or beneficiary should provide the signature. Given that beneficiary address is in most cases a cold wallet and operator account should not leave the node, we need to provide some tooling allowing to produce a signature without requiring operators to install 3rd party tools. The simplest option is to add a CLI tool to keep-ecdsa client that will allow signing an arbitrary string.

PR #613, opened 2020-11-18
CLI command to sign and verify operator's signature

nkuba:
We added functions to let the operator running the client calculate and verify ethereum signatures. This gives an easy way of confirming messages with the operator's key without a need to pull the key out of the node running the client or installing additional libraries to handle the signing.

It calculates a signature in ethereum `R, S, V` format. The `V` property is needed to recover a public key (and address) from a signature on verification.

Closes: https://github.com/keep-network/keep-ecdsa/issues/612

#### Documentation
```sh
$ ./keep-ecdsa signing ethereum sign --help
NAME:
   keep-ecdsa signing ethereum sign - Sign a message using the operator's key

USAGE:
   keep-ecdsa signing ethereum sign [command options] [message]

DESCRIPTION:
   Calculates an ethereum signature for a given message.
The message is expected to be provided as a string, it is later hashed with SHA-256
and passed to Ethereum ECDSA signing. Signature is calculated in Ethereum specific
format as a hexadecimal string representation of 65-byte {R, S, V} parameters.

It requires an Ethereum key to be provided in an encrypted file. A path to the key file
can be configured in a config file or specified directly with an 'eth-key-file' flag.

The key file is expected to be encrypted with a password provided as KEEP_ETHEREUM_PASSWORD
environment variable.
  
The result is outputted in a common Ethereum signature format:
{
  "address": "<address>",
  "msg": "<content>",
  "sig": "<signature>",
  "version": "2"
}

If 'output-file' flag is set the result will be stored in a specified file path.


OPTIONS:
   --eth-key-file value, -k value  Path to the ethereum key file. If not provided read the path from a config file.
   --output-file value, -o value   Output file for the signature
```

#### Examples

```sh
# read default config
$ KEEP_ETHEREUM_PASSWORD="password" ./keep-ecdsa signing ethereum sign verySecretMessage
{"address":"0x6299496199d99941193fdd2d717ef585f431ea05","msg":"verySecretMessage","sig":"e2e7f8f540b8d0274e78f666113be50b76050e19c73d9999f840cc09201097d562d93d0aa2fe44daaec50aefece8242178758be297e826a79ea6d826e4dd7e0501","version":2}

# read custom config path
$ KEEP_ETHEREUM_PASSWORD="password" ./keep-ecdsa --config ./configs/config.toml signing ethereum sign verySecretMessage
{"address":"0x6299496199d99941193fdd2d717ef585f431ea05","msg":"verySecretMessage","sig":"e2e7f8f540b8d0274e78f666113be50b76050e19c73d9999f840cc09201097d562d93d0aa2fe44daaec50aefece8242178758be297e826a79ea6d826e4dd7e0501","version":2}

# read key file directly
$ KEEP_ETHEREUM_PASSWORD="password" ./keep-ecdsa signing ethereum sign --eth-key-file /ethereum/keystore/93df7c54c41a9d7fb17c1e8039d387a2a924708c verySecretMessage
{"address":"0x93df7c54c41a9d7fb17c1e8039d387a2a924708c","msg":"verySecretMessage","sig":"d472bf16d88a4b516a7c7d846f9b2c96d4911109ed5b3c87fb3de7304ed06e590bd1d30259e1d3ea68572088ebfca825a541e4240ab3520b098b4d8536af606401","version":2}

# output to a file
$ KEEP_ETHEREUM_PASSWORD="password" ./keep-ecdsa signing ethereum sign --output-file signature.json verySecretMessage
output stored to a file: signature.json
```

#### Documentation
```sh
$ ./keep-ecdsa signing ethereum verify --help
NAME:
   keep-ecdsa signing ethereum verify - Verifies a signature

USAGE:
   keep-ecdsa signing ethereum verify [command options] [ethereum-signature]

DESCRIPTION:
   Verifies if a signature was calculated for a message 
by an ethereum account identified by an address. 

It expects a signature to be provided in a common Ethereum signature format:
{
  "address": "<address>",
  "msg": "<content>",
  "sig": "<signature>",
  "version": "2"
}

If 'input-file' flag is set the input will be read from a specified file path.


OPTIONS:
   --input-file value, -i value  Input file with the signature
```

#### Examples

```sh
# valid signature
$ ./keep-ecdsa signing ethereum verify '{"address":"0x6299496199d99941193fdd2d717ef585f431ea05","msg":"verySecretMessage","sig":"e2e7f8f540b8d0274e78f666113be50b76050e19c73d9999f840cc09201097d562d93d0aa2fe44daaec50aefece8242178758be297e826a79ea6d826e4dd7e0501","version":2}'
signature verified correctly, message [verySecretMessage] was signed by [0x6299496199d99941193Fdd2d717ef585F431eA05]

# valid signature
$ ./keep-ecdsa signing ethereum verify --input-file signature.json
signature verified correctly, message [verySecretMessage] was signed by [0x6299496199d99941193Fdd2d717ef585F431eA05]

# invalid signature
$ ./keep-ecdsa signing ethereum verify --input-file invalid-signature.json
signature verification failed: invalid signer
        expected: 0x93df7c54c41A9D7FB17C1E8039d387a2A924708c
        actual:   0x6299496199d99941193Fdd2d717ef585F431eA05
```


nkuba:
Updated the code to handle signature format specified in https://coda.io/d/Operator-Information_d846PmoUIrs/Interim-BTC-Recovery-Process_sutgB#_lu22Q

dimpar:
```
# read default config
$ KEEP_ETHEREUM_PASSWORD="password" ./keep-ecdsa signing ethereum sign verySecretMessage
```
Do you think we can add to the doc file a note about where the default config is?

```
# output to a file
$ KEEP_ETHEREUM_PASSWORD="password" ./keep-ecdsa signing ethereum sign --output-file signature.json verySecretMessage
```
Is it intentional that we will throw an error `signature.json already exist` upon multiple execution of the command above?

cmd/signing_ethereum.go
```diff
@@ -0,0 +1,225 @@
...
+				cli.StringFlag{
+					Name:  "input-file,i",
+					Usage: "Input file with the signature",
+				},
+			},
+		},
+	},
+}
+
+const ethereumSignDescription = `Calculates an ethereum signature for a given message.
```
dimpar, line 60: Do you think we can add tabs in each line so it would look like this:
```
USAGE:
   keep-ecdsa signing ethereum sign [command options] [message]

DESCRIPTION:
   
  Calculates an ethereum signature for a given message.
  The message is expected to be provided as a string, it is later hashed with SHA-256
  and passed to Ethereum ECDSA signing. Signature is calculated in Ethereum specific
  format as a hexadecimal string representation of 65-byte {R, S, V} parameters.

  It requires an Ethereum key to be provided in an encrypted file. A path to the key file
  can be configured in a config file or specified directly with an 'eth-key-file' flag.

  The key file is expected to be encrypted with a password provided as KEEP_ETHEREUM_PASSWORD
  environment variable.
  
  The result is outputted in a common Ethereum signature format:
  {
    "address": "<address>",
    "msg": "<content>",
    "sig": "<signature>",
    "version": "2"
  }

  If 'output-file' flag is set the result will be stored in a specified file path.
```

cmd/signing_ethereum.go
```diff
@@ -0,0 +1,225 @@
...
+	"address": "<address>",
+	"msg": "<content>",
+	"sig": "<signature>",
+	"version": "2"
+}
+
+If 'output-file' flag is set the result will be stored in a specified file path.
+`
+
+const ethereumVerifyDescription = `Verifies if a signature was calculated for a message
```
dimpar, line 82: Same comment as above regarding indentation.

nkuba:
> Do you think we can add to the doc file a note about where the default config is?

It's a flag inherited from the main command https://github.com/keep-network/keep-ecdsa/blob/master/main.go#L53-L58 the information can be obtained by running `./keep-ecdsa  --help`

nkuba:
> Is it intentional that we will throw an error signature.json already exist upon multiple execution of the command above?

It's not needed in this case. I'll update the code.

cmd/signing_ethereum.go
```diff
@@ -0,0 +1,225 @@
...
+	if ethKeyFilePath = c.String("eth-key-file"); len(ethKeyFilePath) > 0 {
+		ethKeyPassword = os.Getenv(config.PasswordEnvVariable)
+	} else {
+		ethereumConfig, err := config.ReadEthereumConfig(c.GlobalString("config"))
+		if err != nil {
+			return fmt.Errorf("failed while reading config file: [%v]", err)
+		}
+
+		ethKeyFilePath = ethereumConfig.Account.KeyFile
+		ethKeyPassword = ethereumConfig.Account.KeyFilePassword
```
dimpar, line 125: Just to double check.. Do we allow passwords to be stored in config file? in `config.toml.SAMPLE` there is no password for eth account

nkuba: No, we don't. We read it from env variable and store in config struct https://github.com/keep-network/keep-ecdsa/blob/3fb44f02715abe806d41eae6fa4417994f8665b3/internal/config/config.go#L94

cmd/signing_ethereum.go
nkuba: I'm not sure about it. We defined a very clean description as a string and how to print it is up to the library we use 🤷

cmd/signing_ethereum.go
nkuba: See previous answer.

cmd/signing_ethereum.go
dimpar: Ok, I missed that part.

cmd/signing_ethereum.go
dimpar: Up to you @nkuba, but ex. if you `man touch help` you would see a desc with indenations:
```
NAME
     touch -- change file access and modification times

DESCRIPTION
     The touch utility sets the modification and access times of files.  If any file does not exist, it is created with default permissions.

     By default, touch changes both modification and access times.  The -a and -m flags may be used to select the access time or the modification time individually.  Selecting both
     is equivalent to the default.  By default, the timestamps are set to the current time.  The -t flag explicitly specifies a different time, and the -r flag specifies to set the
     times those of the specified file.  The -A flag adjusts the values by a specified amount.

     The following options are available:......
```

I think in our case it should be enough to `tab` every line in the Description section to achieve the same layout.

nkuba: If I tab the description there is a mismatch between the first line (3 spaces) and subsequent lines (2 spaces).
```
DESCRIPTION:
   Calculates an ethereum signature for a given message.
  The message is expected to be provided as a string, it is later hashed with SHA-256
  and passed to Ethereum ECDSA signing. Signature is calculated in Ethereum specific
  format as a hexadecimal string representation of 65-byte {R, S, V} parameters.
```

We could overcome the problem and align the lines eventually but I think it would overcomplicate the code. I'm voting for simplicity and leaving it as is.

cmd/signing.go
```diff
@@ -285,3 +261,43 @@ func SignDigest(c *cli.Context) error {
 
 	return nil
 }
+
+func outputData(c *cli.Context, data []byte, overwrite bool) error {
+	if outputFilePath := c.String("output-file"); len(outputFilePath) > 0 {
+		var fileMode os.FileMode
+
+		if !overwrite {
```
dimpar, line 269: I have 2 possible changes in mind:
- flip logic to get rid of `!`
- or pass `fileMode` as a parameter. I am not sure if we need `os.Stat` check at all, since if a file was written with `0444` then it won't be allowed to be overwritten (permission denied error). If it was written with `0644` then its a green light. Unless I'm missing something here.

cmd/signing_ethereum.go
dimpar: Hm, right, there would need to be an empty line between the header and the description to align everything. Sure, let's leave it.

cmd/signing.go
nkuba: I like the second proposal for it's simplicity. I'll update the code.

dimpar:
Tested:
- scenarios described in examples when producing a signature
- scenarios described in examples when verifying a signature

dimpar (approved):

Issue #553, opened 2020-09-22
ECDSA client is a little too aggressive about submitting the transaction signature.


suprnrdy:
During tBTC redemption, the ECDSA Clients submit the transaction signature too soon, leading to failed transactions from the operators `Fail with error 'Not awaiting a signature`.  Two options is to reduce the time between submissions, or have each operator try in the order they are in from membership list. 

<img width="963" alt="Screen Shot 2020-09-22 at 10 17 28 AM" src="https://user-images.githubusercontent.com/1414501/93915963-a7a75000-fcbd-11ea-89b2-65d14a14d13d.png">

<img width="727" alt="Screen Shot 2020-09-22 at 10 24 06 AM" src="https://user-images.githubusercontent.com/1414501/93916250-1684a900-fcbe-11ea-96ea-058236cfc89d.png">

<img width="752" alt="Screen Shot 2020-09-22 at 10 35 11 AM" src="https://user-images.githubusercontent.com/1414501/93917097-5ef09680-fcbf-11ea-9472-3fb801a322fb.png">

pdyraga:
Thanks for reporting this @suprnrdy, this is true and we'll be working on improving gas cost efficiency.

PR #616, opened 2020-11-19
Add signer publication delay

lukasz-zimnoch:
Closes:  #553
Depends on: #610

This PR introduces a delay mechanism that forces signers to wait an amount of time before publishing the signature for a given keep. The value of the delay is the product of the signer index and a constant step (5 minutes). So:

- member 0 doesn't wait and start the publication process immediately
- member 1 waits 5 minutes 
- member 2 waits 10 minutes

This change should prevent signers to publish the signature at the same time. Publishing at the same time is not optimal as only the first transaction succeeds while the second and third revert and burn some gas by the way.

### DIFF
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -29,7 +29,7 @@ var logger = log.Logger("keep-ecdsa")
 
 const (
 	// Determines the delay which should be preserved before retrying
-	// actions within the signing process.
+	// actions within the key generation and signing process.
 	retryDelay = 1 * time.Second
 
 	// Number of blocks which should elapse before confirming
@@ -39,7 +39,7 @@ const (
 	// Used to calculate the publication delay factor for the given signer index
 	// to avoid all signers publishing the same signature for given keep at the
 	// same time.
-	signerPublicationDelayStep = 5 * time.Minute
+	signaturePublicationDelayStep = 5 * time.Minute
 )
 
 // Node holds interfaces to interact with the blockchain and network messages
@@ -320,7 +320,7 @@ func (n *Node) publishSignature(
 	digest [32]byte,
 	signature *ecdsa.Signature,
 ) error {
-	n.waitSignerPublicationDelay(keepAddress)
+	n.waitSignaturePublicationDelay(keepAddress)
 
 	attemptCounter := 0
 	for {
@@ -433,14 +433,17 @@ func (n *Node) publishSignature(
 	}
 }
 
-func (n *Node) waitSignerPublicationDelay(
+// waitSignaturePublicationDelay waits a certain amount of time appropriately
+// for the given signer index to avoid all signers publishing the same signature
+// for given keep at the same time.
+func (n *Node) waitSignaturePublicationDelay(
 	keepAddress common.Address,
 ) {
 	signerIndex, err := n.getSignerIndex(keepAddress)
 	if err != nil {
-		logger.Error(
-			"could not determine signer publication delay for keep [%s]: "+
-				"[%v]; signer publication delay will not be preserved",
+		logger.Errorf(
+			"could not determine signature publication delay for keep [%s]: "+
+				"[%v]; the signature publication will not be delayed",
 			keepAddress.String(),
 			err,
 		)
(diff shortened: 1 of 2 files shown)


pkg/node/node.go
```diff
@@ -27,9 +27,20 @@ import (
 
 var logger = log.Logger("keep-ecdsa")
 
-const monitorKeepPublicKeySubmissionTimeout = 30 * time.Minute
-const retryDelay = 1 * time.Second
-const blockConfirmations = uint64(12)
+const (
+	// Determines the delay which should be preserved before retrying
+	// actions within the signing process.
```
pdyraga, line 32: It is also used in the key generation process, so I think: `the signing process.` -> `the key generation and signing process.`

pkg/node/node.go
```diff
@@ -27,9 +27,20 @@ import (
...
+	retryDelay = 1 * time.Second
+
+	// Number of blocks which should elapse before confirming
+	// the given chain state expectations.
+	blockConfirmations = uint64(12)
+
+	// Used to calculate the publication delay factor for the given signer index
+	// to avoid all signers publishing the same signature for given keep at the
+	// same time.
+	signerPublicationDelayStep = 5 * time.Minute
```
pdyraga, line 42: Maybe `signaturePublication*`?

pkg/node/node.go
```diff
@@ -438,6 +451,58 @@ func (n *Node) publishSignature(
 	}
 }
 
+func (n *Node) waitSignerPublicationDelay(
```
pdyraga, line 454: `waitSignaturePublicationDelay`? We are publishing *signature* so it's *signature* publication delay IMO.

pdyraga, line 454: Can we add simple docs to this function? Mostly what's already said in docs for `signerPublicationDelayStep`.

pkg/node/node.go
```diff
@@ -438,6 +451,58 @@ func (n *Node) publishSignature(
...
+	keepAddress common.Address,
+) {
+	signerIndex, err := n.getSignerIndex(keepAddress)
+	if err != nil {
+		logger.Error(
+			"could not determine signer publication delay for keep [%s]: "+
+				"[%v]; signer publication delay will not be preserved",
```
pdyraga, line 461: Perhaps:
```
could not determine signature publication delay for keep [%s]: [%v]; the signature publication will not be delayed
```

pkg/node/node.go
```diff
@@ -438,6 +451,58 @@ func (n *Node) publishSignature(
...
+		)
+		return
+	}
+
+	// just in case this function is not invoked in the right context
+	if signerIndex < 0 {
+		logger.Error(
+			"could not determine signer publication delay for keep [%s]: "+
+				"[signer index is less than zero]; "+
+				"signer publication delay will not be preserved",
```
pdyraga, line 473: I propose:
```
could not determine signature publication delay for keep [%s], signer index is less than zero; the signature publication will not be delayed
```

pdyraga:
We should now remove two logs saying `signature for keep [...] already submitted`. See what's happening in logs when the client is not the one that is submitting a signature:
```
2020-11-23T11:22:10.222+0100	INFO	tss-lib	party {1,}: signing finished!
2020-11-23T11:22:10.226+0100	INFO	keep-ecdsa	waiting [5m0s] before publishing signature for keep [0xe1C664a63BD1D017270aaE9bda45A776bA86a078]
2020-11-23T11:27:10.432+0100	INFO	keep-ecdsa	confirming on-chain signature submission for keep [0xe1C664a63BD1D017270aaE9bda45A776bA86a078]
2020-11-23T11:27:10.432+0100	INFO	keep-chainutil	waiting for block [17807] to confirm chain state
2020-11-23T11:29:45.889+0100	INFO	keep-ecdsa	signature for keep [0xe1C664a63BD1D017270aaE9bda45A776bA86a078] successfully submitted and confirmed on-chain
2020-11-23T11:29:45.889+0100	INFO	keep-ecdsa	signature for keep [0xe1C664a63BD1D017270aaE9bda45A776bA86a078] already submitted: [50b2c43fd39106bafbba0da34fc430e1f91e3c96ea2acee2bc34119f92b37750]
```

We do not need the last log, the one saying `signature for keep [0xe1C664a63BD1D017270aaE9bda45A776bA86a078] successfully submitted and confirmed on-chain` is enough. See it's emitted in `confirmSignature` function, so for the two cases I mentioned, we should be fine with removing `logger.Info` and leaving there just `return`.

pdyraga:
When testing this PR, I noticed the party ID logged during TSS signing is different from signer index returned from the chain. I do not think it is blocking for this PR as the problem lies elsewhere so I opened an issue to capture it: https://github.com/keep-network/keep-ecdsa/issues/622

pdyraga:
Scenarios tested:
- Happy path, signer with on-chain index 0 submits a signature to the chain, other clients confirm and exit the publication process.
- Signer with on-chain index 0 does not submit a signature to the chain, signer with on-chain index 1 does it after a delay. All signers confirm the signature on the chain and exit the publication process.
- Signer with on-chain index 0 does not submit the signature, all clients get restarted. Once they all appear online again, they start the signing process, publish the signature, confirm it on-chain, and finish the signing process gracefully.

pdyraga:
A nice improvement to the code presented here would be to start the confirmation process if a signature appeared on the chain despite the fact the client is in the "delay publication" state. I do not think it's worth blocking because of that, and we can revisit this once we implement a new interface for event subscriptions (see https://github.com/keep-network/keep-core/issues/491).

Captured it in the issue: https://github.com/keep-network/keep-ecdsa/issues/623

pdyraga (approved):

Issue #559, opened 2020-09-23
Filter out duplicate keep closed events


pdyraga:
It could happen that the client receives a duplicate keep closed event. This situation leads to a terribly-looking error in the logs saying that the keep directory in the storage does not exist:

> could not archive keep to the storage: [could not read directory [...]: [open ..: no such file or directory]]"

And it makes sense because the keep has been already archived.

We should filter out duplicate events and not emit this error.

pdyraga:
Addressed as a part of https://github.com/keep-network/keep-ecdsa/issues/680.

PR #617, opened 2020-11-19
Adding tracking for keep closed events

dimpar:
Closes https://github.com/keep-network/keep-ecdsa/issues/559

In this PR we catch and filter duplicate `KeepClosed` events that log false positive error message.


pdyraga:
Closing without merging, this has been addressed during the work on broader refactoring in https://github.com/keep-network/keep-ecdsa/issues/680.

Issue #577, opened 2020-10-13
Expose metric for tss params pool size


nkuba:
Add current TSS pre parameters pool size to the metrics API.

Shadowfiend:
Just to check, this is for the _current_ pool size, as opposed to the target pool size?

pdyraga:
That's my understanding, the target pool size is not that interesting as the current value.

nkuba:
Correct, it's for the current number of params sitting in the pool.

PR #618, opened 2020-11-20
Expose metric for TSS params pool size

lukasz-zimnoch:
Closes: #577 

This PR adds the TSS pre params pool size to the metrics API.


pdyraga (approved):
Tested and works as advertised.

PR #619, opened 2020-11-20
Updated message hashing for ethereum's signatures

nkuba:
Ethereum is hasing a message for signing with a dedicated hashing function that is adding more information to a signed message:
```js
keccak256("\x19Ethereum Signed Message:\n"${message length}${message})
```

In this PR, we are using ethereum's hashing function to obtain digest to sign.

We also extracted parts of the functions to be able to test them.

Having these changes signatures should be verifiable on https://mycrypto.com/sign-and-verify-message/verify.

### DIFF
--- a/cmd/signing_ethereum.go
+++ b/cmd/signing_ethereum.go
@@ -2,15 +2,16 @@ package cmd
 
 import (
 	"bytes"
-	"crypto/sha256"
-	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 
+	"github.com/ethereum/go-ethereum/accounts"
+	"github.com/ethereum/go-ethereum/accounts/keystore"
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/keep-network/keep-common/pkg/chain/ethereum/ethutil"
 	"github.com/keep-network/keep-ecdsa/internal/config"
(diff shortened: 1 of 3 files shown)


cmd/signing_ethereum.go
```diff
@@ -181,6 +174,37 @@ func EthereumVerify(c *cli.Context) error {
...
+	fmt.Printf(
+		"signature verified correctly, message [%s] was signed by [%s]\n",
+		ethereumSignature.Message,
+		ethereumSignature.Address.Hex(),
+	)
+
+	return nil
+}
+
+func sign(ethereumKey *keystore.Key, message string) (*EthereumSignature, error) {
```
dimpar, line 191: Just a thought.. I noticed `go-ethereum` has functions [around signatures](https://github.com/ethereum/go-ethereum/blob/v1.9.10/accounts/keystore/keystore.go#L292). Do you think it would be useful to use it here and maybe end up with less code?

dimpar:
The signature that this command produces has the following output:
```
{
	"address": "<address>",
	"msg": "<content>",
	"sig": "<signature>",
	"version": 2
}
```
but if you paste it in [mycrypto](https://mycrypto.com/sign-and-verify-message/verify) you would need to manually change from `version: 2` -> `version: "2"` in order to verify the signature. Maybe we can output version in a String format?

nkuba:
I was able to verify a signature where the version was not a string:
```json
{
    "address": "0x4bcfc3099f12c53d01da46695cc8776be584b946",
    "msg": "verySecretMessage",
    "sig": "0xc8be189ab0ee691de7019eaa3de58558b84775085d9a0840908343ac690e02ca3f6e3d2dc70025b9b214d96c30e38c41f818cccd6f06b7a81c4afd26cbe6d6d600",
    "version": 2
}
```
But maybe it's a good idea to make it consistent with what mycrypto produces directly.

dimpar:
> I was able to verify a signature where the version was not a string:
> 
> ```json
> {
>     "address": "0x4bcfc3099f12c53d01da46695cc8776be584b946",
>     "msg": "verySecretMessage",
>     "sig": "0xc8be189ab0ee691de7019eaa3de58558b84775085d9a0840908343ac690e02ca3f6e3d2dc70025b9b214d96c30e38c41f818cccd6f06b7a81c4afd26cbe6d6d600",
>     "version": 2
> }
> ```
> 
> But maybe it's a good idea to make it consistent with what mycrypto produces directly.

Hm, interesting, if I copy & paste it [here](https://mycrypto.com/sign-and-verify-message/verify) the "verify" button is disabled. Only when I add "2" then it becomes enabled.

nkuba:
I was using beta version of the new mycrypto app https://app.mycrypto.com/verify-message.
I'll update the field to string.

cmd/signing_ethereum.go
nkuba: We're using go-ethereum to produce a signature, two functions:
https://github.com/keep-network/keep-ecdsa/blob/529fc792f24554d46542beb61562b1deac530067/cmd/signing_ethereum.go#L192-L194

I was looking for one function that we could use to handle everything but it would require too much effort and we would end up with even more code and complications.

dimpar:
Tested all scenarios similar to what we have in [these](https://github.com/keep-network/keep-ecdsa/pull/613) examples. Everything looked fine.
Verified produced signature against mycrypto and also looked fine.

dimpar (approved):

Issue #474, opened 2020-06-01
Configurable size of safe prime pool


liamzebedee:
https://github.com/keep-network/keep-ecdsa/blob/a9479a2df1713d693dd0995ebf154b34ea0fbe0e/pkg/node/params_pool.go#L24

Running a local ECDSA network of 3 nodes, generating such a poolSize of 20 params on startup slows down my rather well-equipped Macbook to a halt for ~10-15 minutes. 

Is it possible to improve the dev UX here? I've configured the poolSize to `1` locally, and am finding the lazy generation of parameters for deposits is more than satisfactory (~10s). I'm thinking of potential tBTC dapp developers, that might not have such beefy machines, and are running something like [local-setup](https://github.com/keep-network/local-setup).

Shadowfiend:
Hmmm... Probably not crazy to make this overrideable in config or via command line flag... Probably config if it’s likely to be environmental rather than execution based.

liamzebedee:
When would execution-based be suitable @Shadowfiend? Out of curiosity.

mhluongo:
> When would execution-based be suitable

I'm not @Shadowfiend, but typically you want to separate config for difference "instances" of a running application from normal concerns that would come up in a single instance.

nkuba:
The pool size should be overridable in the config file, as this would be consistent with how we handle other runtime parameters.
 https://github.com/keep-network/keep-ecdsa/blob/941216416079c24ab414c9da32ce955aa8dba01d/configs/config.toml.SAMPLE#L77-L82

Having this would be beneficial not only for development environments but also for production where e.g. bootstrap nodes.

PR #621, opened 2020-11-20
Configurable TSS pre-params target pool size

lukasz-zimnoch:
Closes: #474

Added the ability to configure the TSS pre-parameters target pool size.

### DIFF
--- a/configs/config.toml.SAMPLE
+++ b/configs/config.toml.SAMPLE
@@ -89,9 +89,9 @@
 #
 # TSS pre-parameters target pool size. This value determines the number of
 # pre-parameters that should be generated and kept at hand ready to use.
-# That value should be set reasonably because a low pool size may force the
-# client to generate parameters during protocol executions and cause unwanted
-# delays. On the other hand, a big pool size can cause high CPU usage for
+# That value should be set reasonably because a low target pool size may force
+# the client to generate parameters during protocol executions and cause unwanted
+# delays. On the other hand, a big target pool size can cause high CPU usage for
 # a long time. The default value of this parameter is `20`.
 #  PreParamsTargetPoolSize = 20
 
--- a/configs/config.toml.SAMPLE
+++ b/configs/config.toml.SAMPLE
@@ -86,6 +86,14 @@
 # This is an optional parameter, if not provided timeout for TSS protocol
 # pre-parameters generation will be set to `2 minutes`.
 #  PreParamsGenerationTimeout = "2m30s"
+#
+# TSS pre-parameters target pool size. This value determines the number of
+# pre-parameters that should be generated and kept at hand ready to use.
+# That value should be set reasonably because a low pool size may force the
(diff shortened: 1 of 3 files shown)


configs/config.toml.SAMPLE
```diff
@@ -86,6 +86,14 @@
...
 # pre-parameters generation will be set to `2 minutes`.
 #  PreParamsGenerationTimeout = "2m30s"
+#
+# TSS pre-parameters target pool size. This value determines the number of
+# pre-parameters that should be generated and kept at hand ready to use.
+# That value should be set reasonably because a low pool size may force the
+# client to generate parameters during protocol executions and cause unwanted
+# delays. On the other hand, a big pool size can cause high CPU usage for
+# a long time. The default value of this parameter is `20`.
+#  PreParamsTargetPoolSize = 20
```
pdyraga, line 96: Why not just `PreParamsPoolSize`? Is the `*Target*` part needed here? It's a pool, we don't guarantee it's always at the requested level.

lukasz-zimnoch: > It's a pool, we don't guarantee it's always at the requested level.

Exactly, this is why this parameter is the `target` the pool is aiming for. Also, `PreParamsPoolSize` is rather the current size and it's exposed as such in the metrics API: https://github.com/keep-network/keep-ecdsa/pull/618

pdyraga: Yeah, I've just finished the review of #618 and I now understand why you added `*Target*` here. Resolving this conversation.

configs/config.toml.SAMPLE
```diff
@@ -86,6 +86,14 @@
 # This is an optional parameter, if not provided timeout for TSS protocol
 # pre-parameters generation will be set to `2 minutes`.
 #  PreParamsGenerationTimeout = "2m30s"
+#
+# TSS pre-parameters target pool size. This value determines the number of
+# pre-parameters that should be generated and kept at hand ready to use.
+# That value should be set reasonably because a low pool size may force the
```
pdyraga, line 92: "...a low *target* pool size..." perhaps?

configs/config.toml.SAMPLE
```diff
@@ -86,6 +86,14 @@
...
+# client to generate parameters during protocol executions and cause unwanted
+# delays. On the other hand, a big pool size can cause high CPU usage for
```
pdyraga, line 94: "... a big *target* pool size..." perhaps?

pdyraga:
Tested and works as advertised - once the two remaining comments are addressed I am happy to merge.

configs/config.toml.SAMPLE
nkuba: I'm wondering about naming it `PreParamsPoolSizeTarget`... Would be helpful e.g. when greping the code for `PreParamsPoolSize` this parameter would also be found.

pdyraga (approved):

PR #624, opened 2020-11-23
sample config: put TBTCSystem address into quotes

knarz:


pdyraga:
Thanks for fixing it @knarz 🙌

pdyraga (approved):

PR #625, opened 2020-11-24
Staker rewards script - SLA part

lukasz-zimnoch:
Here we introduce the staker rewards script. This PR aims to introduce:
- The structure of the script
- Implementation of the data cache and helpers
- Computations of the SLA part

The script can be invoked by running the following command:
```
ETH_HOSTNAME=<eth-ws-hostname> \
node --experimental-json-modules rewards.js \
<interval-start-unix-timestamp> <interval-end-unix-timestamp>
```

Once invoked, the script validates the passed interval, refreshes the cache, and performs computations of the SLA for each operator. An example output looks as follows:

sla-output

Specific parameters for each operator are calculated based on keeps the operator is a member of, in the following way:
- `keygenCount`: it's the number of keeps whose creation timestamp is within the given interval
- `keygenFailCount`: it's the part of keeps from the `keygenCount` number which have been eventually terminated due to a keygen fail
- `keygenSLA`: it's a parameter calculated as `100 - ((keygenFailCount * 100) / keygenCount` and rounded to the floor
- `signatureCount`: it's the number of keeps which changed their statuses from `active` to `closed ` or from `active` to `terminated` due to signature fail, within the given interval AND have been asked for signing at least once.
- `signatureFailCount`: it's the part of keeps from the `signatureCount` number which changed their statuses from `active` to `terminated` due to signature fail
- `signatureSLA`: it's a parameter calculated as `100 - ((signatureFailCount * 100) / signatureCount` and rounded to the floor


staker-rewards/inspector/scripts/lib/cache.js
```diff
@@ -0,0 +1,212 @@
...
+
+      this.cache
+          .get("keeps")
+          .find({ address: keepData.address })
+          .assign({ status: currentStatus })
+          .write()
+    }
+  }
+
+  async getKeepStatus(keepData) {
```
pdyraga, line 160: Let's document what's the format of the data returned by this function and especially what are timestamp values for closed/terminated/active cases as the way the first two work are a little different than how the third one work, at least to my understanding.

staker-rewards/inspector/scripts/inspect-staker-rewards.js
```diff
@@ -0,0 +1,185 @@
...
+                keep.status.cause === KeepTerminationCause.SIGNATURE_FAIL
+            )
+
+        const allOperators = new Set()
+        cache.getKeeps().forEach(keep => {
+            keep.members.forEach(member => allOperators.add(member))
+        })
+
+        const operatorSummary = []
+
```
pdyraga: Another comment I would like to make is that I think we should refactor the code for SLA computation so that it's possible to unit test it. And we should add javascript tests to `staker-rewards`. We are doing a lot of quite complex computations here and we need unit tests for them.

staker-rewards/inspector/scripts/lib/context.js
```diff
@@ -0,0 +1,72 @@
...
+
+        const deploymentBlock = await getDeploymentBlockNumber(
+            BondedECDSAKeepFactoryJson,
+            web3
+        )
+
+        const contracts = {
+            BondedECDSAKeepFactory: BondedECDSAKeepFactory,
+            BondedECDSAKeep: BondedECDSAKeep,
+            deploymentBlock: deploymentBlock,
```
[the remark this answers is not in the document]
lukasz-zimnoch: I'd vote for the rename because I prefer keeping all contract artifacts imports in one place in `context.js`

staker-rewards/inspector/scripts/lib/cache.js
```diff
@@ -0,0 +1,212 @@
...
+
+    const actions = []
+    keepsOnChain.forEach((keep) => {
+      const currentlyCached = this.cache
+        .get("keeps")
+        .find({ address: keep.address })
+        .value()
+
+      if (!currentlyCached) {
+        actions.push(() => this.fetchKeep(keep))
```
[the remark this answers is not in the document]
lukasz-zimnoch: Now `fetchKeep` fetches also the `creationTimestamp` and will probably get more things in the future as we implement the next parts of the script. I'd rather leave it as it is.

staker-rewards/inspector/scripts/lib/contract-helper.js
```diff
@@ -0,0 +1,87 @@
+import { EthereumHelpers } from "@keep-network/tbtc.js"
+
+const GET_PAST_EVENTS_BLOCK_INTERVAL = 50000
```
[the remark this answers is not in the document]
lukasz-zimnoch: This is not a limit of the event lookback period. It is used only in case we fail to get the events in one request and we switch to the partial events pulls mode. Then we use `GET_PAST_EVENTS_BLOCK_INTERVAL` to determine the blockspan of each batch.

staker-rewards/inspector/scripts/lib/cache.js
lukasz-zimnoch: I've added a comment. To clarify, the timestamp always means the same. It's the moment when the status has been set. Event for the `active` case despite it's calculated differently. The timestamp of the keep's creation block is actually the moment when the keep becomes `active`.

staker-rewards/inspector/scripts/lib/cache.js
```diff
@@ -0,0 +1,258 @@
...
+        eventName,
+        creationBlock
+    )
+
+    if (events.length > 0) {
+      return (await this.web3.eth.getBlock(events[0].blockNumber)).timestamp
+    }
+  }
+
+  async resolveKeepTerminationCause(keepData) {
```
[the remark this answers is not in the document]
lukasz-zimnoch: Yeah, you're right. I think we can go even further and extract also `getKeepStatus`, `getKeepCloseTime`, `getKeepTerminationTime` and `getKeepEventTimestamp`.

staker-rewards/inspector/scripts/inspect-staker-rewards.js
```diff
@@ -0,0 +1,185 @@
...
+            .filter(keep =>
+                keep.status.name === KeepStatus.TERMINATED &&
+                keep.status.cause === KeepTerminationCause.SIGNATURE_FAIL
+            )
+
+        const allOperators = new Set()
+        cache.getKeeps().forEach(keep => {
+            keep.members.forEach(member => allOperators.add(member))
+        })
+
```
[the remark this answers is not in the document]
lukasz-zimnoch: I refactored the code according to your comment and what we discussed.

staker-rewards/lib/cache.js
```diff
@@ -0,0 +1,181 @@
...
+    console.log(
+        `Number of keeps created since block ` +
+        `[${startBlock}]: ${newKeeps.length}`
+    )
+
+    console.log(`Number of keeps in the cache: ${cachedKeepsCount}`)
+
+    const actions = []
+    newKeeps.forEach((keep) => {
+      const currentlyCached = this.cache
```
pdyraga, line 78: `s/currentlyCached/isInCache (isCached)`? 

Keep once cached is cached forever so "currently" does not fit well here IMO.

staker-rewards/lib/cache.js
```diff
@@ -0,0 +1,181 @@
...
+      )
+    }
+
+    const latestBlockNumber = keepCreatedEvents.slice(-1)[0].blockNumber
+    this.cache.assign({ lastRefreshBlock: latestBlockNumber }).write()
+
+    await this.refreshActiveKeeps()
+  }
+
+  async fetchKeep(keepData) {
```
pdyraga, line 110: What do you think about renaming it to `fetchFullKeepInfo` or `fetchFullKeepData`? The loop where we go `newKeeps.forEach((keep) =>` and call this function can be confusing - we have a keep that we are passing a parameter and we are still fetching a keep... Weird. What we are actually doing - we are having *some* keep data and we are fetching the *remaining* keep data.

staker-rewards/lib/cache.js
```diff
@@ -0,0 +1,181 @@
...
+      })
+
+      console.log(
+          `Successfully fetched information about [${results.length}] new keeps`
+      )
+    }
+
+    const latestBlockNumber = keepCreatedEvents.slice(-1)[0].blockNumber
+    this.cache.assign({ lastRefreshBlock: latestBlockNumber }).write()
+
```
pdyraga, line 106: Would it be possible to extract most of the logic above to `await this.fetchNewKeeps()`?  `refresh` function could be a little more elegant this way: `setup logic + fetchNewKeeps() + refreshActiveKeeps()`.

staker-rewards/inspector/scripts/lib/contract-helper.js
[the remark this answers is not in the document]
pdyraga: Correct. And my question is, why `50000` is a good value for this constant. I am not aware of any limitations of this kind other than ones imposed by Alchemy. And if we are doing it so that the script can handle Alchemy Ethereum hosts as well, number `50000` is probably wrong. From the docs I linked in the first comment:

> NOTE: You can make eth_getLogs requests with up to a 2K block range and no limit on the response size. You can also request any block range with a cap of 10K logs in the response.

staker-rewards/rewards.js
```diff
@@ -0,0 +1,128 @@
...
+    try {
+        // URL to the websocket endpoint of the Ethereum node.
+        const ethHostname = process.env.ETH_HOSTNAME
+        // Start of the interval passed as UNIX timestamp.
+        const intervalStart = process.argv[2]
+        // End of the interval passed as UNIX timestamp.
+        const intervalEnd = process.argv[3]
+        // Determines whether debug logs should be disabled.
+        const isDebugDisabled = process.env.DEBUG !== "on"
+        // Determines whether the cache refresh process should be performed.
```
pdyraga, line 16: We should add a warning that this option is only for developers and one should disable cache refresh only when they really know what they are doing. Also, that not performing cache refresh may cause rewards distribution to be calculated based on outdated information from the chain.

staker-rewards/rewards.js
```diff
@@ -0,0 +1,128 @@
...
+        const isCacheRefreshEnabled = process.env.CACHE_REFRESH !== "off"
+
+        if (!ethHostname) {
+            console.error(clc.red("Wrong ETH_HOSTNAME value"))
```
pdyraga, line 20: "Wrong..." or "Please provide..." ?

staker-rewards/rewards.js
```diff
@@ -0,0 +1,128 @@
...
+    for (const operator of getOperators(cache)) {
+        const operatorSLA = slaCalculator.calculateOperatorSLA(operator)
+        // TODO: other computations
+
+        operatorsRewards.push(
+            new OperatorRewards(
+                operator,
+                operatorSLA
+            )
+        )
```
pdyraga, line 91: We can output here all operators but let's keep in mind we should merkelize only non-zero reward transfers (cc @nkuba, @dimpar).

staker-rewards/rewards.js
```diff
@@ -0,0 +1,128 @@
...
+    this.signatureSLA = operatorSLA.signatureSLA
+}
+
+run()
+    .then(result => {
+        console.log(clc.green("Inspection completed successfully"))
+        process.exit(0)
+    })
+    .catch(error => {
+        console.error(clc.red("Inspection errored out with error: "), error)
```
pdyraga, line 126: `s/Inspection/Staker rewards distribution calculations`?

staker-rewards/lib/sla-calculator.js
```diff
@@ -0,0 +1,148 @@
...
+        // Step 2 of signature SLA: get keeps terminated within the
+        // given interval
+        const terminatedKeeps = cache
+            .getKeeps(KeepStatus.TERMINATED) // get terminated keeps
+            .filter(keep => {
+                // get keeps whose statuses have been changed within the
+                // given interval
+                return interval.start <= keep.status.timestamp &&
+                    keep.status.timestamp < interval.end
+            })
```
pdyraga, line 54: Would it be possible to extract
```
            .filter(keep => {
                // get keeps whose statuses have been changed within the
                // given interval
                return interval.start <= keep.status.timestamp &&
                    keep.status.timestamp < interval.end
            })
```
into a function? Even an anonymous function defined as the first line in `initialize`. We are repeating the same code in 3 places.

staker-rewards/lib/sla-calculator.js
```diff
@@ -0,0 +1,148 @@
...
+
+        // Step 3 of signature SLA: Concatenate keeps closed and terminated
+        // within the given interval but exclude the keeps terminated due to
+        // keygen fail as they are not relevant for the signature SLA. This way
+        // we obtain an array of keeps whose statuses have been changed from
+        // `active` to `closed`/`terminated`. Implicitly, this means a keep
```
pdyraga, line 60: > This way we obtain an array of keeps whose statuses have been changed from `active` to `closed` or from `active` to `terminated` *but not as a result of a failed key generation*.

Let's maybe make it more precise ☝️ 

(*Before doing anything here, please take a look at my comment to the code below).

staker-rewards/lib/sla-calculator.js
```diff
@@ -0,0 +1,148 @@
...
+        // - keep has been terminated from another reason not related
+        //   with the signature or keygen context
+        const deactivatedKeeps = [].concat(
+            closedKeeps,
+            terminatedKeeps.filter(
+                // get keeps which have been terminated due to causes
+                // other than keygen fail
+                keep => keep.status.cause !== KeepTerminationCause.KEYGEN_FAIL
+            )
+        )
```
pdyraga, line 73: I am having second thoughts whether we should include keeps with `KeepTerminationCause.OTHER` when calculating signature SLA. Those keeps were never asked for a redemption signature and they are only increasing success rate. Please consider the following scenario: I have 101 tBTC deposits, 100 of them were heavily undercollateralized and were liquidated. I was never asked to provide a redemption signature for them. I have 1 other deposit that was terminated with `KeepTerminationCause.SIGNATURE_FAIL`. What's my SLA?

I think now that we should compute a ratio of `SIGNATURE_FAIL : (CLOSED + SIGNATURE_FAIL)`. WDYT?

staker-rewards/inspector/scripts/lib/contract-helper.js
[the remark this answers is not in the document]
lukasz-zimnoch: `50k` was an arbitrary value chosen to get fewer batches but with a bigger size. I think we can set it to `2k` as well to satisfy Alchemy.

staker-rewards/lib/sla-calculator.js
lukasz-zimnoch: Yeah, agree. I'll get rid of the keeps with the `KeepTerminationCause.OTHER` termination cause.

staker-rewards/inspector/scripts/inspect-staker-rewards.js
lukasz-zimnoch: I have implemented some tests and added them to the Circle CI config.

staker-rewards/README.adoc
```diff
@@ -0,0 +1,33 @@
...
+- Node.js (at least v14.3.0). We recommend to use
+  the https://github.com/nvm-sh/nvm[Node Version Manager] to get the right
+  version with no troubles.
+
+== How to run
+
+1. Make sure you use the right Node.js version:
++
+```
+nvm use 14.3.0
```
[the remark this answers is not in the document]
pdyraga: Addressed in ea72fc17c6fb94f2b5caf56c597c30a76a0f8594

staker-rewards/test/sla-calculator.test.js
```diff
@@ -0,0 +1,134 @@
...
+        }
+
+        const slaCalculator = SLACalculator.initialize(mockCache, interval)
+
+        const operatorSLA = slaCalculator.calculateOperatorSLA(operator)
+
+        assert.equal(operatorSLA.keygenCount, 2)
+        assert.equal(operatorSLA.keygenFailCount, 0)
+        assert.equal(operatorSLA.keygenSLA, 100)
+    })
```
pdyraga, line 32: Can we add a test covering 100% key-gen SLA when one keep is already closed or terminated? It should be possible if we set the interval start at `8000`.

staker-rewards/test/data/test-cache.json
```diff
@@ -0,0 +1,149 @@
...
+        "0x88Ed51f84c21Ae246c23137D090cdF441009D916",
+        "0xb038a8B2975cD5dE392683f5CD4989d3b1E75648"
+      ],
+      "creationBlock": 9,
+      "creationTimestamp": 9000,
+      "status": {
+        "name": "active",
+        "timestamp": 9000
+      }
+    },
```
pdyraga, line 133: Can we add one more keep here where `0xF1De9490Bf7298b5F350cE74332Ad7cf8d5cB181` is not a member? I see we already have one on 3rd position but just one more at the end would make me feel more comfortable.

staker-rewards/test/sla-calculator.test.js
pdyraga: Ah, my bad - I see now this test covers the scenario I mentioned. I misread `20000` vs `2000` and though we are just taking the first two keeps from the cache.

staker-rewards/test/data/test-cache.json
lukasz-zimnoch: Ok done

staker-rewards/package.json
```diff
@@ -0,0 +1,22 @@
+{
```
nkuba, line 1: As we initiate a new NPM package in `staker-rewards` directory can we also configure JS linting? Defining this now will be super handy for following PRs.

staker-rewards/lib/sla-calculator.js
```diff
@@ -0,0 +1,134 @@
...
+
+        // Step 3 of signature SLA: Concatenate keeps closed and terminated
+        // due to signature fail, within the given interval. This way
+        // we obtain an array of keeps whose statuses have been changed from
+        // `active` to `closed` or from `active` to `terminated` due to
+        // signature fails. Implicitly, this means a keep became not active
+        // due to one of the following causes:
+        // - keep has been closed after delivering a signature successfully
+        // - keep has been terminated after not delivering a signature
+        const deactivatedKeeps = [].concat(closedKeeps, signatureFailKeeps)
```
pdyraga, line 68: There is one edge case we missed. Keep can be closed by the application even though it was never asked to provide a signature. In the case of tBTC, this is a funding timeout scenario. For example, if we generate mainnet report from the moment of deploying contracts and look at the operator `0xF1De9490Bf7298b5F350cE74332Ad7cf8d5cB181`, we see this operator has `signatureCount = 159`:

image


It does not seem right as a significant number of `closedKeeps` [are funding time-out keeps](https://allthekeeps.com/operator/0xf1de9490bf7298b5f350ce74332ad7cf8d5cb181) that were never asked for a signature. 

I think the fix should be to add a function to `keep.js` that will let us say if the given keep was ever asked to provide a signature (one occurrence of `SignatureRequested` is enough) and if not, filter out such keeps. For example:

```
const deactivatedAndAskedForSigning = deactivatedKeeps.filter(keep => wasAskedForSignature(keep))
```

lukasz-zimnoch: Ok, done. Unfortunately, this change increases the script execution time significantly. For now, I don't see an easy way to optimize those checks.

staker-rewards/lib/sla-calculator.js
```diff
@@ -0,0 +1,152 @@
...
+    // due to one of the following causes:
+    // - keep has been closed after delivering a signature successfully
+    // - keep has been terminated after not delivering a signature
+    const deactivatedKeeps = [].concat(closedKeeps, signatureFailKeeps)
+
+    // Step 4 of signature SLA: from the deactivated keeps set, get the ones
+    // which have been asked for signing.
+    const deactivatedAndAskedForSigning = []
+    for (const keep of deactivatedKeeps) {
+      if (await wasAskedForSignature(keep, contracts, web3)) {
```
pdyraga, line 81: Could be worth logging `Checking signature requests for keep 0x...` similarly how we do with `Checking current status of keep 0x...`. Right now, there is a lot of work the script is doing in the background but from the user's perspective, it looks if it hanged: there was a lot of `Checking current status of keep 0x...` logged and then nothing is happening.

staker-rewards/lib/sla-calculator.js
```diff
@@ -0,0 +1,152 @@
...
+    // we obtain an array of keeps whose statuses have been changed from
+    // `active` to `closed` or from `active` to `terminated` due to
+    // signature fails. Implicitly, this means a keep became not active
+    // due to one of the following causes:
+    // - keep has been closed after delivering a signature successfully
+    // - keep has been terminated after not delivering a signature
+    const deactivatedKeeps = [].concat(closedKeeps, signatureFailKeeps)
+
+    // Step 4 of signature SLA: from the deactivated keeps set, get the ones
+    // which have been asked for signing.
```
pdyraga, line 78: Nitpick: `... asked for signing at least one time`? We need to be very self-explanatory in this script.

staker-rewards/lib/sla-calculator.js
pdyraga: Yeah, agreed. We have to live with it for now. In the future, we could be caching if a keep was asked for a signature refreshing all cases when the keep wasn't asked. Probably, adding some optimization with interval start/end date and whether there was a chance the signature was requested (e.g. it is not a case for closed keeps before interval start date).

Let's leave this discussion open for now - maybe we can implement some optimization in a separate PR at the very end, once we'll have all other pieces working together.

staker-rewards/test/sla-calculator.test.js
```diff
@@ -0,0 +1,249 @@
...
+
+      const mockContext = createMockContext()
+
+      // Mock the `BondedECDSAKeep` just to make this test pass as it is not
+      // relevant in the context of this scenario.
+      mockContext.contracts.BondedECDSAKeep = {
+        at: () => ({
+          getPastEvents: () => [],
+        }),
+      }
```
pdyraga, line 40: We have two cases: we either mock `getPastEvents` call to return `[]` or `[{}]`. Given that the way how mocking is set up is quite tricky, maybe we can have two functions declared next to `createMockContext` that would accept `context` and mock those scenarios? For example `recordNoSignatureRequests` and `recordSignatureRequestsForAllKeeps`?

pdyraga: I will leave the naming to you, it is not really record-replay mocking, heh.

pdyraga (approved):

PR #627, opened 2020-11-26
Keep rewards distributor

dimpar:
This PR will add a functionality to claim KEEP rewards by the stakers. Rewards distribution is verified using a merkle tree structure. Merkle tree is generated for each interval (ex. every week) and the next step is allocating KEEP rewards where merkle root is the key for that given interval. The end result of a merkle tree generation `.json` will consist of: 
- merkle root which is "assigned" to a specific interval
- list of rewarded accounts with their reward amount, proofs and index in the merkle tree structure

Merkle tree is generated thanks to the Uniswap [merkle-distributor source](https://github.com/Uniswap/merkle-distributor)

Generation of merkle root & proofs:
- `./rewards-merkle-generator.sh --input path/to/file`
- `./rewards-merkle-generator.sh` if no path was provided, default path is used (now it is an example file)


solidity/contracts/ECDSARewardsDistributor.sol
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    function _setClaimed(uint256 index, bytes32 merkleRoot) private {
+        uint256 claimedWordIndex = index / 256;
+        uint256 claimedBitIndex = index % 256;
+        claimedBitMap[merkleRoot][claimedWordIndex] = claimedBitMap[merkleRoot][claimedWordIndex] | (1 << claimedBitIndex);
+    }
+
+    function claim(uint256 index, address account, uint256 amount, bytes32 merkleRoot, bytes32[] calldata merkleProof) external {
+        require(!isClaimed(index, merkleRoot), 'MerkleDistributor: Drop already claimed.');
```
pdyraga, line 80: Maybe `Drop already claimed` -> `Reward already claimed`?

solidity/contracts/ECDSARewardsDistributor.sol
```diff
@@ -0,0 +1,96 @@
...
+        return claimedWord & mask == mask;
+    }
+
+    function _setClaimed(uint256 index, bytes32 merkleRoot) private {
+        uint256 claimedWordIndex = index / 256;
+        uint256 claimedBitIndex = index % 256;
+        claimedBitMap[merkleRoot][claimedWordIndex] = claimedBitMap[merkleRoot][claimedWordIndex] | (1 << claimedBitIndex);
+    }
+
+    function claim(uint256 index, address account, uint256 amount, bytes32 merkleRoot, bytes32[] calldata merkleProof) external {
```
pdyraga, line 79: I believe there is a vulnerability here! We are not checking whether the user-provided `merkleRoot` is defined and what is the allocation for it. I think an adversary can drain funds from this contract only if they provide a valid root and proof for their inputs.

solidity/contracts/ECDSARewardsDistributor.sol
```diff
@@ -0,0 +1,96 @@
...
+        bytes32 node = keccak256(abi.encodePacked(index, account, amount));
+
+        require(MerkleProof.verify(merkleProof, merkleRoot, node), 'MerkleDistributor: Invalid proof.');
+
+        // Mark it claimed and send the token.
+        _setClaimed(index, merkleRoot);
+        require(IERC20(token).transfer(account, amount), 'MerkleDistributor: Transfer failed.');
+
+        // Update KEEP amount for the give merkleRoot
+        merkleRoots[merkleRoot] -= amount;
```
pdyraga, line 92: Let's use `SafeMath`.

pdyraga (changes requested):

pdyraga:
I am fine with having the contract code in `/solidity/contracts` but just a heads up that when we'll be working on the part generating merkle root and scripts automating the process, we should place this code in `/staker-rewards` directory where the code for generating reward distribution numbers lives (see #625).

solidity/contracts/ECDSARewardsDistributor.sol
```diff
@@ -0,0 +1,96 @@
...
+contract ECDSARewardsDistributor is Ownable {
+    using SafeERC20 for KeepToken;
+    using BytesLib for bytes;
+
+    KeepToken public token;
+
+    // This event is triggered whenever a call to #claim succeeds.
+    event Claimed(uint256 index, address account, uint256 amount);
+
+    event MerkleRootAdded(bytes32 merkleRoot);
```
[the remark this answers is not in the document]
dimpar: Sure, added.

solidity/contracts/ECDSARewardsDistributor.sol
```diff
@@ -0,0 +1,96 @@
...
+    /// @param _extraData merkle root (32 bytes)
+    function receiveApproval(
+        address _from,
+        uint256 _amount,
+        address _token,
+        bytes memory _extraData
+    ) public {
+        require(IERC20(_token) == token, "Unsupported token");
+
+        // TODO: check _amount
```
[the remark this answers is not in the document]
dimpar: added here: https://github.com/keep-network/keep-ecdsa/pull/627/commits/ae08a824cf937dfc6f1104520a43a5a51f679c64

solidity/contracts/ECDSARewardsDistributor.sol
```diff
@@ -0,0 +1,96 @@
...
+import "@openzeppelin/contracts/cryptography/MerkleProof.sol";
+
+contract ECDSARewardsDistributor is Ownable {
+    using SafeERC20 for KeepToken;
+    using BytesLib for bytes;
+
+    KeepToken public token;
+
+    // This event is triggered whenever a call to #claim succeeds.
+    event Claimed(uint256 index, address account, uint256 amount);
```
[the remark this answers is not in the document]
dimpar: Done

solidity/contracts/ECDSARewardsDistributor.sol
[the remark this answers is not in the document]
dimpar: yes, yes.. it was just a draft, I was more concerned about the overall approach in the first commits rather than error messages. Removed

solidity/contracts/ECDSARewardsDistributor.sol
```diff
@@ -0,0 +1,96 @@
...
+
+pragma solidity 0.5.17;
+
+import "openzeppelin-solidity/contracts/ownership/Ownable.sol";
+import "@keep-network/keep-core/contracts/utils/BytesLib.sol";
+import "openzeppelin-solidity/contracts/token/ERC20/SafeERC20.sol";
+import "@keep-network/keep-core/contracts/KeepToken.sol";
+import "@openzeppelin/contracts/cryptography/MerkleProof.sol";
+
+contract ECDSARewardsDistributor is Ownable {
```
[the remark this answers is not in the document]
dimpar: Added first chunk of docs mentioning `MerkleDistributor`

solidity/contracts/ECDSARewardsDistributor.sol
```diff
@@ -0,0 +1,96 @@
...
+
+    // This event is triggered whenever a call to #claim succeeds.
+    event Claimed(uint256 index, address account, uint256 amount);
+
+    event MerkleRootAdded(bytes32 merkleRoot);
+
+    // merkleRoot -> total amount for distribution for a given week
+    mapping(bytes32 => uint256) private merkleRoots;
+
+    // This is a packed array of booleans.
```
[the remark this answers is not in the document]
dimpar: Done

solidity/contracts/ECDSARewardsDistributor.sol
dimpar: Sure.

solidity/contracts/ECDSARewardsDistributor.sol
```diff
@@ -0,0 +1,109 @@
...
+pragma solidity 0.5.17;
+
+import "openzeppelin-solidity/contracts/ownership/Ownable.sol";
+import "@keep-network/keep-core/contracts/utils/BytesLib.sol";
+import "openzeppelin-solidity/contracts/token/ERC20/SafeERC20.sol";
+import "@keep-network/keep-core/contracts/KeepToken.sol";
+import "@openzeppelin/contracts/cryptography/MerkleProof.sol";
+
+/// @title ECDSA Rewards distributor
+/// @notice This contract is based on the Uniswap's Merkle Distributor
```
pdyraga, line 24: This part could be `@dev`. `@notice` should explain how this contract works and how to use it.

solidity/contracts/ECDSARewardsDistributor.sol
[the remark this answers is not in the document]
pdyraga: But does the added `require(_amount <= token.balanceOf(_from), "Sender must have enough amount")` makes sense in `receiveApproval`? If the sender does not have enough tokens, `safeTransfer` should fail. I do not think we can do any validation for the amount in this function that wouldn't be redundant. What am I missing?

solidity/contracts/ECDSARewardsDistributor.sol
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    function _setClaimed(uint256 index, bytes32 merkleRoot) private {
+        uint256 claimedWordIndex = index / 256;
+        uint256 claimedBitIndex = index % 256;
+        claimedBitMap[merkleRoot][claimedWordIndex] = claimedBitMap[merkleRoot][claimedWordIndex] | (1 << claimedBitIndex);
+    }
+
+    function claim(uint256 index, address account, uint256 amount, bytes32 merkleRoot, bytes32[] calldata merkleProof) external {
+        require(!isClaimed(index, merkleRoot), 'Rewards already claimed');
```
pdyraga, line 93: `Rewards` or `Reward`?

solidity/contracts/ECDSARewardsDistributor.sol
```diff
@@ -0,0 +1,96 @@
...
+        bytes32 merkleRoot = _extraData.toBytes32();
+
+        merkleRoots[merkleRoot] = _amount;
+
+        emit MerkleRootAdded(merkleRoot);
+    }
+
+    function isClaimed(uint256 index, bytes32 merkleRoot) public view returns (bool) {
+        uint256 claimedWordIndex = index / 256;
+        uint256 claimedBitIndex = index % 256;
```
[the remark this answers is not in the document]
dimpar: as an input I'm expecting similar json like in [here](https://github.com/keep-network/keep-ecdsa/blob/keep-rewards-distributor/staker-rewards/staker-rewards-input.json)

are you suggesting to add an additional validation to filter out 0 KEEP rewards?
or can we maybe generate an input without 0 KEEP rewards at the beginning?

solidity/contracts/ECDSARewardsDistributor.sol
dimpar: I think both work actually https://www.usingenglish.com/forum/threads/27257-reward-vs-rewards depending on how you look at it.

Anyway, I can change it to `Reward`.

solidity/contracts/ECDSARewardsDistributor.sol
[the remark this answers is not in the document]
dimpar: Right, `safeTransferFrom` will do it for us. Removed additional check.

solidity/contracts/ECDSARewardsDistributor.sol
```diff
@@ -0,0 +1,96 @@
...
+
+    KeepToken public token;
+
+    // This event is triggered whenever a call to #claim succeeds.
+    event Claimed(uint256 index, address account, uint256 amount);
+
+    event MerkleRootAdded(bytes32 merkleRoot);
+
+    // merkleRoot -> total amount for distribution for a given week
+    mapping(bytes32 => uint256) private merkleRoots;
```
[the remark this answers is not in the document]
dimpar: Done

solidity/contracts/ECDSARewardsDistributor.sol
dimpar: Done

solidity/contracts/ECDSARewardsDistributor.sol
dimpar: Sure, done.

solidity/contracts/ECDSARewardsDistributor.sol
```diff
@@ -0,0 +1,129 @@
...
+    /// @param _from The original sender of the tokens.
+    /// @param _amount The amount of KEEP tokens to fund.
+    /// @param _token The KEEP token to fund the rewards in.
+    /// @param _extraData Merkle root (32 bytes).
+    function receiveApproval(
+        address _from,
+        uint256 _amount,
+        address _token,
+        bytes memory _extraData
+    ) public {
```
dimpar, line 93: @pdyraga do we want to add `onlyOwner` modifier here?

solidity/contracts/ECDSARewardsDistributor.sol
dimpar: Not sure if I'm following. Merkle root is generated using the list of stakers `<addresses : KEEP_amount>`. If someone will call this function providing a proper `merkleRoot` & `merkleProofs` but invalid `amount`, we would revert with `Invalid proof` error. We also have a check for if the rewards were already claimed for the given staker.

solidity/contracts/ECDSARewardsDistributor.sol
dimpar: https://github.com/keep-network/keep-ecdsa/pull/627/commits/33126b5097a20469d2d00dc0895e35431daf0d0b

solidity/contracts/ECDSARewardsDistributor.sol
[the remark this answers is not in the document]
pdyraga: The script that will generate reward distribution should filter out accounts with 0 KEEP reward. The script that will generate merkle tree shouldn't worry about it. In other words, it is not a problem for this PR.

solidity/contracts/ECDSARewardsDistributor.sol
pdyraga: `merkleRoot` and `merkleProof` are passed by the user. The user can generate any input of their choice including a tree that will include proof of a reward for their account. We do not check whether the merkle tree passed as a parameter has an allocation.

pdyraga: In other words, the fact we have `onlyOwner` on `receiveApproval` does not help if the user can pass any `merkleRoot` and we dot validate if there is an allocation for this root.

dimpar: Good catch! https://github.com/keep-network/keep-ecdsa/pull/627/commits/733e67d023178a7a1ad23f4eb358f8a6851db3a4

solidity/contracts/ECDSARewardsDistributor.sol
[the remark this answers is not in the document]
dimpar: Ok

rewards-merkle-generator.sh
```diff
@@ -0,0 +1,30 @@
...
+WORKDIR=$PWD
+
+printf "${LOG_START}Initializing merkle-distributor submodule...${LOG_END}"
+
+git submodule update --init --recursive --remote --rebase --force
+
+printf "${LOG_START}Installing dependencies...${LOG_END}"
+
+cd "$WORKDIR/include/merkle-distributor"
+yarn
```
pdyraga, line 17: To avoid adding another package management tool to our stack, we can fork `Uniswap/merkle-distributor` and use https://github.com/imsnif/synp to generate `package-lock.json`. I am not yet sure if it's the best strategy, however, the part in #625 uses `npm`, and this part requires `yarn` so requiring having two package managers to generate rewards end-to-end does not feel right. 

We should go either with `yarn` or with `npm` on all levels and since we still have `npm` everywhere else and we are consistent about it, forking and adjusting to `npm` even as an intermediary step does not feel bad.

solidity/contracts/ECDSARewardsDistributor.sol
```diff
@@ -0,0 +1,134 @@
...
+    /// @param _from The original sender of the tokens.
+    /// @param _amount The amount of KEEP tokens to fund.
+    /// @param _token The KEEP token to fund the rewards in.
+    /// @param _extraData Merkle root (32 bytes).
+    function receiveApproval(
+        address _from,
+        uint256 _amount,
+        address _token,
+        bytes memory _extraData
+    ) public onlyOwner {
```
pdyraga, line 98: To avoid `onlyOwner` on `receiveApproval` (it can be problematic because for ERC20 `token.approveAndCall`, `token` is the `msg.sender` in `receiveApproval`), we can add a separate function `function allocate(bytes32 merkleRoot, uint256 amount) public onlyOwner` that will call `token.safeTransferFrom(msg.sender, address(this), amount);`. Then, in the contract calling this function, we'll do:
```
token.approve(address(distributor), amount);
distributor.allocate(merkleRoot, amount);
```

solidity/contracts/ECDSARewardsDistributor.sol
```diff
@@ -0,0 +1,134 @@
...
+///   roots
+contract ECDSARewardsDistributor is Ownable {
+    using SafeERC20 for KeepToken;
+    using BytesLib for bytes;
+    using SafeMath for uint256;
+
+    KeepToken public token;
+
+    // This event is triggered whenever a call to #claim succeeds.
+    event RewardsClaimed(uint256 index, address account, uint256 amount);
```
pdyraga, line 44: We should add `bytes32 merkleRoot` to this event.

solidity/contracts/ECDSARewardsDistributor.sol
```diff
@@ -0,0 +1,134 @@
...
+    // This event is triggered whenever rewards are allocated.
+    event RewardsAllocated(bytes32 merkleRoot, uint256 amount);
+
+    // Merkle root -> total amount for distribution for a given timeframe.
+    mapping(bytes32 => uint256) private merkleRoots;
```
pdyraga, line 49: There is a delicate naming problem with this field, how it is modified in `claim` function, and how it is exposed to the world by `getAllocation`.

If I call `getAllocation`, I expect to receive an original reward allocation for the given merkle root (interval). This is not what I am getting because of `sub` in `claim` function. So one option could be to remove `sub` instruction in `claim`. On the other hand, given that this value is written once, it never changes and does not provide any valuable information to the contract, I think we can get rid of `getAllocation` as it can be read now from `RewardsAllocated` event, update `mapping(bytes32 => uint256) private merkleRoots;` to `mapping(bytes32 => bool) private merkleRoots;`, and update the `require` in `claim` accordingly. What we are really interested in is whether the given `merkleRoot` was blessed by the contract owner, that's all. Also, this change will make weekly allocations a bit cheaper.

dimpar: > I think we can get rid of getAllocation as it can be read now from RewardsAllocated event

but if for some reason you would like to check remaining KEEP for a given `merkleRoot` then how would you do this? I agree regarding `RewardsAllocated` event that carry data on the initial funding for a 'merkleRoot' (interval), but having an additional function that would check remaining KEEP after claiming some rewards might be handy. We can rename it ex. `getRemainingAllocation(bytes32 merkleRoot)`. I might be missing some product rules here, so this is just to double check that we don't need this function.

solidity/contracts/ECDSARewardsDistributor.sol
dimpar: Added.

solidity/contracts/ECDSARewardsDistributor.sol
dimpar: Yes, it is problematic and I experienced that in tests. Changed in 
20f7d19

rewards-merkle-generator.sh
dimpar: I think I was able to simplify it without forking and adding intermediary step. Please see https://github.com/keep-network/keep-ecdsa/pull/627/commits/7cbe29f319e3c854a69696c1cef2f5fa6b12071c

solidity/contracts/ECDSARewardsDistributor.sol
```diff
@@ -0,0 +1,125 @@
...
+        emit RewardsClaimed(index, account, amount, merkleRoot);
+    }
+
+    /// Allocates amount of KEEP for a given merkle root.
+    /// @param merkleRoot Merkle root for a given interval.
+    /// @param amount The amount of KEEP tokens allocated for the merkle root.
+    function allocate(bytes32 merkleRoot, uint256 amount) public onlyOwner {
+        token.safeTransferFrom(msg.sender, address(this), amount);
+
+        merkleRoots[merkleRoot] = amount;
```
dimpar, line 97: @pdyraga are we planning on adding tokens for a given merkle root? As for now, we would assign it once but if there was a mistake and we would want to change this amount, then we would need to call it again with the right value, but the overall tokens on the contract will increase to the new amount. The question is do we want to leave it as is or maybe:
`merkleRoots[merkleRoot] = merkleRoots[merkleRoot].add(amount);`?

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "include/merkle-distributor"]
+	path = include/merkle-distributor
```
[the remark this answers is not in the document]
dimpar: Yes, moved it to /staker-rewards

rewards-merkle-generator.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
dimpar: I moved it to `/staker-rewards`

solidity/contracts/ECDSARewardsDistributor.sol
```diff
@@ -0,0 +1,125 @@
...
+
+    // This event is triggered whenever a call to #claim succeeds.
+    event RewardsClaimed(
+        uint256 index,
+        address account,
+        uint256 amount,
+        bytes32 merkleRoot
+    );
+    // This event is triggered whenever rewards are allocated.
+    event RewardsAllocated(bytes32 merkleRoot, uint256 amount);
```
pdyraga, line 49: We need to standardize whether `merkleRoot` is the first or the last parameter, both in events and in functions.

solidity/contracts/ECDSARewardsDistributor.sol
```diff
@@ -0,0 +1,125 @@
...
+
+    KeepToken public token;
+
+    // This event is triggered whenever a call to #claim succeeds.
+    event RewardsClaimed(
+        uint256 index,
+        address account,
+        uint256 amount,
+        bytes32 merkleRoot
+    );
```
pdyraga, line 47: We should index at least `merkleRoot` and `account` so that stakers can inspect whether they have already withdrawn their rewards in one call. It will also greatly help KEEP Token Dashboard implementation.

solidity/contracts/ECDSARewardsDistributor.sol
pdyraga: From the staker perspective, you need to inspect events anyway to see if you have already claimed rewards for the given interval and it's probably just one call, assuming we index event fields properly.

I do agree, seeing the number of tokens that have not been claimed yet for the given interval without having to inspect events could be handy *for us* in some scenarios but... in the same scenarios, we would have to inspect events anyway to see which stakers haven't claimed their rewards yet so that we can maybe ping them.

Having `getRemainingAllocation` - or maybe even better `unclaimedRewards` given that we have `claim` and `isClaimed` - could help us start the inspection. On the other hand:
- For small stakers, gas costs can be crucial. We have already experienced a situation with extremely high gas prices and the cost of beacon rewards withdrawal to be the same as the reward for some small stakers.
- If we do not count not-yet-claimed rewards here, we do not have to worry whether we should allow adding tokens for a given merkle root ([your comment](https://github.com/keep-network/keep-ecdsa/pull/627#discussion_r534035982)). If we do a mistake, we'll just send more tokens. The issue we are having is that we are duplicating the same information in two places - amounts are already represented in the merkle root and claimed bit map and at the same time, we have a separate counter that is also tracking them.

dimpar: Okay, agree, bottom line is that we can remove tracking of `merkleRoot -> (unclaimed)amount`. 04bf914

solidity/contracts/ECDSARewardsDistributor.sol
dimpar: see: https://github.com/keep-network/keep-ecdsa/pull/627#discussion_r534057824

solidity/contracts/ECDSARewardsDistributor.sol
dimpar: Done

solidity/contracts/ECDSARewardsDistributor.sol
dimpar: Done

.gitignore
```diff
@@ -30,3 +30,4 @@ build/
 configs
 !configs/config.toml.SAMPLE
 external-contracts.js-e
+include/
```
pdyraga, line 33: Is it still needed given that the submodule lives in `staker-rewards/merkle-distributor`?

dimpar: No, just a leftover, I'll remove it

rewards-merkle-generator.sh
pdyraga: I still see the benefit of forking the repo in an unlikely case the original one gets deleted. Also, I am a little worried about whether `npm i` results without a lock file will not get us into unpredictable problems one day. Let's

solidity/contracts/ECDSARewardsDistributor.sol
```diff
@@ -0,0 +1,119 @@
...
+/// @notice This contract can be used by stakers to claim their rewards for
+/// participation in the keep network for operating ECDSA nodes.
+/// @dev This contract is based on the Uniswap's Merkle Distributor
+/// https://github.com/Uniswap/merkle-distributor with some modifications:
+/// - added a map of merkle root keys with the amount of KEEP (value) that will
+///   be allocated for those merkle roots
+/// - added receiveApproval() function that will be called each time to allocate
+///   new KEEP rewards for a given merkle root. Merkle root is going to be generated
+///   regulary (ex. every week) and it is also means that an interval for that
+///   merkle root has passed
```
pdyraga, line 32: Those two bullets are no longer up to date.

solidity/contracts/ECDSARewardsDistributor.sol
```diff
@@ -0,0 +1,119 @@
...
+/// - changed code accordingly to process claimed rewards using a map of merkle
+///   roots
+contract ECDSARewardsDistributor is Ownable {
+    using SafeERC20 for KeepToken;
+    using SafeMath for uint256;
```
pdyraga, line 37: We probably no longer need `SafeMath`.

solidity/contracts/ECDSARewardsDistributor.sol
```diff
@@ -0,0 +1,119 @@
...
+        bytes32 indexed merkleRoot,
+        uint256 index,
+        address indexed account,
+        uint256 amount
+    );
+    // This event is triggered whenever rewards are allocated.
+    event RewardsAllocated(bytes32 merkleRoot, uint256 amount);
+
+    // Map of merkle roots indicating whether a given interval was allocated with
+    // KEEP token rewards.
```
pdyraga, line 52: Could be worth adding one more sentence that each interval is a separate merkle root.

solidity/contracts/ECDSARewardsDistributor.sol
```diff
@@ -0,0 +1,119 @@
...
+    ) external {
+        require(!isClaimed(merkleRoot, index), "Reward already claimed");
+        require(
+            merkleRoots[merkleRoot],
+            "Rewards must be allocated for a given merkle root"
+        );
+
+        // Verify the merkle proof.
+        bytes32 node = keccak256(abi.encodePacked(index, account, amount));
+
```
pdyraga, line 76: Can we drop the empty line between those two? `// Verify the merkle proof` is both the declaration of `bytes32 node` variable and the `require` call.

solidity/contracts/ECDSARewardsDistributor.sol
```diff
@@ -0,0 +1,119 @@
...
+        uint256 index,
+        address account,
+        uint256 amount,
+        bytes32[] calldata merkleProof
+    ) external {
+        require(!isClaimed(merkleRoot, index), "Reward already claimed");
+        require(
+            merkleRoots[merkleRoot],
+            "Rewards must be allocated for a given merkle root"
+        );
```
pdyraga, line 72: Could be worth executing this check before the previous one. Any operation on a non-existing merkle root is not allowed. Looking at `isClaimed` code, we shouldn't get a result that could confuse users for non-existing root but better be safe than sorry.

solidity/contracts/ECDSARewardsDistributor.sol
dimpar: Updated

solidity/contracts/ECDSARewardsDistributor.sol
dimpar: Yup, removed

solidity/contracts/ECDSARewardsDistributor.sol
dimpar: Added.

solidity/contracts/ECDSARewardsDistributor.sol
dimpar: Dropped.

solidity/contracts/ECDSARewardsDistributor.sol
dimpar: Agree.

solidity/test/staker-rewards/MerkleDistributorTest.js
```diff
@@ -0,0 +1,210 @@
...
+      await expectRevert(
+        rewardsDistributor.claim(merkleRoot, index, account, amount, proof),
+        "Reward already claimed"
+      )
+    })
+  })
+
+  async function allocateTokens(merkleRoot, amount) {
+    await keepToken.approve(rewardsDistributor.address, amount)
+    const allocated = await rewardsDistributor.allocate(merkleRoot, amount)
```
pdyraga, line 202: Can we rename `allocated` to `receipt`? This is a pattern we use in other places in the code. `allocated` suggests we are getting a number equal to the number of tokens allocated while we are actually getting TX receipt.

solidity/test/staker-rewards/rewardsData.js
```diff
@@ -0,0 +1,93 @@
+const testValues = {
+  merkleObject0: {
```
pdyraga, line 2: Thoughts on renaming `merkleObject0` and `merkleObject1` to `interval0` and `interval1`? I feel it would make understanding what we are actually allocating in tests a little easier.
```
await allocateTokens(
  testValues.merkleObject0.merkleRoot,
  testValues.merkleObject0.amountToAllocate
)

vs

await allocateTokens(
  testValues.interval0.merkleRoot,
  testValues.interval0.amountToAllocate
)
```

solidity/test/staker-rewards/MerkleDistributorTest.js
```diff
@@ -0,0 +1,210 @@
...
+
+    it("should fail allocating KEEP tokens without prior approval", async () => {
+      await expectRevert(
+        rewardsDistributor.allocate(
+          testValues.merkleObject0.merkleRoot,
+          testValues.merkleObject0.amountToAllocate
+        ),
+        "SafeERC20: low-level call failed"
+      )
+    })
```
pdyraga, line 53: Can we add another test ensuring `RewardsAllocated` event is emitted with correct values?

solidity/test/staker-rewards/MerkleDistributorTest.js
```diff
@@ -0,0 +1,210 @@
...
+
+      expectEvent(claimed, "RewardsClaimed", {
+        merkleRoot,
+        index,
+        account,
+        amount,
+      })
+    })
+
+    it("should successfuly claim rewards and updated contract balance", async () => {
```
pdyraga, line 94: s/updated/update

solidity/test/staker-rewards/MerkleDistributorTest.js
```diff
@@ -0,0 +1,210 @@
...
+        amount,
+        proof
+      )
+
+      expectEvent(claimed, "RewardsClaimed", {
+        merkleRoot,
+        index,
+        account,
+        amount,
+      })
```
pdyraga, line 91: Should we `expect(await keepToken.balanceOf(account)).to.eq.BN(amount)` ?

solidity/test/staker-rewards/MerkleDistributorTest.js
pdyraga: Also, maybe "multiple rewards" to differentiate this test from the previous one where we don't execute `claim` in a loop?

solidity/test/staker-rewards/MerkleDistributorTest.js
```diff
@@ -0,0 +1,210 @@
...
+      let claimedAmounts = new BN(0)
+      for (let i = 0; i < merkle0.claims.length; i++) {
+        const merkleRoot = merkle0.merkleRoot
+        const index = merkle0.claims[i].index
+        const account = merkle0.claims[i].account
+        const amount = merkle0.claims[i].amount
+        const proof = merkle0.claims[i].proof
+
+        const value = new BN(amount)
+        claimedAmounts = claimedAmounts.add(value)
```
pdyraga, line 108: We should be able to use `addn` so that we don't have to declare `value`.

solidity/test/staker-rewards/MerkleDistributorTest.js
```diff
@@ -0,0 +1,210 @@
...
+      let claimedAmounts = new BN(0)
+
+      let merkleRoot = merkle0.merkleRoot
+      let index = merkle0.claims[0].index
+      let account = merkle0.claims[0].account
+      let amount = merkle0.claims[0].amount
+      let proof = merkle0.claims[0].proof
+
+      let value = new BN(amount)
+      claimedAmounts = claimedAmounts.add(value)
```
pdyraga, line 159: We can use `addn` here as well.

rewards-merkle-generator.sh
dimpar: Sure, I forked `Uniswap/merkle-distributor` to `keep-network/merkle-distributor` and added dependecy. 

Regarding potential problems with `npm i` I think we greatly reduced the risk by having a local copy of this project. As long as we don't need to modify `package.json` in our local copy, I think it should work just fine. If we want to be extra careful, we can add a `package-lock.json` as well.

solidity/test/staker-rewards/MerkleDistributorTest.js
```diff
@@ -0,0 +1,210 @@
...
+
+        await rewardsDistributor.claim(
+          merkleRoot,
+          index,
+          account,
+          amount,
+          proof
+        )
+      }
+
```
pdyraga, line 118: Should we `expect(await keepToken.balanceOf(account)).to.eq.BN(amount)` and do the same after the second claim?

solidity/test/staker-rewards/MerkleDistributorTest.js
```diff
@@ -0,0 +1,210 @@
...
+      await rewardsDistributor.claim(merkleRoot, index, account, amount, proof)
+
+      await expectRevert(
+        rewardsDistributor.claim(merkleRoot, index, account, amount, proof),
+        "Reward already claimed"
+      )
+    })
+  })
+
```
pdyraga, line 199: Let's add a test scenario for `isClaimed` as well.

rewards-merkle-generator.sh
pdyraga: Let's add `package-lock.json` as well. Based on my understanding of `npm i` and `npm ci`, we are in trouble as long as we do not have `package-lock.json` or as long as `package.json` versions of dependencies are not fixed. The latter is debatable given there is no guarantee transitive dependencies have all versions of their dependencies fixed.

dimpar: https://github.com/keep-network/merkle-distributor/pull/1

solidity/contracts/ECDSARewardsDistributor.sol
```diff
@@ -0,0 +1,124 @@
...
+///   roots
+contract ECDSARewardsDistributor is Ownable {
+    using SafeERC20 for KeepToken;
+
+    KeepToken public token;
+
+    // This event is triggered whenever a call to #claim succeeds.
+    event RewardsClaimed(
+        bytes32 indexed merkleRoot,
+        uint256 index,
```
pdyraga, line 43: Just in case, I'd add `indexed` to `index` field as well. This may help KEEP Token Dashboard for a scenario when one beneficiary has multiple operators and we are displaying which of them claimed rewards.

solidity/test/staker-rewards/MerkleDistributorTest.js
pdyraga: Also, can we please add a test that will cover `Rewards must be allocated for a given merkle root` scenario?

solidity/test/staker-rewards/MerkleDistributorTest.js
dimpar: Done.

solidity/test/staker-rewards/rewardsData.js
dimpar: Done.

solidity/test/staker-rewards/MerkleDistributorTest.js
dimpar: Done.

solidity/test/staker-rewards/MerkleDistributorTest.js
dimpar: Done.

solidity/test/staker-rewards/MerkleDistributorTest.js
dimpar: Done.

solidity/test/staker-rewards/MerkleDistributorTest.js
dimpar: Done.

solidity/test/staker-rewards/MerkleDistributorTest.js
dimpar: Done.

solidity/test/staker-rewards/MerkleDistributorTest.js
```diff
@@ -0,0 +1,217 @@
...
+      const proof = merkle0.claims[0].proof
+
+      const claimed = await rewardsDistributor.claim(
+        merkleRoot,
+        index,
+        account,
+        amount,
+        proof
+      )
+
```
pdyraga, line 99: Can we add `expect(await keepToken.balanceOf(account)).to.eq.BN(amount)` here as well? We have it in the loop for multiple rewards case, but having this check covered here would be also good so that we have all the assertions covered in this test.

pdyraga:
Tested locally, works as advertised. Four outstanding comments and I am good to merge.

solidity/test/staker-rewards/MerkleDistributorTest.js
dimpar: - For `isClaimed` I've tested scenarios around the same rewards being claimed multiple times for multiple merkle roots. The actual bit masking was tested in the original "merkle-distributor" project

- Covered `Rewards must be allocated for a given merkle root` scenario

solidity/contracts/ECDSARewardsDistributor.sol
dimpar: Done.

solidity/test/staker-rewards/MerkleDistributorTest.js
dimpar: Done.

pdyraga (approved):

PR #628, opened 2020-12-01
Staker rewards script: Calculating operator assets

lukasz-zimnoch:
Depends on: #625 
Depends on: https://github.com/keep-network/tbtc.js/pull/114

This pull request introduces additional parameters to the staker rewards script. Those parameters are related with operator assets and are defined as follows:
- `keepStaked`: it's the value of the operator's `activeStake` checked at the interval end block
- `ethBonded`: it's the sum of all operator bonds active at the interval start block
- `ethUnbonded`: it's the value of the operator's `availableUnbondedValue` checked at the interval start block
- `ethTotal`: it's the sum of `ethBonded` and `ethUnbonded`

### DIFF
--- a/staker-rewards/lib/assets-calculator.js
+++ b/staker-rewards/lib/assets-calculator.js
@@ -43,9 +43,11 @@ export default class AssetsCalculator {
   }
 
   async calculateOperatorAssets(operator) {
-    console.log(
-      `Calculating KEEP and ETH under management for operator ${operator}`
-    )
+    if (process.env.NODE_ENV !== "test") {
+      console.log(
+        `Calculating KEEP and ETH under management for operator ${operator}`
+      )
+    }
 
     const keepStaked = await this.calculateKeepStaked(operator)
     const ethBonded = await this.calculateETHBonded(operator)
--- a/staker-rewards/lib/assets-calculator.js
+++ b/staker-rewards/lib/assets-calculator.js
@@ -22,8 +22,7 @@ export default class AssetsCalculator {
   }
 
   async calculateKeepStaked(operator) {
-    // TODO: Check at interval end block
-    const block = "latest"
+    const block = this.interval.endBlock
 
     return await callWithRetry(
       this.tokenStaking.methods.activeStake(
--- a/staker-rewards/lib/assets-calculator.js
+++ b/staker-rewards/lib/assets-calculator.js
@@ -46,7 +46,7 @@ export default class AssetsCalculator {
     const keepStaked = await this.calculateKeepStaked(operator)
     const ethBonded = await this.calculateETHBonded(operator)
     const ethUnbonded = await this.calculateETHUnbonded(operator)
-    const ethTotal = Math.round((ethBonded + ethUnbonded) * 100) / 100
+    const ethTotal = ethBonded + ethUnbonded
 
     return new OperatorAssets(
       operator,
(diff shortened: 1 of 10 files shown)


staker-rewards/lib/assets-calculator.js
```diff
@@ -0,0 +1,182 @@
...
+    this.bondedECDSAKeepFactory = bondedECDSAKeepFactory
+    this.keepBonding = keepBonding
+    this.sortitionPoolAddress = sortitionPoolAddress
+  }
+
+  static async initialize(context, interval) {
+    const { contracts } = context
+
+    const bondedECDSAKeepFactory = await contracts.BondedECDSAKeepFactory.deployed()
+    const tbtcSystem = await contracts.TBTCSystem.deployed()
```
pdyraga, line 26: Instead of importing `tbtc` as a dependency and calling `TBTCSystem.deployed()`, I'd hardcode `TBTCSystem` address as a `SanctionedApplication` constant in `rewards.js`.

Ideally, `keep-ecdsa/staker-rewards` shouldn't need to know anything about applications other than the client's config file and `staker-rewards` should be application-agnostic. All `staker-rewards` scripts should now are what the sanctioned applications are.

staker-rewards/lib/assets-calculator.js
```diff
@@ -0,0 +1,182 @@
...
+    await assetsCalculator.fetchBondEvents()
+
+    return assetsCalculator
+  }
+
+  async calculateOperatorAssets(operator) {
+    const keepStaked = await this.calculateKeepStaked(operator)
+    const ethBonded = await this.calculateETHBonded(operator)
+    const ethUnbonded = await this.calculateETHUnbonded(operator)
+    const ethTotal = Math.round((ethBonded + ethUnbonded) * 100) / 100
```
pdyraga, line 53: Do we need to round it? We should use the precise number for reward calculation, if possible. We can display them rounded in the table though.

pdyraga: The same comment applies to all other rounding operations in this file.

staker-rewards/rewards.js
```diff
@@ -71,21 +75,39 @@ function validateIntervalTimestamps(interval) {
...
 
 async function calculateOperatorsRewards(context, interval) {
   const { cache } = context
   const slaCalculator = await SLACalculator.initialize(context, interval)
+  const assetsCalculator = await AssetsCalculator.initialize(context, interval)
 
   const operatorsRewards = []
 
   for (const operator of getOperators(cache)) {
+    console.log(`Calculating summary for operator ${operator}`)
```
pdyraga, line 100: I'd rather add a log in `calculateOperatorAssets` saying `Calculating KEEP and ETH under management for operator ${operator}` so that we are precise in logs about what's currently happening. All the other logs that are previously printed inform very precisely what step we are at.

pdyraga (commented):
I've run this script with
```
ETH_HOSTNAME="..." node --experimental-json-modules rewards.js 1600041600 1606821119
```
and all `eth*` values for all operators are 0 in the table rendered.

(EDIT: which is correct because nothing was bonded yet at 1600041600).

staker-rewards/lib/assets-calculator.js
lukasz-zimnoch: Yep good idea. However, I prefer putting the `SanctionedApplication` in the `context.js` file to keep it close to other ABIs imports and avoid importing `reward.js` in `assets-calculator.js`.

pdyraga:
Let's make sure we use adjust `keepStake` value displayed to be `/1e18`, same as `eth*` parameters:
image

pdyraga:
Tested locally with the most recent mainnet data. Works as expected.

pdyraga (approved):

PR #629, opened 2020-12-01
Staker Rewards Script: Frauds Detection

nkuba:
We want to check if an operator committed a fraud. To find that we look for `TokensSlahed` events emitted by `TokenStaking` contract.

Currently slash function from `TokenStaking` contract is used only by ECDSA Keeps. Random Beacon does not use slash, it uses seize. We assume that any slashing event is caused by fraud detected for ECDSA Keep supporting tBTC application.

Events filtering rules have to be revisited if we implement more usage for slash function.

Depends on: https://github.com/keep-network/keep-ecdsa/pull/625


staker-rewards/lib/frauds.js
```diff
@@ -0,0 +1,81 @@
+import clc from "cli-color"
```
lukasz-zimnoch, line 1: What do you think about creating a new class which will contain those functions as we do with `SLACalculator`? We can name it `FraudDetector` or somehow and then we can initialize all the contracts in their `initialize` function to avoid doing this in `isOperatorKeepECDSAFraudulent` and `isKeepECDSATransaction` functions. This way we can also pass the `context` only once, during initialization, which seems more elegant, especially from the unit tests perspective.

staker-rewards/rewards.js
```diff
@@ -77,12 +78,24 @@ function validateIntervalTimestamps(interval) {
...
-function calculateOperatorsRewards(cache, interval) {
+async function calculateOperatorsRewards(context, interval) {
+  const {cache} = context
+
   const slaCalculator = SLACalculator.initialize(cache, interval)
 
   const operatorsRewards = []
 
   for (const operator of getOperators(cache)) {
+    const isFraudulent = await isOperatorKeepECDSAFraudulent(context, operator)
```
lukasz-zimnoch, line 89: I'm wondering if we shouldn't just pass the `isFraudulent` flag as a new column in the `OperatorRewards` table. This way we will have a clean summary of each operator and we can also easily detect the cause if an operator doesn't qualify for the reward.

staker-rewards/lib/context.js
```diff
@@ -4,10 +4,12 @@ import WebsocketSubprovider from "web3-provider-engine/subproviders/websocket.js
 import Subproviders from "@0x/subproviders"
 
 import Cache from "./cache.js"
-import { Contract, getDeploymentBlockNumber } from "./contract-helper.js"
+import {Contract, getDeploymentBlockNumber} from "./contract-helper.js"
 
+import TokenStakingJson from "@keep-network/keep-core/artifacts/TokenStaking.json"
 import BondedECDSAKeepFactoryJson from "@keep-network/keep-ecdsa/artifacts/BondedECDSAKeepFactory.json"
 import BondedECDSAKeepJson from "@keep-network/keep-ecdsa/artifacts/BondedECDSAKeep.json"
+import TBTCSystemJson from "@keep-network/tbtc/artifacts/TBTCSystem.json"
```
lukasz-zimnoch, line 12: Shouldn't we depend on `@keep-network/keep-core` and `@keep-network/tbtc` in the `package.json` file?

staker-rewards/rewards.js
nkuba: Sure, good idea. But do we want to execute `calculateOperatorSLA` for fraudulent operator in case of a fraud?

pdyraga: I think it doesn't hurt to calculate SLA for such operators to display them in the result table. It is not a very time-consuming process for a single operator and we'll have all the information in place in case we need it.

staker-rewards/lib/fraud-detector.js
```diff
@@ -0,0 +1,110 @@
...
+        clc.yellow(
+          `found slashing event for operator [${operator}];` +
+            ` please double check correctness of a calculation result`
+        )
+      )
+    } else {
+      console.debug(
+        `found [${events.length}] slashing events for operator [${operator}]`
+      )
+    }
```
pdyraga, line 60: It is obviously a late discovery but worth considering if those lines are not all we need. As we discussed yesterday:
- In ECDSA keeps, we slash KEEP tokens and seize ETH bonds,
- In the random beacon, we seize KEEP tokens.

Also, the only situation when we slash KEEP tokens in ECDSA keeps is a signature fraud.

Removing the rest of the validation will not only simplify the code but will also decouple staker rewards implementation from tBTC. Other advantages I see:
- Authorizing another application for staking rewards is simple: we add another check regarding the existence of the operator in a sortition pool and make sure signature-fraud-level misbehavior on the application side leads to slashing tokens. This is important because the only place in the code where staker rewards will be aware of tBTC is the sortition pool check. It is quite elegant, especially that the sortition pool is a Keep concept, not an application concept.
- We can test this logic without a lot of mocking, see [here](https://github.com/keep-network/keep-ecdsa/blob/329533d8f877851f0a9a2bf0a21b7032baec928a/staker-rewards/test/sla-calculator.test.js#L19-L37) please.
- Rules are simple: you seriously misbehave, your tokens are slashed - no rewards.

I was thinking whether the validation below does not make it safer but I do not think it does. We do know the only scenario tokens are slashed is the signature fraud and in the unlikely case we sanction another application for staking rewards and that application seizes tokens in case of serious misbehavior, we'll have to implement similar logic as below anyway but only for that application.

staker-rewards/rewards.js
```diff
@@ -77,15 +78,31 @@ function validateIntervalTimestamps(interval) {
...
 
   for (const operator of getOperators(cache)) {
+    const isFraudulent = await fraudDetector.isOperatorKeepECDSAFraudulent(
+      operator
+    )
+
+    if (isFraudulent) {
+      console.log(
+        clc.red(`operator [${operator}] has been fraudulent in Keep ECDSA`)
+      )
```
lukasz-zimnoch, line 95: Do we need this log since we're passing this flag to the output table?

staker-rewards/rewards.js
```diff
@@ -122,7 +140,7 @@ run()
     process.exit(0)
   })
   .catch((error) => {
-    console.error(
+    console.trace(
```
lukasz-zimnoch, line 143: Why are we changing the log level here?

staker-rewards/lib/fraud-detector.js
```diff
@@ -0,0 +1,110 @@
...
+
+  async isKeepECDSATransaction(transactionHash) {
+    const { web3 } = this.context
+
+    const tx = await web3.eth.getTransaction(transactionHash)
+
+    // Check if the called contract is a BondedECDSAKeep.
+    const keepOpenedTimestamp = await this.bondedECDSAKeepFactory.methods
+      .getKeepOpenedTimestamp(tx.to)
+      .call()
```
lukasz-zimnoch, line 92: What do you think about wrapping this call with the `callWithRetry` function like we do in all other places in the script?

staker-rewards/lib/fraud-detector.js
```diff
@@ -0,0 +1,110 @@
...
+
+    const isBondedECDSAKeep = keepOpenedTimestamp > 0
```
lukasz-zimnoch, line 94: Maybe we can check whether `isBondedECDSAKeep` is `true` at this point and return in case it is. This way we can avoid calling the deposit contract.

staker-rewards/lib/fraud-detector.js
```diff
@@ -0,0 +1,110 @@
...
+      }
+    }
+
+    return result
+  }
+
+  async isKeepECDSATransaction(transactionHash) {
+    const { web3 } = this.context
+
+    const tx = await web3.eth.getTransaction(transactionHash)
```
lukasz-zimnoch, line 87: Just a loose thought but maybe we can just decode the function name using the transaction input and contract ABIs? This way we could avoid calls to the ETH node and make the script faster.

staker-rewards/lib/fraud-detector.js
```diff
@@ -0,0 +1,110 @@
...
+  async isOperatorKeepECDSAFraudulent(operator) {
+    // Get all slashing events for the operator.
+    const events = await this.tokenStaking.getPastEvents("TokensSlashed", {
+      fromBlock: this.factoryDeploymentBlock,
+      toBlock: "latest",
+      filter: { operator: operator },
+    })
+
+    if (events.length > 0) {
+      console.log(
```
lukasz-zimnoch, line 50: Can we capitalize all log messages in order to align the output with the rest of the script?

staker-rewards/lib/fraud-detector.js
nkuba: I had similiar thoughts yesterday after our discussion. I'm glad you brought this up. I'll simplify the script.

staker-rewards/rewards.js
nkuba: Outputting a stack trace may help on errors investigation. I wanted to commit it in a separate commit but it got included accidentialy.

staker-rewards/lib/fraud-detector.js
nkuba: We got rid of this part of code.

staker-rewards/lib/fraud-detector.js
nkuba: We got rid of this part of code.

staker-rewards/rewards.js
nkuba: I wanted to bring special attention to this operator. Now it's redundant due to the changes we made. I'll remove it.

staker-rewards/rewards.js
```diff
@@ -77,15 +78,21 @@ function validateIntervalTimestamps(interval) {
 
 async function calculateOperatorsRewards(context, interval) {
   const { cache } = context
+
   const slaCalculator = await SLACalculator.initialize(context, interval)
+  const fraudDetector = await FraudDetector.initialize(context)
```
pdyraga, line 83: Nitpick and definitely non-blocking: given that we are checking fraud before SLA and we put `isFraudulent` before `operatorSLA` in `OperatorRewards`, can we initialize `fraudDetector` before `slaCalculator`?

staker-rewards/rewards.js
pdyraga: Yeah, lesson learned here as well. It's not a log level.

staker-rewards/lib/fraud-detector.js
```diff
@@ -0,0 +1,47 @@
...
+    // on fraud detection. Random Beacon does not use slashing but seizing. We
+    // assume that any slashing event is related to a fraud detection for ECDSA
+    // Keep.
+    // This section has to be revisited in case of implementing additional usage
+    // of slashing function.
+    if (events.length > 0) {
+      console.log(
+        clc.yellow(
+          `Found slashing event for operator [${operator}];` +
+            ` please double check correctness of a calculation result`
```
pdyraga, line 38: I would drop `please double check correctness of a calculation result` part to avoid confusion of other parties running this script. IMO it gives an impression something is potentially wrong. Calculation results should be inspected anyway, and this situation is not exceptional more than e.g. not meeting the SLA or withdrawing bonds.

staker-rewards/lib/fraud-detector.js
```diff
@@ -0,0 +1,47 @@
...
+  static async initialize(context) {
+    const { TokenStaking, factoryDeploymentBlock } = context.contracts
+
+    const tokenStaking = await TokenStaking.deployed()
+
+    return new FraudDetector(context, tokenStaking, factoryDeploymentBlock)
+  }
+
+  async isOperatorFraudulent(operator) {
+    // Get all slashing events for the operator.
```
pdyraga, line 21: Could be worth adding a log:
```
    console.log(`Checking fraudulent activity for operator ${operator}`)
```
so that there is no effect of the script hanging. We log `Checking current status of keep 0x...` and `Checking signature requests for keep 0x` before.

staker-rewards/lib/fraud-detector.js
```diff
@@ -0,0 +1,42 @@
...
+  static async initialize(context) {
+    const { TokenStaking, factoryDeploymentBlock } = context.contracts
+
+    const tokenStaking = await TokenStaking.deployed()
+
+    return new FraudDetector(context, tokenStaking, factoryDeploymentBlock)
+  }
+
+  async isOperatorFraudulent(operator) {
+    console.log(`Checking fraudulent activity for operator: ${operator}`)
```
pdyraga, line 21: We do not put colon for the rest of the logs:
https://github.com/keep-network/keep-ecdsa/blob/78ab836fa005415a47a4388ff2e19b70e750d6c0/staker-rewards/lib/sla-calculator.js#L82

nkuba: Yeah, but maybe we should? I'll remove it for consistency...

pdyraga: (Also, linter is unhappy.)

pdyraga (approved):

pdyraga:
@nkuba Do you mind updating PR description before I hit the merge button? What's in the PR description no longer reflect the implementation.

nkuba:
@pdyraga it's ready.

Issue #602, opened 2020-11-13
Update ECDSA staker rewards to incentivize growing ETH under management


pdyraga:
We need to update allocations to better incentivize growing ETH under management - rather than the churning behavior the current allocation mechanism has yielded. One possible strategy not involving a lot of changes is to adjust ECDSA keep rewards by the lifespan of a keep. From the on-chain perspective, stakers will submit a heartbeat to prove the keep is still active. From the off-chain perspective, to minimize operator's ETH expenditure, we can use application-specific actions introduced in the client to send a heartbeat to `ECDSARewards` contract at a deposit redemption request associated with the given keep.

Agoristen:
If you reward lifespan without taking capital into account, you do not incentivize the growing of ETH supply. Rather you incentivize a new type of abuse.

I've kept my mouth shut about this, because I wanted interval 2 to end before laying out all the details. This strategy would have worked in interval 1 & 2, however if you enable this suggestion it will be far more effective and profitable. 

This is how it works:
1. Deploy minimum 3 nodes. You only need the minimum amount of KEEP in each. 
2. Using a bot, you spam deposit (timeout) transactions by lot size in descending order to fill up available capacity.
3. Assign ETH bond to your 3 nodes.
4. Spam the network with real deposits while monitoring the other nodes

At point 4. if any of the other nodes add capacity, you repeat 2. 

Using this strategy, all real deposits will be sent to your own nodes. With all available capacity used up, you can use a lot size of 0.01 TBTC and therefore tie up very little real capital (ETH). 

Here is a simple cost-analysis on filling up the network:

![node_capacity_by_lotsv2](https://user-images.githubusercontent.com/72700696/99153538-6d4ca480-26a9-11eb-9665-4e9bda222723.png)


The table above shows you what it costs to fill up all available slots for each lot size. My assumption is that it costs 0.25 ETH per deposit (timeout). The cost for filling up all available node capacity is therefore 56.75 ETH. 

Quite frankly, a pittance. 

Depending on the ETH capital you're willing to deploy and the capacity of the network you can select your preferable lot size. If you find 0.01 TBTC to be too costly, the attack may still be viable at 0.1 TBTC and so forth.

Real deposits are of course not free either, but historically it's been very profitable to churn, and more so when each keep is spread across 3 nodes of which you are the sole owner. All it takes is 1 single actor abusing this system. He could make millions of dollars at the expense of every other participant, all while providing zero utility and capacity to the network. 

[...]

PR #631, opened 2020-12-03
Keep rewards distributor beneficiary

pdyraga:
~~Depends on #627~~ 
Refs: https://github.com/keep-network/keep-ecdsa/issues/602

Introducing a new token holder contract allowing the contract owner to allocate rewards to `ECDSARewardsDistributor` in phases. Each phase corresponds to a new reward interval allocated with its own Merkle root used by stakers to withdraw their rewards. The contract can receive funds from any other `PhasedEscrow` with `ECDSARewardsEscrowBeneficiary`.

ECDSA staker rewards have been withdrawn from `ECDSARewards` contract to a `PhasedEscrow` (https://etherscan.io/address/0x973005c57872bd7bffb2157e88a6408d428a0f0a). The team initiated the upgrade in interval 2 and finalized it after interval 2 ended to implement [updates to the rewards mechanism](https://blog.keep.network/a-new-rewards-mechanism-deef3412c3e1). `ECDSARewardsDistributor` introduced in #627 will be allocated each interval with an allocation and Merkle root corresponding to that interval's rewards. The missing building block is `ECDSARewardsDistributorEscrow` introduced in this PR.

`ECDSARewardsDistributorEscrow` should receive all funds from `PhasedEscrow` contract that holds now ECDSA staker KEEP rewards. Then, for each new reward interval, the contract owner will call `allocateInterval(bytes32 merkleRoot, uint256 amount)` function to allocate a new reward interval on `ECDSARewardsDistributor` from the tokens held by the contract.

To withdraw funds from `PhasedEscrow` to `ECDSARewardsDistributorEscrow`, it is enough to deploy `ECDSARewardsEscrowBeneficiary` pointing to `ECDSARewardsDistributorEscrow` and set it as a beneficiary of `PhasedEscrow`. This scenario is exercised in `"funding"` describe in unit tests.


solidity/test/rewards/TestECDSARewardsDistributorEscrow.js
```diff
@@ -0,0 +1,133 @@
...
+
+      const eventList = await rewardsDistributor.getPastEvents(
+        "RewardsAllocated",
+        {
+          fromBlock: 0,
+          toBlock: "latest",
+        }
+      )
+
+      assert.equal(eventList.length, 1, "incorrect number of emitted events")
```
dimpar, line 81: Do you think we can use `expect` here and in all the rest assertions? We mix 2 approaches in this test (assert/expect), but it looks like we can use `expect` only.

`expect(eventList.length).to.equal(1, "incorrect number of emitted events")`

solidity/test/rewards/TestECDSARewardsDistributorEscrow.js
```diff
@@ -0,0 +1,133 @@
...
+        }
+      )
+
+      assert.equal(eventList.length, 2, "incorrect number of emitted events")
+      const event1 = eventList[0].returnValues
+      assert.equal(event1.merkleRoot, merkleRoot)
+      assert.equal(event1.amount, amount)
+      const event2 = eventList[1].returnValues
+      assert.equal(event2.merkleRoot, merkleRoot2)
+      assert.equal(event2.amount, amount2)
```
dimpar, line 109: can we add 
`expect(await token.balanceOf(rewardsDistributor.address)).to.eq.BN(amount.add(amount2))`?

solidity/test/rewards/TestECDSARewardsDistributorEscrow.js
pdyraga: Good point, I'll fix it!

pdyraga: https://github.com/keep-network/keep-ecdsa/pull/631/commits/a37f774f429f1c8c431d9404c87cb809489b95ff

solidity/test/rewards/TestECDSARewardsDistributorEscrow.js
pdyraga: https://github.com/keep-network/keep-ecdsa/pull/631/commits/38fec90fadfc269091d874c5c7dc1b325feec75c

solidity/test/rewards/TestECDSARewardsDistributorEscrow.js
```diff
@@ -0,0 +1,137 @@
...
+        token.address,
+        escrow.address,
+        {from: owner}
+      )
+      await beneficiary.transferOwnership(fundingEscrow.address, {from: owner})
+      await fundingEscrow.setBeneficiary(beneficiary.address, {from: owner})
+
+      await fundingEscrow.withdraw(allTokens, {from: owner})
+      expect(await token.balanceOf(escrow.address)).to.eq.BN(allTokens)
+    })
```
dimpar, line 135: What do you think about continue with a transfer in this test to the point when the funds are in `ECDSARewardsDistributorEscrow` contract? This way we connect all the dots end 2 end.

pdyraga: I am not sure about this one. If we do that, we'll repeat the test from the previous `describe`.

This test ends with `expect(await token.balanceOf(escrow.address)).to.eq.BN(allTokens)` and it is a starting point of `describe("allocateInterval"`.

So maybe what we can do here is:
- switch the order of those two describes
- add check in `allocateInterval`'s `beforeEach` making sure all tokens were transferred (checking the same as this assertion)?

dimpar: Yes, sounds good. My point is just to show the flow of funds from `PhasedEscrow` to `ECDSARewardsDistributorEscrow` in e2e test so when reading you can clearly see what goes where step by step. Switching the order and adding a check will help with understanding the flow.

pdyraga: Makes sense 👍 Please check f973c08ee19f4040b2f97a254786faa59129e130 to see if we achieved this goal.

dimpar: lgtm!

dimpar (dismissed):
I'll leave the final approval / merge to Kuba or Antonio.

solidity/test/rewards/TestECDSARewardsDistributorEscrow.js
```diff
@@ -0,0 +1,143 @@
...
+describe("ECDSARewardsDistributorEscrow", () => {
+  const owner = accounts[0]
+  const thirdParty = accounts[1]
+
+  let token
+  let rewardsDistributor
+  let escrow
+
+  before(async () => {
+    token = await KeepToken.new({from: owner})
```
nkuba, line 25: We assume tokens are minted on `KeepToken` deployment. Can we add a check if `token.balanceOf(owner)` is greater than zero or even equal to a fixed value? 
We could also replace `allTokens` in tests with a fixed value instead of `token.balanceOf(owner)`. If the balance is zero some of the tests can give false positive results.

solidity/contracts/ECDSARewardsDistributorEscrow.sol
```diff
@@ -0,0 +1,42 @@
...
+    {
+        ecdsaRewardsDistributor = _ecdsaRewardsDistributor;
+    }
+
+    function allocateInterval(bytes32 merkleRoot, uint256 amount)
+        external
+        onlyOwner
+    {
+        token.approve(address(ecdsaRewardsDistributor), amount);
+        ecdsaRewardsDistributor.allocate(merkleRoot, amount);
```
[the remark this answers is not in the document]
pdyraga: Correct.

solidity/test/rewards/TestECDSARewardsDistributorEscrow.js
pdyraga: Done in https://github.com/keep-network/keep-ecdsa/pull/631/commits/cef62684f746804ca80a2ea566df7c085c54932d. I used the fixed supply for ECDSA staker rewards.

dimpar (approved):
new changes look good to me

solidity/test/rewards/TestECDSARewardsDistributorEscrow.js
```diff
@@ -0,0 +1,144 @@
...
+  "ECDSARewardsEscrowBeneficiary"
+)
+const ECDSARewardsDistributor = contract.fromArtifact("ECDSARewardsDistributor")
+const ECDSARewardsDistributorEscrow = contract.fromArtifact(
+  "ECDSARewardsDistributorEscrow"
+)
+
+describe("ECDSARewardsDistributorEscrow", () => {
+  const owner = accounts[0]
+  const thirdParty = accounts[1]
```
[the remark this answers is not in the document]
pdyraga: >  Would it require a lot of changes in the current tests?

Two lines :) https://github.com/keep-network/keep-ecdsa/pull/631/commits/e1a52401fcecc9dcfeb9379e313c7c37efdb9237

nkuba (approved):

solidity/contracts/ECDSARewardsDistributorEscrow.sol
```diff
@@ -0,0 +1,42 @@
...
+
+import "@keep-network/keep-core/contracts/PhasedEscrow.sol";
+import "./ECDSARewardsDistributor.sol";
+
+/// @title ECDSARewardsDistributorEscrow
+/// @notice A token holder contract allowing contract owner to allocate rewards
+///         to ECDSARewardsDistributor in phases. Each phase corresponds to
+///         a new reward interval allocated with its own merkle root used by
+///         stakers to withdraw their rewards.
+contract ECDSARewardsDistributorEscrow is PhasedEscrow {
```
nkuba, line 25: Should this contract implement `IStakerRewards`?

pdyraga: The problem lies in `keep-core`. `IStakerRewards declares `receiveApproval` as `external`. And `PhasedEscrow` declares `IStakerRewards` as `public`. This contract inherits from `PhasedEscrow` and has `receiveApproval` as `public`. If I make it inherit `IStakerRewards`, the code will not compile.

This should be fixed in `keep-core` but given the time pressure, other tasks in the queue, and the task it's well-known `receiveApproval` signature I decided to leave it without implementing `IStakerRewards`.

PR #633, opened 2020-12-04
Staker Rewards: Authorizations Checks

nkuba:
In this PR we implement checks for conditions around authorizations.

We want to verify id operator:
- had operator contract authorized at the interval start
- had sortition pool authorized at the interval start
- had deauthorized sortition pool during the interval

For checks at interval start we execute authorization check functions with the block of the interval start.

Deauthorization check was quite challenging as the method doesn't emitt any event, so we have to find transactions calling the function on blockchain. There were two possible solutions: check all transactions made by authroizer block by block or use third-party API to fetch required data. We decided to use Tenderly's API. We are aware that not every user of the script will have access to Tenderly's API, so we will fetch the data and cache them in the repo.

Additional environemnt variable is expected by the script: `TENDERLY_API_KEY`. It is optional, if not provided the script will depend on cached data. If the key is provided the script will call Tenderly API and store fetched data in the cache.

TODO:
- [x] add unit tests

Depends on: https://github.com/keep-network/keep-ecdsa/pull/632/
Refs: #602


staker-rewards/lib/conditions.js
```diff
@@ -0,0 +1,180 @@
...
+
+    if (deauthorizations && deauthorizations.length > 0) {
+      for (let i = 0; i < deauthorizations.length; i++) {
+        const transaction = deauthorizations[i]
+
+        console.debug(`Checking transaction ${transaction.hash}`)
+
+        if (
+          transaction.block_number <= this.interval.startBlock ||
+          transaction.block_number >= this.interval.endBlock
```
pdyraga, line 68: In other scripts we do `start <= interval < end` so that we do not have intervals overlapping on one block. For example, in `sla-calculator.js`:
```
    const isInInterval = (timestamp) =>
      interval.start <= timestamp && timestamp < interval.end
```

staker-rewards/lib/conditions.js
```diff
@@ -0,0 +1,180 @@
...
+        ) {
+          console.debug(
+            `Skipping transaction made in block ${transaction.block_number}`
+          )
+          continue
+        }
+
+        const operator = transaction.decoded_input[0].value
+        const inputSortitionPool = transaction.decoded_input[1].value
```
pdyraga, line 77: ~~Why `inputSortitionPool` and not just `sortitionPool`?~~ OK, I do now. We have `sortitionPoolAddress` in the mix and we want to differentiate them. In this case, maybe `transactionOperator` and `transactionSortitionPool`?

pdyraga (commented):
Waiting for unit tests checkbox from PR description before doing a final pass and testing.

We are missing the following requirements which I assume will be added in another PR:
- If the operator does not have a minimum KEEP stake at interval start, they are not getting any rewards for that interval.
- If the operator has more ETH unbonded than the minimum unbonded value and is not in the sortition pool at interval start, they are not getting any rewards for that interval.
- Unbonded ETH during the interval requirement (still to clarify).

staker-rewards/lib/tenderly.js
```diff
@@ -0,0 +1,43 @@
...
+      `Looking for calls to contract [${contractAddress}] function [${functionSignature}]`
+    )
+
+    const functionSelector = this.web3.eth.abi.encodeFunctionSignature(
+      functionSignature
+    )
+
+    return new Promise((resolve, reject) => {
+      request.get(
+        `https://api.tenderly.co/api/v1/account/thesis/project/keep/transactions?contractId[]=eth:1:${contractAddress}&functionSelector=${functionSelector}`,
```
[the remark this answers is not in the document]
nkuba: We could do that but I'm not sure if we need it. I expect the URL to be the same all the times and we're expected to run tenderly sync.

staker-rewards/lib/conditions.js
nkuba: Awesome point!

nkuba: Do we assume end block of an interval will be a starting block for the following interval? Won't we start a new interval with end block +1?

nkuba: Oh yeah, timestamps are start and end of the interval.

staker-rewards/lib/tenderly.js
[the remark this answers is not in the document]
pdyraga: The point is that anyone having access to Tenderly should be able to run this script without modification of the source code. And who knows, maybe one day we will add support for another API (Graph? Etherscan?) to fetch this information.

What we can do without making script execution more complex on our side: add `tenderlyProjectUrl` (and maybe rename `apiAccessKey` to `tenderlyAccessKey`?) and add a condition that if `tenderlyProjectUrl` is not set, the default value of `https://api.tenderly.co/api/v1/account/thesis/project/keep` is used.

staker-rewards/test/helpers/blockchain.js
```diff
@@ -0,0 +1,37 @@
...
+    (block) => block.blockNumber === blockNumber
+  )
+
+  return (
+    block &&
+    block.contracts.find((contract) => contract.address === contractAddress)
+  )
+}
+
+// inputCheck is expected to be a function executed to match input parameters
```
pdyraga, line 14: A reference to `test-blockchain.json` here would help to understand what the input is matched against.

staker-rewards/test/requirements.test.js
```diff
@@ -0,0 +1,122 @@
...
+              inputs.poolAddress === sortitionPool,
+            false
+          ),
+        }),
+      },
+    }),
+  }
+
+  return context
+}
```
pdyraga, line 59: Non-blocking and let's not work on this right now but I'd consider putting the content of `mock.js` and `blockchain.js` in one file and doing all of this setup there. 

In this scenario, I'd only look at `test-blockchain.js` and if I'd need anything custom, I'd then be required to do the setup above.

staker-rewards/test/requirements.test.js
```diff
@@ -0,0 +1,122 @@
...
+
+    it("discovers authorizations on interval start", async () => {
+      const operator = "0xA000000000000000000000000000000000000001"
+
+      const result = await requirements.checkAuthorizations(operator)
+
+      expect(result).deep.equal(
+        new OperatorAuthorizations(operator, true, true, false)
+      )
+    })
```
pdyraga, line 110: Ideally, we should expand tests to explicitly cover:
- no factory authorized on interval start
- no sortition pool authorized on interval start

staker-rewards/data/cache-transactions.json
```diff
@@ -0,0 +1,43 @@
+{
+  "transactions": [
+    {
+      "hash": "0xc00626a9541466079ea12662ad5356311cbccb2a478a11fcb01c6bc769fc4aa5",
+      "from": "0x13aba80a5bf56720ec44e26c06e88a21f4d0a6d9",
+      "to": "0x27321f84704a599ab740281e285cc4463d89a3d5",
+      "block_number": 11092262,
+      "method": "deauthorizeSortitionPoolContract",
+      "decoded_input": [
```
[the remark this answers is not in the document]
nkuba: If we decide to add another API we can store plain data in the cache and decode it later when we use it.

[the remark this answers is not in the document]
pdyraga: Having a cache per implementation is another option though it could be risky given that we have multiple sources of truth in this case. I'd rather consider implementing a decoder that would turn API-specific data format into a common one, that would land in a cache. Another benefit of this approach is that even if one API gets disabled, you can pump the cache from another API, starting from the moment the previously used API failed on. Not worth wasting time on this discussion though, the current solution is 👌 for what we currently need.

staker-rewards/test/requirements.test.js
pdyraga: LOL :) Yeah, they are reading values from the mock but at least you know that your code is executing the required checks that go to the source of truth (database/cache/chain) instead of doing, for example:
```
if (wasSortitionPoolDeauthorized(operator)) {
  return (true, false, true);
} else {
  return (true, true, false);
}
```

nkuba: I've added a test case. Would it be enough or do you think we should add more?

pdyraga: We are good.

staker-rewards/test/helpers/mock.js
```diff
@@ -0,0 +1,23 @@
...
+        (tx) => tx.to.toLowerCase() === to.toLowerCase() && tx.method === method
+      ),
+  },
+  web3: {
+    utils: {
+      toBN: (value) => new BN(value),
+      fromWei: (value) =>
+        new BN(value).div(new BN("1000000000000000000")).toString(),
+    },
+  },
```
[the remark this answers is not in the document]
pdyraga: Do we still need `web3` function?

staker-rewards/lib/tenderly.js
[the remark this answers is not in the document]
pdyraga: What about this one?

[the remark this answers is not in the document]
nkuba: It was handled in 96e046547164a9e995bb1fa3eafe0fc8157865ab

staker-rewards/test/helpers/mock.js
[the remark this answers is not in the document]
nkuba: Looks like we don't. I'll remove it.

pdyraga (approved):

PR #634, opened 2020-12-07
Staker rewards script: Fetch operators based on ETH deposits

lukasz-zimnoch:
Refs: #602 
Depends on: #632 

The operators list in the staker reward script should be constructed using all unique operators from `UnbondedValueDeposited` events ever emitted. This way all operators who made an ETH deposit are taken into account.


pdyraga:
The code looks good but I'll hold on with testing and merge until #632 and maybe #633 are merged.

pdyraga (approved):

PR #636, opened 2020-12-08
Calling rewards calculation and merkle object generation

dimpar:
Refs: #602 

This PR is connecting rewards calculation and generation of a merkle object for a given interval which consist of:
- merkle root
- account address
- index
- rewards amount (in hex)
- proofs

This is all information that will be needed for a staker to claim their KEEP rewards for a given merkle root (interval).

Example usage:
```
./rewards-merkle-generator.sh -v -w <eth_host> -s <start_interval> -e <end_interval> -r <rewards> -u <tenderly_url> -t <tenderly_token>"
   -v Optional verify flag just to display the results without generation of a merkle tree"
   -h Websocket endpoint of the Ethereum node"
   -s Start of the interval passed as UNIX timestamp"
   -e End of the interval passed as UNIX timestamp"
   -r Total KEEP rewards distributed within the given interval passed as 18-decimals number"
   -u Optional Tenderly API project URL"
   -t Optional access token for Tenderly API used to fetch transactions from the chain"
```

`./rewards-merkle-generator.sh -w wss://eth-mainnet... -s 1605225600 -e 1605430400 -r 3934656000000000000000000`

The output of merkle `.json` file will be written to `staker-rewards/distributor/output-merkle-objects.json` and will look similar to this:
```
"<merkle_root>": {
    "tokenTotal": "<total_amount_allocated_for_this_interval>",
    "claims": {
      "<staker_address_0>": {
        "index": 0,
        "amount": "<staker_rewards_0>",
        "proof": [
          "<proof_0>",
          "<proof_1>",
          "<proof_2>",
        ]
      },
      "<staker_address_1>": {
        "index": 1,
        "amount": "<staker_rewards_1>",
        "proof": [
          "<proof_0>",
          "<proof_1>",
          "<proof_2>",
        ]
      },
    ...
   }
  }
next merkle root..
```


staker-rewards/rewards.js
```diff
@@ -76,27 +78,38 @@ async function run() {
...
       )
-    )
+
+      if (operatorReward.totalRewards != 0) {
+        const rewardsBN = new BigNumber(
+          operatorReward.totalRewards.toFormat(noDecimalPlaces, format)
+        )
+        rewards[operatorReward.operator] = rewardsBN.toString(16) // convert BN to hex
+      }
+    }
```
lukasz-zimnoch, line 110: `operatorReward.totalRewards` is already a `BigNumber` so:
- the if condition should be either `!operatorReward.totalRewards.isEqualTo(new BigNumber(0))` or `operatorReward.totalRewards.isGreaterThan(new BigDecimal(0))`
- the if body can be simplified to `rewards[operatorReward.operator] = operatorReward.totalRewards.integerValue(BigNumber.ROUND_DOWN).toString(16)`

staker-rewards/rewards.js
```diff
@@ -269,6 +282,13 @@ function shortenSummaryValues(summary) {
   return summary
 }
 
+function writeOperatorsRewardsToFile(rewards) {
+  fs.writeFileSync(
+    "./generated-rewards/rewards-input.json",
```
lukasz-zimnoch, line 287: Maybe we could parametrize this via an environment variable with a default value?

staker-rewards/rewards-merkle-generator.sh
```diff
@@ -11,20 +11,59 @@ printf "${LOG_START}Initializing merkle-distributor submodule...${LOG_END}"
...
+then
+   echo "Some or all of the parameters are empty";
+   help
 fi
 
-npm run generate-merkle-root -- --input="$WORKDIR/$REWARDS_INPUT_PATH"
\ No newline at end of file
+printf "${LOG_START}Calculating staker rewards...${LOG_END}"
+
+ETH_HOSTNAME="$eth_host" OUTPUT_MODE="text" node --experimental-json-modules rewards.js "$start" "$end" "$rewards"
```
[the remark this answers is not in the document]
dimpar: Sure, done 4234ad8

staker-rewards/rewards.js
dimpar: Agree, I used `isZero()` for BigNumber validation. 7b46ae8

staker-rewards/rewards.js
dimpar: Good idea e11dd14

staker-rewards/generated-rewards/generate-merkle-root.ts
```diff
@@ -0,0 +1,34 @@
+import { parseBalanceMap } from '../merkle-distributor/src/parse-balance-map'
```
pdyraga, line 1: What do you think about renaming `staker-rewards/generated-rewards` to `staker-rewards/distributor`?

When I saw `generated-rewards`, my first though was that it's the directory where we'll publish interval results.

staker-rewards/generated-rewards/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "staker-rewards",
+  "description": "Staker KEEP rewards",
```
pdyraga, line 3: Let's make it clear it's about generating merkle tree for the distribution.

staker-rewards/rewards-merkle-generator.sh
```diff
@@ -5,26 +5,73 @@ set -e
...
+
+help()
+{
+   echo ""
+   echo "Usage: $0 -h <eth_host> -s <start_interval> -e <end_interval> -r <rewards>"
+   echo -e "\t-h Websocket endpoint of the Ethereum node"
+   echo -e "\t-s Start of the interval passed as UNIX timestamp"
+   echo -e "\t-e End of the interval passed as UNIX timestamp"
+   echo -e "\t-r Total KEEP rewards distributed within the given interval passed as 18-decimals number"
+   exit 1 # Exit script after printing help
```
pdyraga, line 42: We also need optional parameters for `TENDERLY_PROJECT_URL` and `TENDERLY_ACCESS_TOKEN`: 

https://github.com/keep-network/keep-ecdsa/blob/decaa4abd405e9266ade4596ac239ab7f1101621/staker-rewards/rewards.js#L35-L41

staker-rewards/rewards.js
```diff
@@ -117,7 +119,17 @@ async function run() {
            )}
           `.replace(/\n/g, "\t")
       )
-    )
+
+      if (!operatorRewards.totalRewards.isZero()) {
+        rewards[
+          operatorRewards.operator
+        ] = operatorRewards.totalRewards
+          .integerValue(BigNumber.ROUND_DOWN)
```
pdyraga, line 127: Do we need to `ROUND_DOWN`? Not sure about this one (cc @lukasz-zimnoch)

staker-rewards/generated-rewards/generate-merkle-root.ts
dimpar: Sure, done in 3a123ba

staker-rewards/generated-rewards/package.json
dimpar: Done a330599

staker-rewards/rewards.js
lukasz-zimnoch: If we don't `ROUND_DOWN` explicitly, the default rounding mode `ROUND_HALF_UP` is used. This doesn't crop the decimal part as expected which may cause the sum of the total rewards (calculated by the script) to be bigger than the one from the input.

staker-rewards/rewards-merkle-generator.sh
pdyraga: Also, if I set an environment variable `rewards.js` uses but that is not listed in flags of `rewards-merkle-generator-sh.`, e.g. `DEBUG = on`, I assume it's also going to work, correct?

staker-rewards/rewards-merkle-generator.sh
```diff
@@ -5,26 +5,73 @@ set -e
...
 fi
 
-npm run generate-merkle-root -- --input="$WORKDIR/$REWARDS_INPUT_PATH"
\ No newline at end of file
+# default file name
+STAKER_REWARD="distributor/staker-reward-allocation.json"
+
+printf "${LOG_START}Calculating staker rewards...${LOG_END}"
+
+ETH_HOSTNAME="$eth_host" OUTPUT_MODE="text" REWARDS_PATH="$WORKDIR/$STAKER_REWARD" node --experimental-json-modules rewards.js "$start" "$end" "$rewards"
```
pdyraga, line 68: I'd include an option to enable OUTPUT_MODE = "text" and would make it disabled by default so that the presented data are more clear to people validating the result.

staker-rewards/rewards.js
```diff
@@ -117,7 +121,17 @@ async function run() {
...
+          .toString(16) // convert BN to hex
```
pdyraga, line 130: And we write them in hex because otherwise, merkle generator has problems with precision? If so, let's make sure to add a comment about it.

staker-rewards/rewards-merkle-generator.sh
dimpar: Okay, but the way it works now is that "output" merkle object is generated only when `OUTPUT_MODE = "text"`. When it is not set, the merkle object file will NOT be generated. Do we want to leave it this way?

pdyraga: In #642 I proposed the following: let's add option `-v` (verify) that runs the script without `OUTPUT_MODE="text"`, displays result in a human-friendly form as a table and does not generate the merkle tree. When option `-v` is not there, we proceed with `OUTPUT_MODE="text"` and generate merkle object.

staker-rewards/rewards-merkle-generator.sh
dimpar: - Regarding the first comment - done in a03b30
- Yes, if you run ex: 
```
DEBUG="on" ./rewards-merkle-generator.sh -h wss://eth-mainnet.. -s 1605225600 -e 1605430400 -r 3934656000000000000000000
```
in `rewards.js` `process.env.DEBUG` will be set to `on`

staker-rewards/rewards-merkle-generator.sh
dimpar: Added in 5088071

staker-rewards/rewards.js
dimpar: Done in 35f79f9

staker-rewards/rewards-merkle-generator.sh
```diff
@@ -7,24 +7,91 @@ LOG_END='\n\e[0m' # new line + reset color
...
+  help
+fi
+
+printf "${LOG_START}Processing input parameters...${LOG_END}"
+
+while getopts "vh:s:e:r:u:t:" opt
+do
+   case "$opt" in
+      v ) verify=true ;;
+      h ) eth_host="$OPTARG" ;;
```
pdyraga, line 46: Thoughts on renaming this option to `w` (websocket)? `h` is usually reserved for help.

dimpar: Sure, done 37c7db5

lukasz-zimnoch (approved):

PR #637, opened 2020-12-08
Staker rewards script: Check unbonded value withdrawals

lukasz-zimnoch:
Refs: #602
Depends on: #632

So far, the `ETHUnbonded` value for each operator has been determined by calling the `KeepBonding.availableUnbondedValue` method at interval start block. However, that approach ignores the operator's unbonded value withdrawals that occurred within the given interval. This pull request changes this by introducing some logic which takes those withdrawals into account and produces a more accurate result.


staker-rewards/test/data/test-blockchain.json
```diff
@@ -62,6 +62,25 @@
...
+      {
+        "address": "0x39d2aCBCD80d80080541C6eed7e9feBb8127B2Ab",
+        "events": [
+          {
+            "name": "UnbondedValueWithdrawn",
+            "returnValues": {
+              "operator": "0xF1De9490Bf7298b5F350cE74332Ad7cf8d5cB181",
+              "amount": "10000000000000000000"
+            }
+          },
```
pdyraga, line 75: Can we add another event for this operator to cover the fact withdrawals should be summed up?

staker-rewards/lib/assets-calculator.js
```diff
@@ -165,7 +181,25 @@ export default class AssetsCalculator {
...
+    for (const valueWithdrawnEvent of valueWithdrawnEvents) {
+      const withdrawnAmount = new BigNumber(
+        valueWithdrawnEvent.returnValues.amount
+      )
+      weiUnbonded = weiUnbonded.minus(withdrawnAmount)
+    }
+
+    if (weiUnbonded.isLessThan(new BigNumber(0))) {
+      weiUnbonded = new BigNumber(0)
+    }
```
pdyraga, line 200: That's actually more complex than that, unfortunately. If you generate a report for interval `1606928319 1607533119` (~last week) and look at operator `0xacdc9ad09e565170194625911bf9b1a445a9736e`, you'll see that their ETH bonded is `6658.17` and their ETH unbonded is `0.00`. As a result, their ETH locked is `6658.17` and ETH score is `5938.57`.

Now, if you look at the value currently bonded you'll see that operator has `5990.86` ETH and `0.98` ETH is unbonded. So even though, `667` ETH evaporated from the system, this is not reflected in ETH locked and as a result, is not reflected in ETH score.

Similar situation is for operator `0x3a0495aD7EFdAe541455846fFe4a6D3858211C4E` and I recommend looking at `withdrawBond` transactions from the last week and generating the report for that interval to see there is something wrong with ETH score:
https://etherscan.io/tx/0x22b1cae47c1604a38c7e2f0441b0d506b1450d538f9f9a7d076828568294c9ae
https://etherscan.io/tx/0x5da6eb357e9efe221512fc0ce22849ae92531f669d631649b7e103b469d099c0
https://etherscan.io/tx/0xb9474e7e710489b54da29508a8dec8aa54a94da92cb9b061b455ddf98e0d100f
https://etherscan.io/tx/0xb581b8710dc19a05f75841304ca596ac04ca2becfa6a06330e0b1b3e24d85f78
https://etherscan.io/tx/0x49cc929cfdcc13cb02e915fefd7078418a2f9310b150072394177c627e0bfb79
https://etherscan.io/tx/0x992a3bd6c2d43ce5a36037b4fcffdd531e2442326ee68e3ffc9ed02f81ed4745
https://etherscan.io/tx/0x3f2bf89b919f4e0c3e358ba46bf8e0605264e1e3fb644360be93d7daa8e05a49
https://etherscan.io/tx/0xe1cfb5a092315a66fd4b107f639b8af7078ad66d60d06fa6bf38e3f666caa54b
https://etherscan.io/tx/0x0200cc0833ddcec71ef76c636139046488064507cd0220f6b469b540300d67b6

The reason for this discrepancy is that those operators withdrew their unbonded value as soon as some deposits got redeemed. So the flow was ETH_bonded -> ETH_unbonded -> withdrawn within the interval.

What I think we can do here:
- Add a third column `ethWithdrawn` and put there a sum of `UnbondedValueWithdrawn` amounts.
- Calculate `ethTotal` as `ethBonded + ethUnbonded - ethWithdrawn`.
- When working on a `README` for `staker-rewards` be extremely explicit about how this mechanism works (separate PR, I'll probably work on that).

lukasz-zimnoch: Ok, done.

pdyraga:
Generated report for interval `1606928319 - 1607533119`  and compared numbers against transactions referenced [here](https://github.com/keep-network/keep-ecdsa/pull/637#discussion_r539512902). Works as expected.

pdyraga (approved):

PR #638, opened 2020-12-08
Staker Rewards: Minimum stake and unbonded value at interval start checks

nkuba:
In this PR we define two more requirements checks:

>* If the operator does not have a minimum KEEP stake at T_start, they are not getting any rewards for interval T. 
>
>* If the operator has more ETH_unbonded than the minimum unbonded value and is not in the sortition pool at T_start, they are not getting any rewards for interval T.
Note: This point can be summarized basically as: if you are eligible to be in the sortition pool, you have to be in the sortition pool. Operators who are not in the sortition pool because all of their ETH is locked are still getting rewards because that ETH is still under the system’s management.

TODO:
- [x] unit tests 

Depends on: https://github.com/keep-network/keep-ecdsa/pull/633
Refs: #602

### DIFF
--- a/staker-rewards/lib/numbers.js
+++ b/staker-rewards/lib/numbers.js
@@ -1,0 +1,6 @@
+import BigNumber from "bignumber.js"
+
+export const decimalPlaces = 2
+
+export const shorten18Decimals = (value) =>
+  new BigNumber(value).dividedBy(new BigNumber(1e18)).toFixed(decimalPlaces)
--- a/staker-rewards/lib/numbers.js
+++ b/staker-rewards/lib/numbers.js
@@ -1,6 +1,7 @@
 import BigNumber from "bignumber.js"
 
 export const decimalPlaces = 2
+export const noDecimalPlaces = 0
 
 export const shorten18Decimals = (value) =>
   new BigNumber(value).dividedBy(new BigNumber(1e18)).toFixed(decimalPlaces)
--- a/staker-rewards/lib/requirements.js
+++ b/staker-rewards/lib/requirements.js
@@ -239,7 +239,7 @@ export default class Requirements {
 
   // If the operator has at least minimum unbonded value available they have
   // to be registered in the sortition pool. Operators who are not in the sortition
-  // pool because all of their ether is locked are still getting rewards because
+  // pool because all of their ether is bonded are still getting rewards because
   // that ether is still under the system’s management.
   async checkWasInPoolIfRequiredAtIntervalStart(operator) {
     const unbondedValueAtStart = new BigNumber(
--- a/staker-rewards/lib/requirements.js
+++ b/staker-rewards/lib/requirements.js
@@ -174,7 +174,7 @@ export default class Requirements {
       operator
     )
 
-    const minimumUnbondedValueRegisteredAtStart = await this.checkUnbondedValueRegisteredAtIntervalStart(
+    const inPoolIfRequired = await this.checkWasInPoolIfRequiredAtIntervalStart(
       operator
     )
 
@@ -184,7 +184,7 @@ export default class Requirements {
       poolAuthorizedAtStart,
       poolDeauthorizedInInterval,
       minimumStakeAtStart,
-      minimumUnbondedValueRegisteredAtStart
+      inPoolIfRequired
     )
   }
 
@@ -241,7 +241,7 @@ export default class Requirements {
   // to be registered in the sortition pool. Operators who are not in the sortition
   // pool because all of their ether is locked are still getting rewards because
   // that ether is still under the system’s management.
-  async checkUnbondedValueRegisteredAtIntervalStart(operator) {
+  async checkWasInPoolIfRequiredAtIntervalStart(operator) {
     const unbondedValueAtStart = new BigNumber(
       await callWithRetry(
         this.keepBonding.methods.unbondedValue(operator),
@@ -275,12 +275,12 @@ export function OperatorRequirements(
   poolAuthorizedAtStart,
   poolDeauthorizedInInterval,
   minimumStakeAtStart,
-  minimumUnbondedValueRegisteredAtStart
+  wasInPoolIfRequiredAtIntervalStart
 ) {
   ;(this.address = address),
     (this.factoryAuthorizedAtStart = factoryAuthorizedAtStart),
     (this.poolAuthorizedAtStart = poolAuthorizedAtStart),
     (this.minimumStakeAtStart = minimumStakeAtStart),
-    (this.minimumUnbondedValueRegisteredAtStart = minimumUnbondedValueRegisteredAtStart),
+    (this.wasInPoolIfRequiredAtIntervalStart = wasInPoolIfRequiredAtIntervalStart),
     (this.poolDeauthorizedInInterval = poolDeauthorizedInInterval)
 }
--- a/staker-rewards/lib/requirements.js
+++ b/staker-rewards/lib/requirements.js
@@ -275,12 +275,12 @@ export function OperatorRequirements(
   poolAuthorizedAtStart,
   poolDeauthorizedInInterval,
   minimumStakeAtStart,
-  wasInPoolIfRequiredAtIntervalStart
+  wasInPoolIfRequiredAtStart
 ) {
   ;(this.address = address),
     (this.factoryAuthorizedAtStart = factoryAuthorizedAtStart),
     (this.poolAuthorizedAtStart = poolAuthorizedAtStart),
     (this.minimumStakeAtStart = minimumStakeAtStart),
-    (this.wasInPoolIfRequiredAtIntervalStart = wasInPoolIfRequiredAtIntervalStart),
+    (this.wasInPoolIfRequiredAtStart = wasInPoolIfRequiredAtStart),
     (this.poolDeauthorizedInInterval = poolDeauthorizedInInterval)
 }
--- a/staker-rewards/lib/requirements.js
+++ b/staker-rewards/lib/requirements.js
@@ -1,4 +1,6 @@
 import { callWithRetry } from "./contract-helper.js"
+import BigNumber from "bignumber.js"
+import { shorten18Decimals } from "./numbers.js"
 
 export default class Requirements {
   constructor(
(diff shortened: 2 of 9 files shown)


staker-rewards/lib/numbers.js
```diff
@@ -0,0 +1,6 @@
+import BigNumber from "bignumber.js"
+
+export const decimalPlaces = 2
+
+export const shorten18Decimals = (value) =>
+  new BigNumber(value).dividedBy(new BigNumber(1e18)).toFixed(decimalPlaces)
```
pdyraga, line 6: When we are displaying a number in 1e18 format, we should not display any decimals as we do not want to go beyond 1e18 precision.

Please see https://github.com/keep-network/keep-ecdsa/pull/632/commits/e382851b99dc1efa9ec413337ffff1b7083bf48d and https://github.com/keep-network/keep-ecdsa/pull/632/commits/1441b5f6d219071a8f624d190e1792e55742870e.

staker-rewards/lib/requirements.js
```diff
@@ -135,9 +161,34 @@ export default class Requirements {
...
+      factoryAuthorizedAtStart,
+      poolAuthorizedAtStart,
+      poolDeauthorizedInInterval,
+    } = await this.checkAuthorizations(operator)
+
+    const minimumStakeAtStart = await this.checkMinimumStakeAtIntervalStart(
+      operator
+    )
+
+    const minimumUnbondedValueRegisteredAtStart = await this.checkUnbondedValueRegisteredAtIntervalStart(
```
pdyraga, line 177: Thoughts on naming it `inPoolIfRequired`? What we are checking here actually is whether the operator is in the sortition pool if it should be there:
> If the operator has more ETH_unbonded than the minimum unbonded value and is not in the sortition pool at T_start, they are not getting any rewards for interval T.

staker-rewards/lib/requirements.js
```diff
@@ -179,16 +229,58 @@ export default class Requirements {
...
+  async checkMinimumStakeAtIntervalStart(operator) {
+    return await callWithRetry(
+      this.bondedECDSAKeepFactory.methods.hasMinimumStake(operator),
+      this.interval.startBlock
+    )
+  }
+
+  // If the operator has at least minimum unbonded value available they have
+  // to be registered in the sortition pool. Operators who are not in the sortition
+  // pool because all of their ether is locked are still getting rewards because
```
pdyraga, line 242: Let's change "all of their ether is locked" to "all of their ether is bonded" because we define ETH_locked as ETH_bonded + ETH_unbonded in other places in staker-rewards scripts.

staker-rewards/lib/requirements.js
```diff
@@ -179,16 +229,58 @@ export default class Requirements {
...
+  // that ether is still under the system’s management.
+  async checkUnbondedValueRegisteredAtIntervalStart(operator) {
```
pdyraga, line 244: `checkIsInPoolIfRequiredAtIntervalStart`?

staker-rewards/rewards.js
```diff
@@ -227,11 +229,15 @@ function OperatorSummary(operator, operatorParameters, operatorRewards) {
...
+      operatorParameters.requirements.factoryAuthorizedAtStart),
     (this.poolAuthorizedAtStart =
-      operatorParameters.authorizations.poolAuthorizedAtStart),
+      operatorParameters.requirements.poolAuthorizedAtStart),
     (this.poolDeauthorizedInInterval =
-      operatorParameters.authorizations.poolDeauthorizedInInterval),
+      operatorParameters.requirements.poolDeauthorizedInInterval),
+    (this.minimumStakeAtStart =
+      operatorParameters.requirements.minimumStakeAtStart),
+    (this.minimumUnbondedValueRegisteredAtStart =
```
pdyraga, line 239: `isInPoolIfRequired`

staker-rewards/lib/numbers.js
nkuba: I'm not sure I follow. It's just a function extracted from `rewards.js`. It's in this form in master:
https://github.com/keep-network/keep-ecdsa/blob/985d0749885087894a385562f30df51c9da7da1f/staker-rewards/rewards.js#L257-L258

staker-rewards/lib/requirements.js
nkuba: Yeah, I didn't like the current name either.

staker-rewards/lib/numbers.js
nkuba: This function will be used to convert wei to ether so decimals are very much required.

pdyraga: I think I missed how this function is used. I will resolve the comment for now and go through the code again.

staker-rewards/rewards.js
```diff
@@ -231,11 +236,15 @@ function OperatorSummary(operator, operatorParameters, operatorRewards) {
...
     (this.poolAuthorizedAtStart =
-      operatorParameters.authorizations.poolAuthorizedAtStart),
+      operatorParameters.requirements.poolAuthorizedAtStart),
     (this.poolDeauthorizedInInterval =
-      operatorParameters.authorizations.poolDeauthorizedInInterval),
+      operatorParameters.requirements.poolDeauthorizedInInterval),
+    (this.minimumStakeAtStart =
+      operatorParameters.requirements.minimumStakeAtStart),
+    (this.wasInPoolIfRequiredAtIntervalStart =
+      operatorParameters.requirements.wasInPoolIfRequiredAtIntervalStart),
```
pdyraga, line 247: Two comments:
1. For other column headers we use `*AtStart` format:`factoryAuthorizedAtStart`, `poolAuthorizedAtStart`, `minimumStakeAtStart`. Let's use it also here (`*AtIntervalStart` -> `AtStart`) to keep it consistent.
2. I think the name `wasInPoolIfRequiredAtIntervalStart` is confusing when the table is rendered. I generated the report for interval `1606928319 1607533119` and saw that operator:
    ```
     ┌─────────┬──────────────────────────────────────────────┬──────────────┬──────────────────────────┬───────────────────────┬────────────────────────────┬─────────────────────┬────────────────────────────────────┬─────────────┬─────────────────┬───────────┬────────────────┬────────────────────┬──────────────┬──────────────┬───────────┬─────────────┬───────────┬───────────┬─────────┬──────────────┬──────────────┐
     │ (index) │                   operator                   │ isFraudulent │ factoryAuthorizedAtStart │ poolAuthorizedAtStart │ poolDeauthorizedInInterval │ minimumStakeAtStart │ wasInPoolIfRequiredAtIntervalStart │ keygenCount │ keygenFailCount │ keygenSLA │ signatureCount │ signatureFailCount │ signatureSLA │  keepStaked  │ ethBonded │ ethUnbonded │ ethTotal  │ ethScore  │  boost  │ rewardWeight │ totalRewards │
     ├─────────┼──────────────────────────────────────────────┼──────────────┼──────────────────────────┼───────────────────────┼────────────────────────────┼─────────────────────┼────────────────────────────────────┼─────────────┼─────────────────┼───────────┼────────────────┼────────────────────┼──────────────┼──────────────┼───────────┼─────────────┼───────────┼───────────┼─────────┼──────────────┼──────────────┤
     (...)
     │   13    │ '0xc74D5EDAf50C15f5C64683D9CC6644121a4923f9' │    false     │          false           │         false         │           false            │        false        │                true                │      0      │        0        │   'N/A'   │       0        │         0          │    'N/A'     │    '0.00'    │  '0.00'   │   '0.00'    │  '0.00'   │  '0.00'   │ '1.00'  │    '0.00'    │    '0.00'    │
     ``` 
    My first thought after seeing `wasInPoolIfRequiredAtIntervalStart = true` was that we have a bug in the code of `staker-rewards` scripts as it is not possible that someone who has not authorized ECDSA operator contract is in a sortition pool. Then, after a while, I realized that the operator was not required to be there and that's why we have `true` in this column. However, I could realize that only because I know this code very well...

   I think we should rename this column to emphasize the logical alternative: "in pool OR not required to be in the pool". The only option I could come up with that is `inPoolOrNotRequiredInPoolAtStart` but maybe you can find a better one.

staker-rewards/lib/numbers.js
```diff
@@ -0,0 +1,7 @@
+import BigNumber from "bignumber.js"
+
+export const decimalPlaces = 2
+export const noDecimalPlaces = 0
+
+export const shorten18Decimals = (value) =>
```
pdyraga, line 6: Not related to this PR but... can we take the opportunity and use `shorten18Decimals` to fix the value displayed at the beginning of the script execution:
```
OUTPUT_MODE="text" ETH_HOSTNAME="..." node --experimental-json-modules rewards.js 1606928319 1607533119 3934656000000000000000000
(...)
Interval total rewards: 3934656000000000000000000 KEEP
```
?

(should be `3934656 KEEP`)

staker-rewards/test/requirements.test.js
```diff
@@ -103,38 +138,137 @@ describe("requirements", async () => {
...
-        new OperatorAuthorizations(operator, true, true, false)
-      )
+      expect(result).deep.equal({
+        factoryAuthorizedAtStart: true,
+        poolAuthorizedAtStart: true,
+        poolDeauthorizedInInterval: false,
+      })
     })
 
     it("finds if deauthorized during interval", async () => {
```
pdyraga, line 153: Something I probably missed during the previous review: can we rename this test to `finds if the pool was unauthorized during interval`? The only thing that could get deauthorized is pool, thus the only value that could change is `poolDeauthorizedInInterval`.

staker-rewards/test/requirements.test.js
```diff
@@ -103,38 +138,137 @@ describe("requirements", async () => {
...
+    //  | operator | unbondedValue | isOperatorRegistered
+    //  |  1       |  < minimum    |  false
+    //  |  2       |  = minimum    |  true
+    //  |  3       |  = minimum    |  false (*)
+    //  |  4       |  > minimum    |  true
+    //  |  5       |  > minimum    |  false
+    //  |  6       |  2000 ether   |  false
+    //
+    // (*) Operator 3 is registered for another application but isn't registered
+    //   for the sanctioned application we use in test context.
```
pdyraga, line 212: 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏

nkuba: 🙇‍♂️

staker-rewards/rewards.js
nkuba: Ad 2.
I found it very hard to figure out a name for this property. I'm affraid any name we use will require someone seeing it in the table to be familiar with rewards rules, or at least what's behind this specific condition. Maybe it's not worth finding super descriptive and complex name but use a short one that would look good in the table, e.g. `fulfilledPoolRegistrationRequirement` or `fulfilledPoolRequirement`.

The other solution is to reverse this check and align it with the condition described in Coda doc:
>* if the operator has more ETH_unbonded than the minimum unbonded value and is not in the sortition pool at T_start, they are not getting any rewards for interval T.

So the property would be named: `brokePoolRegistrationRequirement`, `failedPoolRegistrationRequirement` and it would return value oposite to the one we return currently.

pdyraga: I'd add `*AtStart` no matter what because that's what we are checking. Especially that we have `poolDeauthorizedInInterval` in another column and that's the only way for the operator to leave the pool.

I'd go with one of those:
```
poolRequirementFulfilledAtStart
inPoolOrNotRequiredInPoolAtStart
```

nkuba: Oh yeah, I was so excited with the main part of the name that dropped the suffix. Of course we need `*AtStart`!

pdyraga (approved):

staker-rewards/lib/numbers.js
```diff
@@ -0,0 +1,7 @@
...
+  new BigNumber(value).dividedBy(new BigNumber(1e18)).toFixed(decimalPlaces)
```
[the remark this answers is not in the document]
nkuba: Covered in https://github.com/keep-network/keep-ecdsa/pull/640

PR #639, opened 2020-12-08
Staker Rewards: Check requirements while calculating rewards

lukasz-zimnoch:
Refs: #602
Depends on: #638

This pull request introduces requirements checks at the rewards calculation stage. The `totalRewards` parameter is set to `0` if at least one requirement violation occurred for the given operator. Possible violations are:
- `isFraudulent` is `true`
- `factoryAuthorizedAtStart` is `false`
- `poolAuthorizedAtStart` is `false`
- `poolDeauthorizedInInterval` is `true`
- `minimumStakeAtStart` is `false`
- `poolRequirementFulfilledAtStart` is `false`
- `keygenSLA` is less than `90` for `keygenCount >= 10` OR `keygenFailCount > 1` for `keygenCount < 10`
- `signatureSLA` is less than `95` for `signatureCount >= 20` OR `signatureFailCount > 1` for `signatureCount < 20`

### DIFF
--- a/staker-rewards/lib/rewards-calculator.js
+++ b/staker-rewards/lib/rewards-calculator.js
@@ -131,8 +131,8 @@ export default class RewardsCalculator {
       violations.push("minimumStakeAtStart")
     }
 
-    if (requirements.minimumUnbondedValueRegisteredAtStart === false) {
-      violations.push("minimumUnbondedValueRegisteredAtStart")
+    if (requirements.poolRequirementFulfilledAtStart === false) {
+      violations.push("poolRequirementFulfilledAtStart")
     }
 
     if (operatorSLA.keygenSLA !== "N/A" && operatorSLA.keygenSLA < 90) {
--- a/staker-rewards/lib/rewards-calculator.js
+++ b/staker-rewards/lib/rewards-calculator.js
@@ -203,5 +203,7 @@ function OperatorRewards(
     (this.boost = boost),
     (this.rewardWeight = rewardWeight),
     (this.totalRewards = totalRewards),
-    (this.requirementsViolations = requirementsViolations)
+    (this.requirementsViolations = requirementsViolations),
+    (this.SLAViolated =
+      requirementsViolations.filter((r) => r.includes("SLA")).length > 0)
 }
--- a/staker-rewards/lib/rewards-calculator.js
+++ b/staker-rewards/lib/rewards-calculator.js
@@ -26,6 +26,10 @@ export default class RewardsCalculator {
     for (const operatorParameters of operatorsParameters) {
       const { keepStaked, ethTotal } = operatorParameters.operatorAssets
 
+      const requirementsViolations = this.checkRequirementsViolations(
+        operatorParameters
+      )
+
       const ethScore = this.calculateETHScore(ethTotal)
       const boost = this.calculateBoost(keepStaked, ethTotal, minimumStake)
       const rewardWeight = ethScore.multipliedBy(boost)
@@ -35,6 +39,7 @@ export default class RewardsCalculator {
         ethScore,
         boost,
         rewardWeight,
+        requirementsViolations,
       })
     }
 
@@ -52,12 +57,15 @@ export default class RewardsCalculator {
     const operatorsRewards = []
 
     for (const operatorRewardsFactors of operatorsRewardsFactors) {
-      const { rewardWeight } = operatorRewardsFactors
+      const { rewardWeight, requirementsViolations } = operatorRewardsFactors
       const rewardRatio = rewardWeightSum.isGreaterThan(new BigNumber(0))
         ? rewardWeight.dividedBy(rewardWeightSum)
         : new BigNumber(0)
 
-      const totalRewards = this.interval.totalRewards.multipliedBy(rewardRatio)
+      const totalRewards =
+        requirementsViolations.length > 0
+          ? new BigNumber(0)
+          : this.interval.totalRewards.multipliedBy(rewardRatio)
 
       operatorsRewards.push(
         new OperatorRewards(
@@ -65,7 +73,8 @@ export default class RewardsCalculator {
           operatorRewardsFactors.ethScore,
           operatorRewardsFactors.boost,
           rewardWeight,
-          totalRewards
+          totalRewards,
+          requirementsViolations
         )
       )
     }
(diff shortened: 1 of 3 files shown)


staker-rewards/lib/rewards-calculator.js
```diff
@@ -98,6 +107,45 @@ export default class RewardsCalculator {
...
+      violations.push("minimumUnbondedValueRegisteredAtStart")
+    }
+
+    if (operatorSLA.keygenSLA !== "N/A" && operatorSLA.keygenSLA < 90) {
+      violations.push("keygenSLA")
+    }
+
+    if (operatorSLA.signatureSLA !== "N/A" && operatorSLA.signatureSLA < 95) {
+      violations.push("signatureSLA")
+    }
```
pdyraga, line 144: #638 introduced `poolRequirementFulfilledAtStart` requirement, let's make sure it's also covered.

staker-rewards/lib/rewards-calculator.js
```diff
@@ -57,20 +62,24 @@ export default class RewardsCalculator {
...
+      const { rewardWeight, requirementsViolations } = operatorRewardsFactors
       const rewardRatio = rewardWeightSum.isGreaterThan(new BigNumber(0))
         ? rewardWeight.dividedBy(rewardWeightSum)
         : new BigNumber(0)
 
-      const totalRewards = this.interval.totalRewards.multipliedBy(rewardRatio)
+      const totalRewards =
+        requirementsViolations.length > 0
+          ? new BigNumber(0)
+          : this.interval.totalRewards.multipliedBy(rewardRatio)
```
pdyraga, line 73: Just to make sure I understand the mechanism correctly: if one operator violated rules, what's happening with their share of rewards? Is it distributed among other operators?

lukasz-zimnoch: Nope, it remains in the pool.

staker-rewards/lib/rewards-calculator.js
```diff
@@ -98,6 +107,45 @@ export default class RewardsCalculator {
...
+      violations.push("poolRequirementFulfilledAtStart")
+    }
+
+    if (operatorSLA.keygenSLA !== "N/A" && operatorSLA.keygenSLA < 90) {
+      violations.push("keygenSLA")
+    }
+
+    if (operatorSLA.signatureSLA !== "N/A" && operatorSLA.signatureSLA < 95) {
+      violations.push("signatureSLA")
+    }
```
pdyraga, line 144: The rules we discussed are:

If the operator does not meet signature SLA between T_start and T_end , they are not getting any rewards for interval T . SLA is:
- For 20 or less closed/terminated keeps, no more than one with a signature timeout.
- For more than 20 closed/terminated keeps/ no more than 5% with a signature timeout.

Example:
For less than 20 redemptions, one of them can fail.
For 20-40 redemptions, only 1 can fail.  
For 50 redemptions, only 2 can fail.

(Signature SLA does NOT include not answered courtesy calls.)

If the operator has more than 10% of unsuccessful key generations between T_start and T_end , they are not getting any rewards for interval T. SLA is:
- For 2 or less key generations, none of them can fail.
- For 2-10 or less key generations, one of them can fail.
- For more than 10 key generations, no more than 10% can fail.

Example:
For 2 key generations, none of them can fail.
For 10 key generations, 1 of them can fail.
For 20 key generations, 2 of them can fail.
For 50 key generations, 5 of them can fail.

Let's also include a CAPITAL LETTER social caveat in the code that if we see DoS behaviors in the network, we can adjust SLAs.

staker-rewards/lib/rewards-calculator.js
pdyraga: How hard would be to change it so that it's distributed between other operators?

lukasz-zimnoch: Quite easy I think. We must set operator weight to `0` so it doesn't count in the rewards weight sum.

pdyraga: OK, let's leave it as-is and we'll discuss it separately.

pdyraga: (I also think it's a better approach than distributing this reward between operators from the current interval as this way we do not incentivize misbehavior DoSing other operators to increase one's reward.)

staker-rewards/lib/rewards-calculator.js
lukasz-zimnoch: Ok, done.

staker-rewards/lib/rewards-calculator.js
```diff
@@ -131,11 +195,13 @@ function OperatorRewards(
...
+  totalRewards,
+  requirementsViolations
 ) {
   ;(this.operator = operator),
     (this.ethScore = ethScore),
     (this.boost = boost),
     (this.rewardWeight = rewardWeight),
-    (this.totalRewards = totalRewards)
+    (this.totalRewards = totalRewards),
+    (this.requirementsViolations = requirementsViolations)
```
pdyraga, line 206: I am not sure if we need to list all violations in a separate column. It is pretty clear looking at `factoryAuthorizedAtStart`, `poolAuthorizedAtStart` and the rest what's wrong with the operator. I think the only problematic case is SLA which has more complex logic.

For that reason, what do you think about removing `requirementsViolations` column altogether and instead display `SLAViolated` column only for SLA? I am fine with having `requirementsViolations` field in a struct and using it in tests, I am just not sure if we need to print the value in the table given that this table already has enormous width.

staker-rewards/test/rewards-calculator.test.js
```diff
@@ -164,4 +180,332 @@ describe("rewards calculator", async () => {
...
+    "should set total rewards to zero if keygenCount >= 10 and " +
+      "failures percentage exceeds 10%",
+    async () => {
+      const mockContext = createMockContext()
+
+      setupContractsMock(mockContext)
+
+      const operatorParameters = createOperatorParameters(operator, 70000, 100)
+
+      operatorParameters.operatorSLA.keygenCount = 10
```
pdyraga, line 340: Can we have a separate test for `== 10` and `>= 10`?

The current format is a little confusing because test name says about `>=10` while we test only `== 10` boundary condition.

staker-rewards/test/rewards-calculator.test.js
```diff
@@ -164,4 +180,332 @@ describe("rewards calculator", async () => {
...
+    "should NOT set total rewards to zero if signatureCount >= 20 and " +
+      "failures percentage doesn't exceed 5%",
+    async () => {
+      const mockContext = createMockContext()
+
+      setupContractsMock(mockContext)
+
+      const operatorParameters = createOperatorParameters(operator, 70000, 100)
+
+      operatorParameters.operatorSLA.signatureCount = 20
```
pdyraga, line 418: Same question as for the key-gen case, can we have two separate tests: one for `== 20` and another for `>= 20`?

staker-rewards/test/rewards-calculator.test.js
lukasz-zimnoch: Isn't it redundant? If we test for `11` we can test for `12` as well.

pdyraga: Yeah, you are right.

pdyraga (approved):

PR #640, opened 2020-12-10
Staker Rewards: Made consistent numbers formatting

nkuba:
We extracted code to numbers formatting to a `toFormat` function and use it consistently. 

This PR is a followup from https://github.com/keep-network/keep-ecdsa/pull/638#discussion_r539948341.


pdyraga:
I'll leave a note for the future, not sure if it's solved in this PR or not. If I generate rewards for interval `1605225600 - 1605830400`, for operator `0x7a9654B888dEa4Fed7bf4c1cD104Bd45217427dA`, rewards value displayed in the table is `1294.31`. However, in text mode, this reward is `1294305182352503820898` and it matches the entry in merkle object: `462a17b9d3ac2d5262`. There is something off when it comes to rounding in the verification (`-v`) mode. We probably use the default rounding there instead of round down.

staker-rewards/lib/numbers.js
```diff
@@ -1,7 +1,18 @@
...
+}
 
 export const shorten18Decimals = (value) =>
-  new BigNumber(value).dividedBy(new BigNumber(1e18)).toFixed(decimalPlaces)
+  toFormat(new BigNumber(value).dividedBy(new BigNumber(1e18)))
+
+export const toFormat = (value, decimals = true) =>
+  new BigNumber(value).toFormat(
+    decimals ? decimalPlaces : noDecimalPlaces,
+    BigNumber.ROUND_DOWN,
```
lukasz-zimnoch, line 16: We should allow passing the rounding mode as a parameter. `BigNumber.ROUND_DOWN` mode is used only in few cases.

lukasz-zimnoch:
@nkuba the problem described by @pdyraga is probably caused by the fact this PR introduces `BigNumber.ROUND_DOWN` mode everywhere. Originally, this mode has been used only in few places while the rest used the default rounding mode.

nkuba:
@pdyraga @lukasz-zimnoch I updated rounding part of the PR.

staker-rewards/lib/rewards-calculator.js
```diff
@@ -51,12 +51,7 @@ export default class RewardsCalculator {
     )
 
     console.log(
-      clc.yellow(
-        `Rewards weight sum ${rewardWeightSum.toFixed(
-          noDecimalPlaces,
-          BigNumber.ROUND_DOWN
-        )}`
-      )
+      clc.yellow(`Rewards weight sum ${toFormat(rewardWeightSum, false)}`)
```
lukasz-zimnoch, line 54: `BigNumber.ROUND_DOWN` should be passed as the third parameter. Otherwise, the default `ROUND_HALF_UP` will be used.

staker-rewards/lib/rewards-calculator.js
```diff
@@ -87,7 +82,7 @@ export default class RewardsCalculator {
     const totalRewardsSum = operatorsRewards.reduce(
       (accumulator, rewards) =>
         accumulator.plus(
-          rewards.totalRewards.toFixed(noDecimalPlaces, BigNumber.ROUND_DOWN)
+          toFormat(rewards.totalRewards, false, BigNumber.ROUND_DOWN)
```
lukasz-zimnoch, line 85: I'd rather leave it as it is. We don't need to format here, just `toFixed` in order to perform addition.

lukasz-zimnoch (changes requested):

lukasz-zimnoch:
@nkuba two last comments and I'm ready to merge.

staker-rewards/lib/rewards-calculator.js
nkuba: Why are we doing `toFixed` here? Can't we just add two big ints?

lukasz-zimnoch: We need to crop the decimal part before adding the value to the `accumulator` to keep the results consistent. However seems like `rewards.totalRewards.integerValue(BigNumber.ROUND_DOWN)` fits better here than `          rewards.totalRewards.toFixed(noDecimalPlaces, BigNumber.ROUND_DOWN)` despite the eventual result will be the same.

nkuba: Cool, cool. Thanks for clarification. I'll update this line.

lukasz-zimnoch (approved):

PR #641, opened 2020-12-11
Transfer rewards to the beneficiary

nkuba:
Refs: #602

We expect the account passed to the rewards `claim` function to be an operator account.
In such case, rewards should be transferred to the operator's beneficiary address.

### DIFF
--- a/solidity/contracts/ECDSARewardsDistributor.sol
+++ b/solidity/contracts/ECDSARewardsDistributor.sol
@@ -15,6 +15,7 @@
 pragma solidity 0.5.17;
 
 import "@keep-network/keep-core/contracts/KeepToken.sol";
+import "@keep-network/keep-core/contracts/TokenStaking.sol";
 import "openzeppelin-solidity/contracts/ownership/Ownable.sol";
 import "openzeppelin-solidity/contracts/token/ERC20/SafeERC20.sol";
 import "openzeppelin-solidity/contracts/cryptography/MerkleProof.sol";
@@ -36,12 +37,14 @@ contract ECDSARewardsDistributor is Ownable {
     using SafeERC20 for KeepToken;
 
     KeepToken public token;
+    TokenStaking public tokenStaking;
 
     // This event is triggered whenever a call to #claim succeeds.
     event RewardsClaimed(
         bytes32 indexed merkleRoot,
         uint256 indexed index,
         address indexed account,
+        address beneficiary,
         uint256 amount
     );
     // This event is triggered whenever rewards are allocated.
@@ -54,8 +57,9 @@ contract ECDSARewardsDistributor is Ownable {
     // Bytes32 key is a merkle root and the value is a packed array of booleans.
     mapping(bytes32 => mapping(uint256 => uint256)) private claimedBitMap;
 
-    constructor(address _token) public {
+    constructor(address _token, address _tokenStaking) public {
         token = KeepToken(_token);
+        tokenStaking = TokenStaking(_tokenStaking);
     }
 
     /// Claim KEEP rewards for a given merkle root (interval) and the given address.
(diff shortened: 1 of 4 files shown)


solidity/contracts/ECDSARewardsDistributor.sol
```diff
@@ -86,9 +90,13 @@ contract ECDSARewardsDistributor is Ownable {
...
         _setClaimed(merkleRoot, index);
-        require(IERC20(token).transfer(account, amount), "Transfer failed");
 
-        emit RewardsClaimed(merkleRoot, index, account, amount);
+        address beneficiary = tokenStaking.beneficiaryOf(account);
+        require(beneficiary != address(0), "Beneficiary address not set");
+
+        require(IERC20(token).transfer(beneficiary, amount), "Transfer failed");
+
+        emit RewardsClaimed(merkleRoot, index, account, beneficiary, amount);
```
dimpar, line 99: Since we're adding 'beneficiary', maybe we should rename 'account' to 'operator' to make it clear what account we have in mind.

pdyraga: Yeah, I also vote for renaming `account` to `operator`, also as a parameter of `claim` function.

solidity/contracts/ECDSARewardsDistributor.sol
```diff
@@ -36,12 +37,14 @@ contract ECDSARewardsDistributor is Ownable {
     using SafeERC20 for KeepToken;
 
     KeepToken public token;
+    TokenStaking public tokenStaking;
 
     // This event is triggered whenever a call to #claim succeeds.
     event RewardsClaimed(
         bytes32 indexed merkleRoot,
         uint256 indexed index,
         address indexed account,
```
pdyraga, line 46: `s/account/operator`

solidity/test/staker-rewards/rewardsData.js
```diff
@@ -85,6 +90,7 @@ const testValues = {
           "0x20779de3e5a9a44f012cedf6f56e78af493fcb441d19cd884471e28457bbf697",
           "0x172864ef2714d9feaf400fe1342b595b61fc740234de3a7cd4e3d4eb8be3fc36",
         ],
+        beneficiary: "0xfc8437956EeaCCE97996Ad01af59b828F3F5A808",
```
pdyraga, line 93: Can we rename `account` to `operator` in this file as well?

solidity/test/staker-rewards/MerkleDistributorTest.js
```diff
@@ -88,6 +102,7 @@ describe("MerkleDistributor", () => {
       const account = merkle0.claims[0].account
```
pdyraga, line 102: `s/account/operator` please, same in other cases in this file

pdyraga (approved):
Looks good to me, I'll leave final approval and merge honors to @dimpar.

solidity/contracts/ECDSARewardsDistributor.sol
```diff
@@ -78,17 +83,21 @@ contract ECDSARewardsDistributor is Ownable {
...
             "Invalid proof"
         );
 
         // Mark it claimed and send the token.
         _setClaimed(merkleRoot, index);
-        require(IERC20(token).transfer(account, amount), "Transfer failed");
 
-        emit RewardsClaimed(merkleRoot, index, account, amount);
+        address beneficiary = tokenStaking.beneficiaryOf(operator);
+        require(beneficiary != address(0), "Beneficiary address not set");
```
dimpar, line 96: I'm not sure if we actually need this check:
- is it possible that we have an operator without beneficiary?
- I think `transfer` already [checks](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/ERC20.sol#L210) for `address(0)`

pdyraga: I initially had the same thoughts. On the other hand, it makes sense to treat this contract independently and this contract does not have to be aware of the rules of `TokenStaking` has. 

For the second point, that's true but I think that printing a more targeted revert makes sense.

nkuba: I think that a check for zero address should be there. As this is already covered by a library we use underneth it's redundant to have a separate require. The message returned by the OZ's library is very fine: `ERC20: transfer to the zero address`. I'll remove the require statement and update tests.

dimpar: I'm okay with both. I know it wouldn't hurt us to check and write a targeted message and I take that the gas usage should be marginal.

pdyraga: `ERC20: transfer to the zero address` message is detailed enough. I haven't checked what it is and assumed it's more generic.

nkuba: fea7a4800752ddbda01ddf7423aa73431e5e502d

dimpar (approved):

PR #645, opened 2020-12-14
Simple liquidity rewards

mhluongo:
Parameterized Synthetix's Unipool implementation for more flexible reward delivery.

@dimpar I'll let you take the PR from here — I haven't tried to build or thought about deploys, but the change to `Unipool` is simple. I'm also confident in the original source that I pulled in from mainnet.

Initially we need `LPRewards` deployed for these pairs

`TBTC/ETH` - https://info.uniswap.org/pair/0x854056fd40c1b52037166285b2e54fee774d33f6
`KEEP/ETH` - https://info.uniswap.org/pair/0xe6f19dab7d43317344282f803f8e8d240708174a
`KEEP/TBTC` - https://info.uniswap.org/pair/0x38c8ffee49f286f25d25bad919ff7552e5daf081


solidity/contracts/LPRewards.sol
```diff
@@ -0,0 +1,211 @@
...
+}
+
+contract LPTokenWrapper {
+    using SafeMath for uint256;
+    using SafeERC20 for IERC20;
+
+    uint256 private _totalSupply;
+    mapping(address => uint256) private _balances;
+
+    IERC20 public wrapped; // LPToken pair KEEP/ETH, TBTC/ETH, KEEP/TBTC
```
nkuba, line 71:
```suggestion
    IERC20 public wrappedToken; // LPToken pair KEEP/ETH, TBTC/ETH, KEEP/TBTC
```

solidity/test/LPRewardsTest.js
```diff
@@ -0,0 +1,119 @@
...
+const chai = require("chai")
+chai.use(require("bn-chai")(BN))
+const expect = chai.expect
+
+describe("LPRewards", () => {
+  const tokenDecimalMultiplier = web3.utils.toBN(10).pow(web3.utils.toBN(18))
+
+  let keepToken
+  let lpRewards
+  let lpToken
```
nkuba, line 19:
```suggestion
  let lpWrappedToken
```

solidity/test/LPRewardsTest.js
```diff
@@ -0,0 +1,119 @@
+const {contract, web3} = require("@openzeppelin/test-environment")
+const {createSnapshot, restoreSnapshot} = require("./helpers/snapshot")
+const {accounts} = require("@openzeppelin/test-environment")
+
+const LPRewards = contract.fromArtifact("LPRewards")
+const KeepToken = contract.fromArtifact("KeepToken")
+const LPToken = contract.fromArtifact("TestToken")
```
nkuba, line 7:
```suggestion
const LPWrappedToken = contract.fromArtifact("TestToken")
```

solidity/contracts/LPRewards.sol
nkuba: After looking at tests I have second thoughts... Maybe this field should be named `lpToken`. I'm not sure. Anyways we need to align naming in tests with naming in the contract.

solidity/test/LPRewardsTest.js
nkuba: Yeah, second thoughts I mentioned in the other comment. Let's align the name with the one we decide to use in the contract.

nkuba (commented):
Leaving a first part of comments. I hope we will add more tests to cover all the functions in the contracts.

dimpar:
> Leaving a first part of comments. I hope we will add more tests to cover all the functions in the contracts.

Yes, I was planning on adding some more tests for our purposes to mainly show the different flows, but please note that [Unipool.sol](https://github.com/Synthetixio/Unipool/blob/master/contracts/Unipool.sol) was already tested [here](https://github.com/Synthetixio/Unipool/blob/master/test/Unipool.js). I take that this is a [trusted source](https://www.flowdock.com/app/cardforcoin/tech/threads/9iONO4-aowD16gzMOCU4bpz-cMd). Our contracts are heavily based on Unipool.sol with some minor changes.

mhluongo:
Indeed, this has been audited and in highly-visible production for the past 8 months. We just need confidence on the diff

solidity/contracts/LPRewards.sol
mhluongo: Note the contract works for any wrapped ERC-20, so the wrapped token probably shouldn't have "LP" in the name. I thought it made more sense for the contract but technically neither are restricted

solidity/contracts/LPRewards.sol
```diff
@@ -0,0 +1,201 @@
...
+        _totalSupply = _totalSupply.sub(amount);
+        _balances[msg.sender] = _balances[msg.sender].sub(amount);
+        wrapped.safeTransfer(msg.sender, amount);
+    }
+}
+
+// Or LPTokenRewards?
+contract LPRewards is LPTokenWrapper, IRewardDistributionRecipient {
+    IERC20 public rewardToken; // KEEP in our case. |||| snx in Unipool case
+    uint256 public constant DURATION = 7 days;
```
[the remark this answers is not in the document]
mhluongo: Can we keep the code as similar as possible to the source? `Unipool` is basically a standard

dimpar:
I compared the imported libs to make sure we're not breaking anything under the hood by using [OpenZeppelin v2.3.0](https://github.com/keep-network/keep-ecdsa/blob/master/solidity/package.json#L40) instead of Unipool's [OpenZeppeling v2.4.0](https://github.com/Synthetixio/Unipool/blob/master/package.json#L11). Here is their [Changelog for v2.4.0](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/CHANGELOG.md#240-2019-10-29). I don't see anything in "Breaking changes" that can affect us.

I compared the initial commit df98c9b with OZ v2.3.0 and here's my findings:
- `Math.sol` - same
- `SafeMath.sol` - updated to throw custom errors on subtraction overflows.
- `Context.sol` - instead of using `_msgSender()` which returns `msg.sender` now we just return `msg.sender`
- `Ownable.sol` - same, it used '_msgSender()' from `Context.sol` which just returned `msg.sender`
- `IERC20.sol` - same, changes were only in the comments
- `Address.sol` - is used by `SafeERC20`. We are interested in one function `Address.isContract()`. This function changed in v2.4.0 to lower the gas consumption, HOWEVER later they [figured out](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/1802#issuecomment-657668876) that it is not cheaper and switched the code back. Master has exactly the same code as v2.3.0 (which what we currently use). I think we are good on this front.
- `SafeERC20.sol` - minor change in error message when subtracting.

solidity/contracts/LPRewards.sol
```diff
@@ -0,0 +1,211 @@
...
+
+    function withdraw(uint256 amount) public {
+        _totalSupply = _totalSupply.sub(amount);
+        _balances[msg.sender] = _balances[msg.sender].sub(amount);
+        wrapped.safeTransfer(msg.sender, amount);
+    }
+}
+
+contract LPRewards is LPTokenWrapper, IRewardDistributionRecipient {
+    IERC20 public rewardToken; // KEEP
```
pdyraga, line 99: If it's always KEEP - even the comment says that - why don't we hardcode it?

EDIT: OK, I see - tests.

solidity/contracts/LPRewards.sol
```diff
@@ -0,0 +1,211 @@
...
+*
+* THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+* IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+* FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+* AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+* LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+* OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+*/
+
+pragma solidity 0.5.17;
```
pdyraga, line 36: Can we add a comment with a reference to mainnet contract address from where we borrowed the code?

solidity/test/LPRewardsTest.js
```diff
@@ -0,0 +1,119 @@
...
+const LPRewards = contract.fromArtifact("LPRewards")
+const KeepToken = contract.fromArtifact("KeepToken")
+const WrappedToken = contract.fromArtifact("TestToken")
+
+const BN = web3.utils.BN
+const chai = require("chai")
+chai.use(require("bn-chai")(BN))
+const expect = chai.expect
+
+describe("LPRewards", () => {
```
pdyraga, line 14: What flows do we want to cover with tests?

solidity/test/LPRewardsTest.js
```diff
@@ -0,0 +1,220 @@
...
+
+  let keepToken
+  let lpRewards
+  let wrappedToken
+  let owner
+  let wallet1
+  let wallet2
+
+  before(async () => {
+    owner = accounts[0]
```
nkuba, line 25: Thoughts on naming it `rewardDistribution` as it is in the contract.

nkuba, line 25: Let's use different accounts than default `0`.

solidity/test/LPRewardsTest.js
```diff
@@ -0,0 +1,220 @@
...
+    await restoreSnapshot()
+  })
+
+  describe("tokens allocation", () => {
+    it("should successfully allocate KEEP tokens", async () => {
+      let keepBalance = await keepToken.balanceOf(lpRewards.address)
+      expect(keepBalance).to.eq.BN(0)
+
+      const rewards = web3.utils.toBN(1000042).mul(tokenDecimalMultiplier)
+      await fundKEEPReward(lpRewards.address, rewards)
```
nkuba, line 54: Thoughts on removing `fundKEEPReward` function and using ` await keepToken.approveAndCall(address, amount, "0x0")` instead? This way we will see how the transfer is made and that we're testing `approveAndCall` path.

nkuba, line 54: Let's add a line break to separate parameter configuiration from the actual test function call.
```suggestion
      const rewards = web3.utils.toBN(1000042).mul(tokenDecimalMultiplier)
      
      await fundKEEPReward(lpRewards.address, rewards)
```

solidity/test/LPRewardsTest.js
```diff
@@ -0,0 +1,220 @@
...
+
+      keepBalance = await keepToken.balanceOf(lpRewards.address)
+      expect(keepBalance).to.eq.BN(rewards)
```
nkuba, line 57: We could drop the first `expect` and use the `initialBalance` in the final check.
```js
let initialBalance = await keepToken.balanceOf(lpRewards.address)

// (...)

const finalBalance = await keepToken.balanceOf(lpRewards.address)
expect(finalBalance).to.eq.BN(initialBalance.add(rewards))
```

solidity/test/LPRewardsTest.js
```diff
@@ -0,0 +1,220 @@
...
+  beforeEach(async () => {
+    await createSnapshot()
+  })
+
+  afterEach(async () => {
+    await restoreSnapshot()
+  })
+
+  describe("tokens allocation", () => {
+    it("should successfully allocate KEEP tokens", async () => {
```
nkuba, line 49:
```suggestion
    it("should successfully allocate KEEP tokens via receiveApproval function", async () => {
```

solidity/test/LPRewardsTest.js
```diff
@@ -0,0 +1,220 @@
...
+    keepToken = await KeepToken.new()
+    // This is a "Pair" Uniswap Token which is created here:
+    // https://github.com/Uniswap/uniswap-v2-core/blob/master/contracts/UniswapV2Factory.sol#L23
+    //
+    // There are 3 addresses for the following pairs:
+    // - KEEP/ETH (https://info.uniswap.org/pair/0xe6f19dab7d43317344282f803f8e8d240708174a)
+    // - TBTC/ETH (https://info.uniswap.org/pair/0x854056fd40c1b52037166285b2e54fee774d33f6)
+    // - KEEP/TBTC (https://info.uniswap.org/pair/0x38c8ffee49f286f25d25bad919ff7552e5daf081)
+    wrappedToken = await WrappedToken.new()
+    lpRewards = await LPRewards.new(keepToken.address, wrappedToken.address)
```
nkuba, line 37: Can we add new variables `keepTokenContractOwner`, `wrappedTokenContractOwner` and `lpRewardsContractOwner` that would be different account than default `0` and deploy contracts from these accounts?

solidity/contracts/LPRewards.sol
nkuba: True, true! I vote for `wrappedToken` then.

solidity/contracts/LPRewards.sol
nkuba: So maybe we could name the field `keepToken`?

solidity/contracts/LPRewards.sol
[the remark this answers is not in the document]
nkuba: Yes, that's a good point. I'm resolving most of the naming comments.

solidity/contracts/LPRewards.sol
```diff
@@ -0,0 +1,211 @@
...
+
+    function receiveApproval(
+        address from,
+        uint256 value,
+        address token,
+        bytes memory
+    ) public {
+        require(IERC20(token) == rewardToken, "Unsupported token");
+
+        rewardToken.safeTransferFrom(from, address(this), value);
```
nkuba, line 187: Shouldn't we call `stake(value)` function here?

nkuba: Oh, we souldn't, `stake` is for wrapped token. Here we add rewards token.

solidity/test/LPRewardsTest.js
```diff
@@ -0,0 +1,220 @@
...
+  let lpRewards
+  let wrappedToken
+  let owner
+  let wallet1
+  let wallet2
+
+  before(async () => {
+    owner = accounts[0]
+    wallet1 = accounts[1]
+    wallet2 = accounts[2]
```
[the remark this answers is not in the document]
nkuba: Or maybe `staker*` will be better. As they put wrapped tokens with `stake` function.

solidity/contracts/LPRewards.sol
dimpar: 70e7328

@mhluongo I'm not 100% sure if the source is correct. Can you please confirm?

solidity/test/LPRewardsTest.js
dimpar: What I wanted to add are some happy paths that can help us better understand the flows in these contracts:
- KEEP rewards allocations
- Wrapped Token (lp tokens) allocations
- User's KEEP rewards earnings after staking
- User's KEEP rewards and Wrapped tokens withdrawals

solidity/contracts/LPRewards.sol
dimpar: Correct, this guy is for KEEP tokens (rewards)

solidity/contracts/LPRewards.sol
dimpar: done in 85819dc

solidity/test/LPRewardsTest.js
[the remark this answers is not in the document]
dimpar: Done in f49f7c8

solidity/test/LPRewardsTest.js
```diff
@@ -0,0 +1,220 @@
...
+      )
+    })
+  })
+
+  describe("tokens distribution", () => {
+    const precision = 1
+
+    it("should be possible to check earned rewards", async () => {
+      const amount = new BN(10000)
+      const keepRewards = amount.mul(tokenDecimalMultiplier)
```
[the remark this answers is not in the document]
dimpar: 1d96dfc

solidity/contracts/LPRewards.sol
```diff
@@ -0,0 +1,216 @@
...
+* THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+* IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+* FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+* AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+* LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+* OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+*/
+
+/// @notice These contracts reward users for adding liquidity to Uniswap https://uniswap.org/
+/// @dev These contracts were obtained from Synthetix and added some minor changes.
```
pdyraga, line 37: Should we use `@notice` and `@dev` here? This comment does not look like something that goes to generated contract documentation.

solidity/test/LPRewardsTest.js
dimpar: 044bcab

dimpar: 044bcab

solidity/contracts/LPRewards.sol
```diff
@@ -0,0 +1,201 @@
...
+                    .mul(1e18)
+                    .div(totalSupply())
+            );
+    }
+
+    function earned(address account) public view returns (uint256) {
+        return
+            balanceOf(account)
+                .mul(rewardPerToken().sub(userRewardPerTokenPaid[account]))
+                .div(1e18)
```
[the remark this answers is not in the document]
pdyraga: It is better to keep it consistent with the audited contract from which we borrowed the code although I do agree, it is not consistent with ECDSA rewards distributor in this repo. Life. 🤷

solidity/contracts/LPRewards.sol
pdyraga: Source code comparison says it's a match.

solidity/contracts/LPRewards.sol
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    function withdraw(uint256 amount) public {
+        _totalSupply = _totalSupply.sub(amount);
+        _balances[msg.sender] = _balances[msg.sender].sub(amount);
+        wrappedToken.safeTransfer(msg.sender, amount);
+    }
+}
+
+contract LPRewards is LPTokenWrapper, IRewardDistributionRecipient {
```
nkuba, line 103: We need to define a dedicated contract  for each of the three wrapped tokens. This will simplify usage in the dashboard.

solidity/test/LPRewardsTest.js
pdyraga: I am not sure if it's an allocation we are testing here. `receiveApproval` is just transferring tokens. The real allocation is `receiveApproval` + `notifyRewardsAmount`. Thoughts on renaming test scenario?

solidity/test/LPRewardsTest.js
pdyraga: Yeah, I'd also use `approveAndCall` directly, we are having just one line there, no rocket science inside.

solidity/test/LPRewardsTest.js
pdyraga: Or, maybe even better, we can add `notifyRewardAmount` call to this scenario. I do not think there is a scenario where we call `receiveApproval` without calling `notifyRewardsAmount`.

solidity/test/LPRewardsTest.js
```diff
@@ -0,0 +1,219 @@
...
+
+      const wrappedTokensToStake = web3.utils
+        .toBN(2000)
+        .mul(tokenDecimalMultiplier)
+      await lpRewards.stake(wrappedTokensToStake, {from: staker1})
+
+      const future = (await time.latest()).add(time.duration.days(7))
+      await timeIncreaseTo(future)
+
+      const actualEarnings = (await lpRewards.earned(staker1)).div(
```
nkuba, line 140: I'm not a fan of doing `.div(tokenDecimalMultiplier)` and `.mul(tokenDecimalMultiplier)` in so many places. Could we define values at the begining on the test including `tokenDecimalMultiplier` and don't use it aferwards?

solidity/contracts/LPRewards.sol
```diff
@@ -0,0 +1,216 @@
...
+
+    function receiveApproval(
+        address from,
+        uint256 value,
+        address token,
+        bytes memory
+    ) public {
+        require(IERC20(token) == keepToken, "Unsupported token");
+
+        keepToken.safeTransferFrom(from, address(this), value);
```
pdyraga, line 192: We should call `notifyRewardAmount` from here. The reward distribution contract will be a beneficiary of `PhasedEscrow`.

pdyraga (changes requested):

solidity/test/LPRewardsTest.js
pdyraga: Nope, heh. `receiveApproval` needs to call `notifyRewardsAmount` because `PhasedEscrow`'s beneficiary will be the reward allocation contract.

solidity/test/LPRewardsTest.js
```diff
@@ -0,0 +1,219 @@
...
+        lpRewards.address,
+        staker1,
+        initialWrappedTokenStakerBallance1
+      )
+      await mintAndApproveWrappedTokens(
+        wrappedToken,
+        lpRewards.address,
+        staker2,
+        initialWrappedTokenStakerBallance2
+      )
```
pdyraga, line 75: Does it make sense to have this functionality extracted to a separate function? We now have 12 lines of code instead of 4.

solidity/test/LPRewardsTest.js
```diff
@@ -0,0 +1,219 @@
...
+
+      let wrappedTokenBalance = await wrappedToken.balanceOf(lpRewards.address)
+      expect(wrappedTokenBalance).to.eq.BN(0)
+
+      let wrappedTokenStakerBalance1 = await wrappedToken.balanceOf(staker1)
+      expect(wrappedTokenStakerBalance1).to.eq.BN(initialWrappedTokenStakerBallance1)
+      let wrappedTokenStakerBalance2 = await wrappedToken.balanceOf(staker2)
+      expect(wrappedTokenStakerBalance2).to.eq.BN(initialWrappedTokenStakerBallance2)
```
pdyraga, line 83: Do we need those assertions? Are we not testing `token.mint` this way?

solidity/test/LPRewardsTest.js
```diff
@@ -0,0 +1,220 @@
...
+      expect(keepBalance).to.eq.BN(0)
+
+      const rewards = web3.utils.toBN(1000042).mul(tokenDecimalMultiplier)
+      await fundKEEPReward(lpRewards.address, rewards)
+
+      keepBalance = await keepToken.balanceOf(lpRewards.address)
+      expect(keepBalance).to.eq.BN(rewards)
+    })
+
+    it("should successfully allocate wrapped tokens", async () => {
```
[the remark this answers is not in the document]
pdyraga: Shouldn't this test be named `... stake wrapped tokens`? We are testing `stake` function here.

solidity/contracts/LPRewards.sol
```diff
@@ -0,0 +1,201 @@
...
+}
+
+contract LPTokenWrapper {
+    using SafeMath for uint256;
+    using SafeERC20 for IERC20;
+
+    uint256 private _totalSupply;
+    mapping(address => uint256) private _balances;
+
+    IERC20 public wrapped; // Uni pair KEEP/ETH, TBTC/ETH, KEEP/TBTC
```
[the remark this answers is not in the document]
pdyraga: It's named `wrappedToken` now. Resolving.

solidity/contracts/LPRewards.sol
pdyraga: It's named `wrappedToken` now, resolving.

solidity/test/LPRewardsTest.js
```diff
@@ -0,0 +1,219 @@
...
+  let lpRewards
+  let wrappedToken
+  let rewardDistribution
+  let staker1
+  let staker2
+
+  before(async () => {
+    rewardDistribution = accounts[5]
+    staker1 = accounts[1]
+    staker2 = accounts[2]
```
pdyraga, line 27: We can declare them as constants earlier - that's how we do it in other tests so it's worth keeping it consistent.

solidity/test/LPRewardsTest.js
```diff
@@ -0,0 +1,219 @@
...
+      const actualRewardPerToken = (await lpRewards.rewardPerToken()).div(
+        tokenDecimalMultiplier
+      )
+      const expectedRewardPerWrappedToken = new BN(5)
+      expect(actualRewardPerToken).to.gte.BN(
+        expectedRewardPerWrappedToken.subn(precision)
+      )
+      expect(actualRewardPerToken).to.lte.BN(
+        expectedRewardPerWrappedToken.addn(precision)
+      )
```
nkuba, line 160: Woud could try to calculate the exact amount like in the example below. But also we can hardcode the expected value.

```js
      const duration = await lpRewards.DURATION.call()
      const rewardRate = keepRewards.div(duration)

      const expectedRewardPerWrappedToken = rewardRate
        .mul(duration)
        .mul(tokenDecimalMultiplier)
        .div(stakedAmount)
```

nkuba: Hardcoding can be the cleaniest solution. Of course we need to add a comment next to the value definition with steps how it was calculated.

solidity/test/LPRewardsTest.js
```diff
@@ -0,0 +1,219 @@
...
+
+      // Withdraw wrapped tokens and KEEP rewards
+      await lpRewards.exit({from: staker1})
+
+      // Earned KEEP rewards for adding liquidity
+      const keepEarnedRewards = (await keepToken.balanceOf(staker1)).div(
+        tokenDecimalMultiplier
+      )
+      expect(keepEarnedRewards).to.gte.BN(rewardsAmount.subn(precision))
+      expect(keepEarnedRewards).to.lte.BN(rewardsAmount.addn(precision))
```
nkuba, line 197: I think we can hardcode `expectedEarnedRewards` so we can have just one exact equality check.

pdyraga: How would that help? It's a time precision issue to my understanding.

solidity/test/LPRewardsTest.js
dimpar: 0e2e07c

solidity/test/LPRewardsTest.js
dimpar: fc61d96

solidity/contracts/LPRewards.sol
dimpar: In b148edc we removed `approveAndCall` and instead transferring funds from `notifyRewardAmount()`

solidity/contracts/LPRewards.sol
dimpar: Removed

solidity/test/LPRewardsTest.js
dimpar: b148edc Now we transfer funds in `notifyRewardsAmount()` and `receiveApproval` is gone

solidity/test/LPRewardsTest.js
dimpar: done

solidity/test/LPRewardsTest.js
dimpar: Done, aligned with the one in contract

solidity/test/LPRewardsTest.js
dimpar: Approach of transferring funds was changed, `approveAndCall` is not relevant anymore.

dimpar: done

solidity/test/LPRewardsTest.js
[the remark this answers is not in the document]
dimpar: dc0f47c

solidity/contracts/LPRewards.sol
dimpar: https://github.com/keep-network/keep-ecdsa/pull/650

solidity/test/LPRewardsTest.js
dimpar: 09d89f3 yes, I think we can remove it too

solidity/test/LPRewardsTest.js
dimpar: done

solidity/test/LPRewardsTest.js
```diff
@@ -0,0 +1,219 @@
...
+      expect(wrappedTokenStakerBalance2).to.eq.BN(initialWrappedTokenStakerBallance2)
+
+      await lpRewards.stake(web3.utils.toBN(4000).mul(tokenDecimalMultiplier), {
+        from: staker1,
+      })
+      await lpRewards.stake(web3.utils.toBN(5000).mul(tokenDecimalMultiplier), {
+        from: staker2,
+      })
+
+      wrappedTokenBalance = await wrappedToken.balanceOf(lpRewards.address)
```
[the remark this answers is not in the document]
dimpar: 7beae72

solidity/test/LPRewardsTest.js
```diff
@@ -0,0 +1,219 @@
...
+      await lpRewards.setRewardDistribution(rewardDistribution)
+      await lpRewards.notifyRewardAmount(keepRewards, {from: rewardDistribution})
+
+      const wrappedTokensToStake = web3.utils
+        .toBN(2000)
+        .mul(tokenDecimalMultiplier)
+      await lpRewards.stake(wrappedTokensToStake, {from: staker1})
+
+      const future = (await time.latest()).add(time.duration.days(7))
+      await timeIncreaseTo(future)
```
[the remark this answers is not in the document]
dimpar: Not sure about that. Having a separate duration of 7 days in tests also tests duration in the contract.

solidity/test/LPRewardsTest.js
dimpar: I know it doesn't look pretty, but I think there is a value when testing with `tokenDecimalMultiplier` which reflects the scenario when working with real tokens. When transferring tokens to a contract and then retrieving them, we need to divide by `10**18` which will also show a precision error that we're dealing with. This is a similar approach in other tests for rewards, ex. `TestECDSARewards.js`.
How about instead of multiplication we would use `web3.utils.toWei()` function 4e8d014?

solidity/test/LPRewardsTest.js
dimpar: Refactored d00548e

pdyraga:
@dimpar Can you update this branch with `master` please?

solidity/test/LPRewardsTest.js
```diff
@@ -0,0 +1,211 @@
...
+const WrappedToken = contract.fromArtifact("TestToken")
+
+const BN = web3.utils.BN
+const chai = require("chai")
+chai.use(require("bn-chai")(BN))
+const expect = chai.expect
+
+describe("LPRewards", () => {
+  const tokenDecimalMultiplier = web3.utils.toBN(10).pow(web3.utils.toBN(18))
+  const allocationForDistribution = web3.utils.toWei("10000000")
```
pdyraga, line 16: Nitpick: I'd rather express it as `web3.utils.toBN("10000000").mul(tokenDecimalMultiplier)` given that we are not transforming ether to wei but KEEP into 1e18 format. Same in all other places in this file where we use `toWei` but we are dealing with KEEP.

solidity/test/LPRewardsTest.js
```diff
@@ -0,0 +1,211 @@
...
+    // - TBTC/ETH (https://info.uniswap.org/pair/0x854056fd40c1b52037166285b2e54fee774d33f6)
+    // - KEEP/TBTC (https://info.uniswap.org/pair/0x38c8ffee49f286f25d25bad919ff7552e5daf081)
+    wrappedToken = await WrappedToken.new({from: wrappedTokenOwner})
+    lpRewards = await LPRewards.new(keepToken.address, wrappedToken.address, {
+      from: lpRewardsOwner,
+    })
+
+    await keepToken.approve(rewardDistribution, allocationForDistribution, {
+      from: keepTokenOwner,
+    })
```
pdyraga, line 45: Is this call needed?

solidity/test/LPRewardsTest.js
```diff
@@ -0,0 +1,211 @@
...
+    await createSnapshot()
+  })
+
+  afterEach(async () => {
+    await restoreSnapshot()
+  })
+
+  describe("tokens allocation", () => {
+    it("should successfully transfer KEEP and notify LPRewards contract on rewards distribution", async () => {
+      const initialBalance = await keepToken.balanceOf(lpRewards.address)
```
pdyraga, line 61: I'd rather assert it is zero, just like in the next test, or do not care about initial balance at all assuming it's zero.

solidity/test/LPRewardsTest.js
```diff
@@ -0,0 +1,217 @@
...
+  }
+
+  async function mintAndApproveWrappedTokens(staker, amount) {
+    await wrappedToken.mint(staker, amount)
+    await wrappedToken.approve(lpRewards.address, amount, {from: staker})
+  }
+
+  async function timeIncreaseTo(seconds) {
+    const delay = 10 - new Date().getMilliseconds()
+    await new Promise((resolve) => setTimeout(resolve, delay))
```
pdyraga, line 214: Why do we need it?

solidity/test/LPRewardsTest.js
```diff
@@ -0,0 +1,217 @@
...
+      )
+      expect(actualRewardPerToken).to.lte.BN(
+        expectedRewardPerWrappedToken.addn(precision)
+      )
+    })
+
+    it("should be possible to withdraw rewards after staking wrapped tokens", async () => {
+      const rewardsAmount = new BN(5000)
+      const keepAllocated = rewardsAmount.mul(tokenDecimalMultiplier)
+      const wrappedTokenStakerBallance = web3.utils.toWei("10000")
```
pdyraga, line 169: Typo: ballance->balance.

solidity/test/LPRewardsTest.js
dimpar: I've added a comment showing the math behind a reward per token. b0165bd

I think checking math on contract with a simple expected value in test should work, especially this is considered to be a trusted source.

solidity/test/LPRewardsTest.js
dimpar: Precision error is very small, we had the same issue with `TestECDSARewards.js` and handled it the same way there. Besides, looks like `Unipool.js` also tests values from Solidy in a similar fashion by using `almost equal` approach.

pdyraga (commented):
Five comments to tests, I consider neither of them blocking. Need some time to go through the contract code one more time. @dimpar if you can address them in the meantime - perfect, if not, I do not want to block here on them.

solidity/test/LPRewardsTest.js
```diff
@@ -0,0 +1,220 @@
...
+        tokenDecimalMultiplier
+      )
+      expect(keepEarnedRewards).to.gte.BN(rewardsAmount.subn(precision))
+      expect(keepEarnedRewards).to.lte.BN(rewardsAmount.addn(precision))
+
+      // Check that all wrapped tokens were transferred back to the wallet
+      const wrappedTokenWalletBalance = await wrappedToken.balanceOf(wallet1)
+      expect(wrappedTokenWalletBalance).to.eq.BN(wrappedTokenWalletBallance)
+    })
+  })
```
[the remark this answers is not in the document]
dimpar: `exit()` function above in this test calls `getReward()` and `withdraw()`. Then we check if a user actually received the rewards by calling `getReward()`

solidity/test/LPRewardsTest.js
dimpar: A leftover :( we don't need it.

solidity/test/LPRewardsTest.js
dimpar: Fixed

solidity/test/LPRewardsTest.js
dimpar: Removed

solidity/test/LPRewardsTest.js
dimpar: Yes, it won't work otherwise

solidity/test/LPRewardsTest.js
dimpar: okay, reverted to the original version

solidity/test/LPRewardsTest.js
pdyraga: We are still having the typo in other places, probably as a result of revert, e.g.:
```
      const wrappedTokenStakerBallance1 = web3.utils
        .toBN(10000)
        .mul(tokenDecimalMultiplier)
      const wrappedTokenStakerBallance2 = web3.utils
        .toBN(20000)
        .mul(tokenDecimalMultiplier)

      await mintAndApproveWrappedTokens(staker1, wrappedTokenStakerBallance1)
      await mintAndApproveWrappedTokens(staker2, wrappedTokenStakerBallance2)
```

dimpar: yes, sorry, done

solidity/test/LPRewardsTest.js
pdyraga: Hmm... I am pretty sure it does - just tried removing it and tests are passing. We do not need to `approve` to `transfer`. If that `approve` was added to execute the very next line. If that `approve` was added for something else, I think that *else* is no longer there.

dimpar: Darn, was looking at the wrong `approve` call. Yes, we don't need this one. 91b3bf3

pdyraga (approved):

PR #650, opened 2020-12-18
Dedicated LPRewards contracts for three pairs

nkuba:
We created dedicated LPRewards implementations for TBTC-ETH, KEEP-ETH
and KEEP-TBTC pairs. Those were mostly created to simplify usage in the
dashboard, as we will produce separate artifact for each pair.

Depends on: #645 
Depends on: https://github.com/keep-network/keep-core/pull/2225


solidity/contracts/LPRewards.sol
```diff
@@ -214,3 +221,27 @@ contract LPRewards is LPTokenWrapper, IRewardDistributionRecipient {
         }
     }
 }
+
+/// @title KEEP rewards for TBTC-ETH liquidity pool.
+contract LPRewardsTBTCETH is LPRewards {
+    constructor(KeepToken keepToken, IERC20 tbtcEthUniswapPair)
```
nkuba, line 227: The first parameter could be also `IERC20`. I'm not sure what's better, use specific `KeepToken` type or more general `IERC20`?

```suggestion
    constructor(IERC20 keepToken, IERC20 tbtcEthUniswapPair)
```

pdyraga: Given that those contracts are for specific pairs, I think it makes more sense to keep the code as-is.

solidity/contracts/LPRewards.sol
```diff
@@ -100,7 +103,11 @@ contract LPTokenWrapper {
     }
 }
 
-contract LPRewards is LPTokenWrapper, IRewardDistributionRecipient {
+contract LPRewards is
+    LPTokenWrapper,
+    IRewardDistributionRecipient,
+    IStakerRewards
```
pdyraga, line 109: It is no longer needed, `LPRewards` will receive tokens via `notifyRewardAmount` called by `StakingPoolRewardsEscrowBeneficiary` (see https://github.com/keep-network/keep-core/pull/2230) - it is the same mechanism as for `CurveRewards`.

solidity/contracts/LPRewardsEscrowBeneficiary.sol
```diff
@@ -0,0 +1,13 @@
...
+
+/// @title LPRewardsEscrowBeneficiary
+/// @notice Transfer the received tokens from PhasedEscrow to a designated
+///         LPRewards contract.
+contract LPRewardsEscrowBeneficiary is StakerRewardsBeneficiary {
+    constructor(IERC20 _token, IStakerRewards _lpRewardsContract)
+        public
+        StakerRewardsBeneficiary(_token, _lpRewardsContract)
+    {}
+}
```
pdyraga, line 13: We do not need it, `StakingPoolRewardsEscrowBeneficiary` from https://github.com/keep-network/keep-core/pull/2230 will take care of it.

solidity/contracts/LPRewards.sol
nkuba: Does it mean that it has to implement `IStakingPoolRewards` instead of `IStakerRewards`?

pdyraga: Yeah, it can implement it. I'd say it does not have to but we are also fine with implementing it ;-)

pdyraga:
Tests are failing, I think we need to publish a new `keep-core` pre version and update the reference in dependencies.

solidity/test/LPRewardsTest.js
```diff
@@ -222,6 +230,168 @@ describe("LPRewards", () => {
...
+      )
+
+      return beneficiary
+    }
+  })
+
+  async function mintAndApproveWrappedTokens(token, address, staker, amount) {
+    await token.mint(staker, amount)
+    await token.approve(address, amount, {from: staker})
+  }
```
pdyraga, line 393: Is it used anywhere? Looks like some merge conflict resolution leftover.

pdyraga:
The code looks good, left one comment about probably unused function. As soon as the tests start passing, I am good to merge.

pdyraga (approved):

PR #620, opened 2020-11-20
Log not ready members on readiness phase timeout

lukasz-zimnoch:
Closes: #578 

This PR introduces additional logging to the readiness protocol. Specifically, addresses of not ready members are logged once the timeout is exceeded.

PR #654, opened 2020-12-28
Who is knocking: load operator addresses during readiness and announcement phases

### LINKED ISSUE #578: Log addresses of members the client is waiting for in readiness anouncment phase
We should enhance message logged on deadline exceeding when waiting for members' readiness with addresses of peers the client is waiting for.

Currently, we're logging:`failed to generate threshold signer: [readiness signaling protocol failed: [waiting for readiness timed out after: [2m0s]]]`

pdyraga:
~~Depends on https://github.com/keep-network/keep-common/pull/61~~
~~Depends on https://github.com/keep-network/keep-core/pull/2242~~

Closes #578

The goal of this change is to help the network operators understand what's happening when their node fails to complete the announcement or readiness phase. From now on, the address of each keep member who announced its presence/readiness is logged on INFO level. If not all keep members announced their presence/readiness, missing member addresses are logged on ERROR level.

A simple approach for the readiness phase was proposed in https://github.com/keep-network/keep-ecdsa/pull/620. The change proposed here implements it also for the announcement phase. Last but not least, it does not add an import to `github.com/ethereum/go-ethereum/crypto` in `tss`.

`keep-ecdsa` `node.go` is tightly-coupled to ethereum chain using `common.Address` as well as some other `github.com/ethereum/go-ethereum/common` references. This is wrong given `chain.Handle` can represent any host chain (Ethereum, Celo, ZCash,...) but it is how it is given how the understanding of requirements evolved. Eventually, we should get rid of all `go-ethereum` references from generic client/node code and use chain interface abstractions. The goal here was to do a minimal change that will let us convert pub key to address without tightly-coupling the code to any particular implementation even more. There will come a time for bigger refactoring of `node.go` and friends but not in this PR.


cmd/signing.go
```diff
@@ -197,6 +199,10 @@ func SignDigest(c *cli.Context) error {
 
 	signingOutcomesChannel := make(chan *signingOutcome, len(signers))
 
+	pubKeyToAddressFn := func(publicKey cecdsa.PublicKey) []byte {
```
nkuba, line 202: Why do we want to define this function here and pass it to the `CalculateSignature`? Can't we define it inside `tss` package? In `tss` package we have functions related to member identification https://github.com/keep-network/keep-ecdsa/blob/3099d5f5e8902839b17d5c847437355825173854/pkg/ecdsa/tss/member.go#L14-L16
And I believe it does the same thing as we define here.

pkg/ecdsa/tss/protocol_announce.go
```diff
@@ -44,9 +48,24 @@ func AnnounceProtocol(
 			case msg := <-announceInChan:
 				// Since broadcast channel has an address filter, we can
 				// assume each message come from a valid group member.
-				receivedMemberIDs[msg.SenderID.String()] = msg.SenderID
+				pubKey, err := msg.SenderID.PublicKey()
+				if err != nil {
+					logger.Errorf(
+						"could not get public key for member of keep [%v]",
+						keepAddress,
```
nkuba, line 55:
```suggestion
						"could not get public key for member [%s] of keep [%v]",
						msg.SenderID.String(),
						keepAddress,
```

pkg/ecdsa/tss/protocol_announce.go
```diff
@@ -44,9 +48,24 @@ func AnnounceProtocol(
 			case msg := <-announceInChan:
 				// Since broadcast channel has an address filter, we can
 				// assume each message come from a valid group member.
-				receivedMemberIDs[msg.SenderID.String()] = msg.SenderID
+				pubKey, err := msg.SenderID.PublicKey()
```
nkuba, line 51:
```suggestion
				publicKey, err := msg.SenderID.PublicKey()
```

pkg/ecdsa/tss/protocol_announce.go
```diff
@@ -81,6 +100,18 @@ func AnnounceProtocol(
 
 	switch ctx.Err() {
 	case context.DeadlineExceeded:
+		for _, member := range keepMemberAddresses {
+			_, announcedPresence := receivedMemberIDs[member]
```
nkuba, line 104: What about the current member? Will it be included in `keepMemberAddresses` and `receivedMemberIDs`?

pkg/ecdsa/tss/protocol_ready.go
```diff
@@ -86,6 +108,30 @@ func readyProtocol(
 
 	switch ctx.Err() {
 	case context.DeadlineExceeded:
+		for _, memberID := range group.groupMemberIDs {
```
nkuba, line 111: What about the current member? Will it be included in both `groupMemberIDs` and `readyMembers`?

pkg/ecdsa/tss/protocol_ready.go
```diff
@@ -97,3 +143,22 @@ func readyProtocol(
 		return fmt.Errorf("unexpected context error: [%v]", ctx.Err())
 	}
 }
+
+func memberIDToAddress(
+	memberID MemberID,
+	groupID string,
```
nkuba, line 149: We need this property to construct error message only. We shouldn't pass it.

pkg/ecdsa/tss/protocol_ready.go
```diff
@@ -97,3 +143,22 @@ func readyProtocol(
 		return fmt.Errorf("unexpected context error: [%v]", ctx.Err())
 	}
 }
+
+func memberIDToAddress(
```
nkuba, line 147: I think we should handle it as a function for `MemberID` and not pass `pubKeyToAddressFn`.

pkg/node/node.go
```diff
@@ -87,11 +87,23 @@ func (n *Node) AnnounceSignerPresence(
 		return nil, fmt.Errorf("failed to set broadcast channel filter: [%v]", err)
 	}
 
+	// TODO: Use generic type for representing address in node.go instead of
```
nkuba, line 90: Do we really need to convert the implementation to be chain agnostic? Isn't it too much effort and complication? I believe the only thing we wanted to achive was to log IDs of members that were not ready.

cmd/signing.go
pdyraga: If we define it inside `tss` we are tightly-coupling `tss` package with one particular way of calculating address from a public key. This way, `tss` will only work with one particular chain implementation, or a family of chain implementations using the same algorithm for pub key -> address conversion.

pkg/ecdsa/tss/protocol_announce.go
pdyraga: The current member also writes to this channel, so yes, it will be included there.

cmd/signing.go
nkuba: In TSS we should use `MemberID` as a way to identify members. We're creating the ID from PublicKey: https://github.com/keep-network/keep-ecdsa/blob/509a5a431c540ac58bb0e4245109e5ed6c10521a/pkg/node/node.go#L147
https://github.com/keep-network/keep-ecdsa/blob/509a5a431c540ac58bb0e4245109e5ed6c10521a/pkg/ecdsa/tss/protocol_announce.go#L79
Isn't it currently the same what we try to introduce here?

pkg/ecdsa/tss/protocol_ready.go
pdyraga: Correct. `groupMemberIDs` should contain identifiers of all group members and all members write to the broadcast channel.

pkg/ecdsa/tss/protocol_ready.go
pdyraga: That was the intention here. If we do not pass it, how to log it? Or shouldn't we log it? We can have multiple key generations and signings happening at the same time, having the context of keep address for which the operation failed could be helpful.

pkg/ecdsa/tss/protocol_ready.go
```diff
@@ -86,6 +108,30 @@ func readyProtocol(
...
+			)
+			if err != nil {
+				logger.Errorf(
+					"could not convert member ID to address: [%v]",
+					err,
+				)
+				break
+			}
+			_, announcedReadiness := readyMembers[memberAddress]
+			if !announcedReadiness {
```
dimpar, line 125: I think it can be simplified a bit:

```
if ok := readyMembers[memberAddress]; !ok {
	logger.Errorf(
		"member [%v] has not announced its readiness for keep [%v]; "+
			"check if keep client for that operator is active and "+
			"connected",
		memberAddress,
		group.groupID,
	)
}
```

cmd/signing.go
nkuba: Additionally isn't `elliptic.Marshal(publicKey.Curve, publicKey.X, publicKey.Y)` already a way of generic and chain agnostic identification?

pkg/ecdsa/tss/protocol_ready.go
pdyraga: This would be a good idea if there was only one way to convert pub key to address. Then, `MemberID` is just a simple value object. Here, we can have multiple strategies.

pkg/ecdsa/tss/protocol_ready.go
nkuba: We could add `groupID` when handling the error returned by this function, as we already do:
https://github.com/keep-network/keep-ecdsa/blob/509a5a431c540ac58bb0e4245109e5ed6c10521a/pkg/ecdsa/tss/protocol_ready.go#L61
https://github.com/keep-network/keep-ecdsa/blob/509a5a431c540ac58bb0e4245109e5ed6c10521a/pkg/ecdsa/tss/protocol_ready.go#L119

pkg/ecdsa/tss/protocol_ready.go
nkuba: > Here, we can have multiple strategies

We don't have them yet, right?

pkg/node/node.go
pdyraga: That's why I left a TODO, I do not want to address it in this PR. Right now, `node.go` can support only Ethereum or - to be precise - chains that have the same algorithm for converting pub key to address. Even if we assume we do not want to look into the future and think about other host chains now, the interfaces are broken on the code level anyway: why to abstract `chain.Handle` if we can't support any chain? Why to abstract chain if we import `go-ethereum` from `node.go` and other code "above" `chain.Handle`. A similar problem is in `keep-core` but it's localized to `pkg/operator`, something we can replace in one week. Here, the effort will be much more significant and my goal was not to make it even more significant, that's all.

pkg/ecdsa/tss/protocol_announce.go
```diff
@@ -44,9 +48,25 @@ func AnnounceProtocol(
 			case msg := <-announceInChan:
 				// Since broadcast channel has an address filter, we can
 				// assume each message come from a valid group member.
-				receivedMemberIDs[msg.SenderID.String()] = msg.SenderID
+				publicKey, err := msg.SenderID.PublicKey()
+				if err != nil {
+					logger.Errorf(
+						"could not get public key for member [%s] of keep [%v]: [%v]",
+						msg.SenderID.String(),
```
dimpar, line 55: Shouldn't formatting [%s] be enough to convert bytes to string? Do we need additional .String()?

pkg/ecdsa/tss/protocol_ready.go
pdyraga: That's a good idea, assuming we can express it clearly for operators who will have to read this error and act upon it. The best I could come up with is:
```
could not convert member ID to address for a member of keep [0x12345ABC]: [could not get public key for member with ID [987654321]: [something went wrong]
```
Already implemented in https://github.com/keep-network/keep-ecdsa/pull/654/commits/883121a3a1f0475068326085373fc110eec3c6db.

pkg/ecdsa/tss/protocol_announce.go
```diff
@@ -81,6 +101,18 @@ func AnnounceProtocol(
 
 	switch ctx.Err() {
 	case context.DeadlineExceeded:
+		for _, member := range keepMemberAddresses {
+			_, announcedPresence := receivedMemberIDs[member]
+			if !announcedPresence {
```
dimpar, line 106: Might be simplified a bit:
```
if _, ok := receivedMemberIDs[member]; !ok {
	logger.Errorf(
		"member [%v] has not announced its presence for keep [%v]; "+
			"check if keep client for that operator is active and "+
			"connected",
		member,
		keepAddress,
	)
}
```

pkg/ecdsa/tss/protocol_ready.go
pdyraga: The entire architecture is organized in a way supporting those strategies. We have two: ethereum and local. We can't ignore `chain.Handle` abstraction and work with ethereum implementation directly. This way, why to have `chain.Handle` at all - see [this](https://github.com/keep-network/keep-ecdsa/pull/654#discussion_r549380907) answer.

Importing anything `go-ethereum` from anything else then `ethereum.go` is bad practice, that's all.

pkg/node/node.go
```diff
@@ -87,11 +87,23 @@ func (n *Node) AnnounceSignerPresence(
...
+	// common.Address from go-ethereum. The current implementation is not
+	// host-chain implementation agnostic.
+	// To do not propagate ethereum-specific types any further, we convert
+	// addresses to strings before passing them to AnnounceProtocol.
+	keepAddressString := keepAddress.Hex()
+	keepMembersAddressesStrings := make([]string, len(keepAddress))
```
dimpar, line 96: not sure if I follow, `keepMembersAddressesStrings` slice needs to be a size of `len(keepAddress)`?

cmd/signing.go
pdyraga: I do agree, 

> In TSS we should use `MemberID` as a way to identify members. We're creating the ID from PublicKey:
> 
> https://github.com/keep-network/keep-ecdsa/blob/509a5a431c540ac58bb0e4245109e5ed6c10521a/pkg/node/node.go#L147
> 
> https://github.com/keep-network/keep-ecdsa/blob/509a5a431c540ac58bb0e4245109e5ed6c10521a/pkg/ecdsa/tss/protocol_announce.go#L79
> 
> Isn't it currently the same what we try to introduce here?

No, we are trying to log operator address, logging public key is not going to help, the information has to be very specific so that the operator can act and get the things right quickly. We are not changing the logic here, public key is still used as an identifier for all the communication. We just need to have a way to convert public key to address so that it can be printed to logs.

pdyraga: > Additionally isn't `elliptic.Marshal(publicKey.Curve, publicKey.X, publicKey.Y)` already a way of generic and chain agnostic identification?

We use operator address as an operator identifier. This would work only for chains that create operator addresses as public keys in (quoting) _uncompressed form specified in section 4.3.6 of ANSI X9.62._ The decision about how public key is mapped to address should be left to the chain, that's why we insert the strategy here.

pkg/chain/chain.go
```diff
@@ -15,13 +15,17 @@ import (
...
 	// StakeMonitor returns a stake monitor.
 	StakeMonitor() (chain.StakeMonitor, error)
 	// BalanceMonitor returns a balance monitor.
 	BalanceMonitor() (chain.BalanceMonitor, error)
 	// BlockCounter returns a block counter.
 	BlockCounter() chain.BlockCounter
+	// Signing returns a signer interface allowing to sign and verify messages
+	// using the chain implementation-specific mechanism as well as to
+	// convert between public key and address.
+	Signing() chain.Signing
```
nkuba, line 28: Maybe we should rename it to `Signer() chain.Signer`? We've got `StakeMonitor`, `BalanceMonitor`, `BlockCounter`, the returned interface is a Signer.

pdyraga: Yeah, I think it should be `Signer()` but `keep-core` has `Signing()` and wanted to keep it consistent.

pkg/node/node.go
pdyraga: Yes, we are transforming `[]common.Address` to `[]string`. In another language, I'd just write `keepAddresses.map(_.Hex())`.

pkg/ecdsa/tss/protocol_announce.go
pdyraga: Yes, from Go docs (https://golang.org/pkg/fmt/):
> 5. If an operand implements method String() string, that method will be invoked to convert the object to a string, which will then be formatted as required by the verb (if any). 

I've left `.String()` to keep it consistent with the rest of the code in `keep-ecdsa` but I am also fine with dropping it.

pkg/ecdsa/tss/protocol_announce.go
pdyraga: Done in https://github.com/keep-network/keep-ecdsa/pull/654/commits/e410b4d7c974b8d302c1e77cfea484636da18b37

pkg/ecdsa/tss/protocol_ready.go
pdyraga: Done in https://github.com/keep-network/keep-ecdsa/pull/654/commits/e410b4d7c974b8d302c1e77cfea484636da18b37

pkg/ecdsa/tss/protocol_ready.go
```diff
@@ -86,6 +109,27 @@ func readyProtocol(
...
+		for _, memberID := range group.groupMemberIDs {
+			memberAddress, err := memberIDToAddress(memberID, pubKeyToAddressFn)
+			if err != nil {
+				logger.Errorf(
+					"could not convert member ID to address for a member of "+
+						"keep [%v]: [%v]",
+					group.groupID,
+					err,
+				)
+				break
```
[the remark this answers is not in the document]
pdyraga: I was thinking about it but... I am not sure. If we can convert public key to address for one member, then something is terribly wrong and I think we should rather fail the readiness phase and start over again. WDYT?

pkg/node/node.go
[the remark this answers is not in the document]
pdyraga: 🤦

pkg/ecdsa/tss/protocol_ready.go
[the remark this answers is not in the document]
nkuba: Here the deadline already exceeded so we want to check which members were not ready. If we break here we won't check subsequent members.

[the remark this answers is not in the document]
pdyraga: Ah, that's the deadline case. Yeah, absolutely - good point.

pkg/node/node.go
dimpar: this is what I meant https://github.com/keep-network/keep-ecdsa/pull/654#discussion_r549397668 :)

pkg/ecdsa/tss/protocol_announce.go
dimpar: I'd vote for dropping it since it will be converted automatically.

dimpar:
Testing on deposit:

```
2020-12-29T13:50:59.135+0100	DEBUG	keep-net-libp2p	creating [32] subscription workers
2020-12-29T13:50:59.135+0100	DEBUG	keep-net-libp2p	creating [12] message workers
2020-12-29T13:50:59.137+0100	INFO	keep-tss	member [0xb7021acafecf289fa345ecc95ecab18077530841] from keep [0x7f8Da570e4bee41b0C24211a097b97917A8B5A1C] announced its presence
2020-12-29T13:50:59.233+0100	INFO	keep-tss	member [0x28bcfad75ee4c87b9089ba4c03a79b8148104da4] from keep [0x7f8Da570e4bee41b0C24211a097b97917A8B5A1C] announced its presence
2020-12-29T13:50:59.266+0100	INFO	keep-tss	member [0x65849fc91f8289b401e9d8f6d306b814b7f055c3] from keep [0x7f8Da570e4bee41b0C24211a097b97917A8B5A1C] announced its presence
2020-12-29T13:50:59.266+0100	INFO	keep-tss	announce protocol completed successfully
2020-12-29T13:50:59.266+0100	DEBUG	keep-tss	initialized key generation member: [{2,}]
2020-12-29T13:50:59.266+0100	INFO	keep-tss	[party:{2,}]: initialized key generation
2020-12-29T13:50:59.266+0100	INFO	keep-tss	signalling readiness
2020-12-29T13:50:59.267+0100	DEBUG	keep-net-libp2p	context is done; removing message handler
2020-12-29T13:50:59.268+0100	INFO	keep-tss	member [0x65849fc91f8289b401e9d8f6d306b814b7f055c3] from keep [0x7f8Da570e4bee41b0C24211a097b97917A8B5A1C] announced its readiness
2020-12-29T13:50:59.269+0100	INFO	keep-tss	member [0xb7021acafecf289fa345ecc95ecab18077530841] from keep [0x7f8Da570e4bee41b0C24211a097b97917A8B5A1C] announced its readiness
2020-12-29T13:50:59.287+0100	INFO	keep-tss	member [0x28bcfad75ee4c87b9089ba4c03a79b8148104da4] from keep [0x7f8Da570e4bee41b0C24211a097b97917A8B5A1C] announced its readiness
2020-12-29T13:50:59.287+0100	INFO	keep-tss	successfully signalled readiness
2020-12-29T13:50:59.287+0100	INFO	keep-tss	[party:{2,}]: starting key generation
```

pkg/ecdsa/tss/protocol_ready.go
```diff
@@ -86,6 +109,27 @@ func readyProtocol(
...
+			if err != nil {
+				logger.Errorf(
+					"could not convert member ID to address for a member of "+
+						"keep [%v]: [%v]",
+					group.groupID,
+					err,
+				)
+				continue
+			}
+			if ok := readyMembers[memberAddress]; !ok {
```
dimpar, line 123: actually I think it should be enough to just: `if readyMembers[memberAddress] {...}`

pdyraga: Heh, yeah... Fixed in https://github.com/keep-network/keep-ecdsa/pull/654/commits/3100c2313a1de3f284c7e1b30326bbf74290910a

pkg/ecdsa/tss/protocol_announce.go
```diff
@@ -40,6 +41,14 @@ func AnnounceProtocol(
 
 	receivedMemberIDs := make(map[string]MemberID) // member address -> memberID
 
+	markAnnounced := func(memberID MemberID, memberAddress string) {
+		receivedMemberIDs[strings.ToLower(memberAddress)] = memberID
```
nkuba, line 45: Do we want to introduce the same changes to signing stage?

pdyraga: Not sure if I follow the question but... announce phase is used only during key-generation. Readiness signaling is used both for key generation and signing. The problem with mixed cases occurred only for announcement protocol where member addresses were passed from the chain. For readiness, it is not a problem because we are reading addresses from locally stored information using the same code.

pkg/chain/chain.go
nkuba: We need to update it some day in both `keep-core` and `keep-ecdsa`. We could do it in https://github.com/keep-network/keep-core/pull/2242/ but now it's too late.

dimpar:
Tested on deposit when one of the nodes was shut down:

```
2020-12-29T16:07:23.046+0100	ERROR	keep-tss	member [0xB7021acafEcf289Fa345ecc95EcAb18077530841] has not announced its presence for keep [0x2A4ee8cFA1Df44B8384D4882eb792e041a32152A]; check if keep client for that operator is active and connected
2020-12-29T16:07:23.047+0100	WARN	keep-ecdsa	failed to announce signer presence: [waiting for announcements timed out after: [2m0s]]
2020-12-29T16:07:24.050+0100	INFO	keep-ecdsa	signer generation for keep [0x2A4ee8cFA1Df44B8384D4882eb792e041a32152A]; attempt [2]
2020-12-29T16:07:24.051+0100	INFO	keep-tss	announcing presence
2020-12-29T16:07:24.052+0100	INFO	keep-tss	member [0x28bcfad75ee4c87b9089ba4c03a79b8148104da4] from keep [0x2A4ee8cFA1Df44B8384D4882eb792e041a32152A] announced its presence
2020-12-29T16:07:24.053+0100	INFO	keep-tss	member [0x65849fc91f8289b401e9d8f6d306b814b7f055c3] from keep [0x2A4ee8cFA1Df44B8384D4882eb792e041a32152A] announced its presence
```

pkg/ecdsa/tss/protocol_announce.go
```diff
@@ -44,9 +57,26 @@ func AnnounceProtocol(
...
+						"could not get public key for member [%s] of keep [%v]: [%v]",
+						msg.SenderID.String(),
+						keepAddress,
+						err,
+					)
+					continue
+				}
 
-				if len(receivedMemberIDs) == membersCount {
+				memberAddress := "0x" + hex.EncodeToString(publicKeyToAddressFn(*publicKey))
```
nkuba, line 71: By adding `0x` we are tighting TSS implementation to ethereum specific prefixing.

pkg/ecdsa/tss/protocol_ready.go
```diff
@@ -48,7 +51,27 @@ func readyProtocol(
...
 					if msg.SenderID.Equal(memberID) {
-						readyMembers[msg.SenderID.String()] = true
+						memberAddress, err := memberIDToAddress(
+							memberID,
+							publicKeyToAddressFn,
+						)
+						if err != nil {
+							logger.Errorf(
+								"could not convert member ID to address for "+
+									"a member of keep [%v]: [%v]",
```
nkuba, line 61:
```suggestion
									"a member of keep [%s]: [%v]",
```

pkg/ecdsa/tss/protocol_ready.go
```diff
@@ -48,7 +51,27 @@ func readyProtocol(
...
+								group.groupID,
+								err,
+							)
+							break
+						}
+						readyMembers[memberAddress] = true
+
+						logger.Infof(
+							"member [0x%v] from keep [%v] announced its readiness",
```
nkuba, line 70: For consistency, we should drop `0x` prefixing in the message as we did already in https://github.com/keep-network/keep-ecdsa/pull/654/commits/a43b3ab873000944f5b55e92f85154ac974393b9#diff-7dac81c8ac9c4245caf522776d5d4bb167a7012110db13c5a702b365b9b80f1aL64
```suggestion
							"member [%v] from keep [%s] announced its readiness",
```

pkg/ecdsa/tss/protocol_ready.go
```diff
@@ -86,6 +109,27 @@ func readyProtocol(
...
+					"could not convert member ID to address for a member of "+
+						"keep [%v]: [%v]",
+					group.groupID,
+					err,
+				)
+				continue
+			}
+			if !readyMembers[memberAddress] {
+				logger.Errorf(
+					"member [0x%v] has not announced its readiness for keep [%v]; "+
```
nkuba, line 125: For consistency, we should drop `0x` prefixing in the message as we did already in https://github.com/keep-network/keep-ecdsa/pull/654/commits/a43b3ab873000944f5b55e92f85154ac974393b9#diff-7dac81c8ac9c4245caf522776d5d4bb167a7012110db13c5a702b365b9b80f1aL64
```suggestion
					"member [%s] has not announced its readiness for keep [%s]; "+
```

pkg/ecdsa/tss/protocol_ready.go
```diff
@@ -86,6 +109,27 @@ func readyProtocol(
 
 	switch ctx.Err() {
 	case context.DeadlineExceeded:
+		for _, memberID := range group.groupMemberIDs {
+			memberAddress, err := memberIDToAddress(memberID, publicKeyToAddressFn)
+			if err != nil {
+				logger.Errorf(
+					"could not convert member ID to address for a member of "+
+						"keep [%v]: [%v]",
```
nkuba, line 117:
```suggestion
						"keep [%s]: [%v]",
```

pkg/ecdsa/tss/protocol_ready.go
```diff
@@ -97,3 +141,19 @@ func readyProtocol(
...
+
+func memberIDToAddress(
+	memberID MemberID,
+	publicKeyToAddressFn func(cecdsa.PublicKey) []byte,
+) (string, error) {
+	pubKey, err := memberID.PublicKey()
+	if err != nil {
+		return "", fmt.Errorf(
+			"could not get public key for member with ID [%v]: [%v]",
+			memberID.String(),
```
nkuba, line 153: Will `%s` work?
```suggestion
			"could not get public key for member with ID [%s]: [%v]",
			memberID,
```

pkg/ecdsa/tss/protocol_announce.go
pdyraga: 0x used here has nothing to do with Ethereum prefixing, it's how hexadecimals are [usually](https://stackoverflow.com/questions/2670639/why-are-hexadecimal-numbers-prefixed-with-0x) prefixed and because converter currently represents messages as byte array it makes sense. The approach abstracting address conversion presented in this PR is not a final solution. For example, it wouldn't work well with BTC addresses. It is a step forward but there is still work ahead of us. I do not think the address should be represented as a byte array eventually. I'd rather go with just a string.

pkg/ecdsa/tss/protocol_ready.go
```diff
@@ -97,3 +141,19 @@ func readyProtocol(
...
+	pubKey, err := memberID.PublicKey()
+	if err != nil {
+		return "", fmt.Errorf(
+			"could not get public key for member with ID [%s]: [%v]",
+			memberID,
+			err,
+		)
+	}
+
+	return "0x" + hex.EncodeToString(publicKeyToAddressFn(*pubKey)), nil
```
dimpar, line 158: nitpicking but I think it can be formatted like so: `return fmt.Sprintf("0x%v", hex.EncodeToString(publicKeyToAddressFn(*pubKey))), nil`

dimpar (approved):

pkg/ecdsa/tss/protocol_announce.go
nkuba: As you mentioned this solution this solution is not generic and won't work well for all chains. We will have to revisit it later.

nkuba (approved):

PR #658, opened 2020-12-31
Replaced ::set-env with env file in npm publishing action config

nkuba:
GitHub actions stopped supporting `::set-env` to set environment
property and require to store variables in file. We updated the
configuration to reflect it.
See: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

### DIFF
--- a/.github/workflows/npm.yml
+++ b/.github/workflows/npm.yml
@@ -2,12 +2,12 @@ name: NPM
 
 on:
   push:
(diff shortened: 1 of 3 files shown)


nkuba:
Closing in favour of https://github.com/keep-network/keep-ecdsa/pull/659

PR #659, opened 2020-12-31
Unchoke solidity contracts `pre` package publication on master merge

nkuba:
Here we update the script for publishing the NPM package in the GitHub Actions job.

The script for publishing is defined also in [Circle CI `publish_npm_package` job](https://github.com/keep-network/keep-ecdsa/blob/a93de9a4d09c3c78a6db10db2e7c244392f05b46/.circleci/config.yml#L193-L236) and we use it there to publish contracts after `keep-test` migration on the release candidate.

We used to publish `pre` packages on `keep-dev` migration but this environment is no longer maintained and migration process to `keep-dev` was disabled in https://github.com/keep-network/keep-ecdsa/pull/442.

We want to publish a `pre` package on each `master` merge. But there is no sense in doing it if nothing has changed in the
solidity contracts. Circle CI doesn't allow us to filter on changed content to run a job, that's why we use GitHub Actions. 
We will publish a new `pre` package to the NPM registry on master merge only if solidity contracts have changed.

We also switched from publishing to GitHub's private package registry to the [npm registry](https://www.npmjs.com).
Publishing to the NPM registry requires an `NPM_TOKEN` to be configured in GH project's secrets variables.


nkuba:
The job is not passing yet. We need to configure `NPM_TOKEN` to access npm registry. After that we need to remove comments around filters that were added for testing.

nkuba:
The job passed in https://github.com/keep-network/keep-ecdsa/pull/659/checks?check_run_id=1645805314

.github/workflows/npm.yml
```diff
@@ -11,38 +11,47 @@ on:
...
+          # Update package.json with the latest published package version matching this
+          # preid to prepare for bumping.
+          echo $(jq -M ".version=\"${latest_version}\"" package.json) > package.json
+
+          # Bump without doing any git work. Versioning is a build-time action for us.
+          # Consider including commit id? Would be +<commit id>.
+          npm version prerelease --preid=$preid --no-git-tag-version
+
+          # Fix resolved dependency versions.
+          npm update
```
nkuba, line 51: @Shadowfiend This part was copied from CircleCI's config. I'm not sure if we really want to update versions of the dependencies? Shouldn't we run `npm ci` command? 
As I understand, with `npm update` we will resolve new versions for dependencies and publish the package with them. We won't have a matching `package-lock.json` in the repository.

nkuba: Or maybe we don't need `npm update` nor `npm ci` at all?

nkuba: As we need to resolve latest versions of `pre` and `rc` prereleeases for `@keep-network` contracts we should still use `npm update`. Although we could pull updating part to the earliest stage of the process and update only specific packages we use: `@keep-network/keep-core` and `@keep-network/sortition-pools`.

lukasz-zimnoch (approved):

PR #663, opened 2021-01-07
Subscription dance: backoff and logging

pdyraga:

Depends on https://github.com/keep-network/keep-common/pull/62

The referenced `keep-common` PR solves three problems and two of them are important for ECDSA client.

Currently, the subscription mechanism used by ECDSA client does not implement a backoff. Each failed subscription is retried 5 seconds after it failed. For many individual subscriptions ECDSA client is opening, a massive retry with no backoff can be interpreted as misbehavior by a third party Ethereum provider and further attempts can be completely blocked. Some operators already experienced it with Alchemy.

The second problem is that all the logging regarding resubscriptions is done on the warning level. With the lack of backoff mentioned in the previous point, this can produce a deadly mixture - keep client could be trying to reconnect for a long time, Ethereum client could be rejecting those attempts interpreting them as DoS/misbehavior, and in this scenario, operator would receive only warnings with no single error. 

Both those points are addressed here by bumping up the dependency on `keep-common` and regenerating contract bindings. Also, the code in `ethereum.go` is getting cleaned up a little to reflect the fact all `Watch*` attempts are getting retried until they succeed or until unsubscribe happens.

For example, let's say that the client was able to establish all subscriptions, they were working fine for two minutes, failed, and were established again successfully. Dropped WS connection is fine assuming it's not happening too often. Two minutes is certainly alarming, so the client will now alarm with an error:

> 2021-01-07T18:29:20.522+0100	ERROR	keep-contract-BondedECDSAKeepFactory	subscription to event BondedECDSAKeepCreated had to be retried [2m0.006063728s] since the last attempt; please inspect Ethereum client connectivity

Now, let's say that the retry will fail. Again, the client will alarm with an error:

> 2021-01-07T18:31:21.532+0100	ERROR	keep-contract-BondedECDSAKeep	subscription to event KeepClosed failed: [dial tcp 127.0.0.1:8546: connect: connection refused]; resubscription attempt will be performed

Last but not least, we now have a backoff. The delay is increased twice until it reaches the maximum backoff time which is set to 2 minutes.


lukasz-zimnoch (approved):
All changes look good! I think we can merge once all comments in https://github.com/keep-network/keep-common/pull/62 are addressed.

nkuba (commented):
All looks fine. I'll hold with approval until we resolve all comments from https://github.com/keep-network/keep-common/pull/62

nkuba (approved):

PR #667, opened 2021-01-11
Staker rewards: Undelegating operators

lukasz-zimnoch:
If an operator doesn't meet the sortition pool requirement at the interval start, it may be the case that they are currently undelegating. In such a case, that operator should not be punished for requirements violation. Instead of that, only the `ETH_bonded` value from their still active keeps should be counted to their `ETH_locked` value. Their `ETH_unbonded` value should be ignored ignored.


staker-rewards/lib/rewards-calculator.js
```diff
@@ -26,12 +26,24 @@ export default class RewardsCalculator {
...
+      if (
+        requirementsViolations.includes("poolRequirementFulfilledAtStart") &&
+        (await this.isUndelegatingOperator(operatorParameters.operator))
+      ) {
+        requirementsViolations = requirementsViolations.filter(
+          (item) => item !== "poolRequirementFulfilledAtStart"
+        )
+
+        ethTotal = operatorParameters.operatorAssets.ethBonded
+      }
```
pdyraga, line 45: This condition does not cover the situation when someone is undelegating but is still in the sortition pool because the sortition pool had no chance to remove that operator yet.

The way it should work:
- if you are undelegating, you are not required to be in the sortition pool
- if you are undelegating, only ETH_bonded counts to ETH_total

Right now, it is:
- if you are undelegating, you are not required to be in the sortition pool
- if you are undelegating and if you are not in the sortition pool, only ETH_bonded counts to ETH_total

staker-rewards/lib/rewards-calculator.js
```diff
@@ -108,6 +120,21 @@ export default class RewardsCalculator {
...
+    const tokenStaking = await this.context.contracts.TokenStaking.deployed()
+
+    const delegationInfo = await callWithRetry(
+      tokenStaking.methods.getDelegationInfo(operatorAddress),
+      block
+    )
+
+    const undelegatedAt = new BigNumber(delegationInfo.undelegatedAt).toNumber()
+
+    return undelegatedAt !== 0 && this.interval.start > undelegatedAt
```
pdyraga, line 135: Do we need `this.interval.start > undelegatedAt` check?

`getDelegationInfo` is checked for `this.interval.startBlock`. If at that time operator was undelegating, I think it's enough to return `true` from this function.

This check would be needed if we were not calling `getDelegationInfo` at the specific block.

pdyraga (changes requested):

staker-rewards/lib/rewards-calculator.js
lukasz-zimnoch: Yeah right.

staker-rewards/lib/rewards-calculator.js
lukasz-zimnoch: Good catch. I will change.

staker-rewards/lib/rewards-calculator.js
```diff
@@ -108,6 +121,21 @@ export default class RewardsCalculator {
...
   }
 
+  async isUndelegatingOperator(operatorAddress) {
+    const block = this.interval.startBlock
+
+    const tokenStaking = await this.context.contracts.TokenStaking.deployed()
+
+    const delegationInfo = await callWithRetry(
+      tokenStaking.methods.getDelegationInfo(operatorAddress),
+      block
```
pdyraga, line 131: Can we use `this.interval.startBlock` here directly? I think it will be easier to read, especially that `block` is used only in one place.

staker-rewards/lib/rewards-calculator.js
```diff
@@ -108,6 +121,21 @@ export default class RewardsCalculator {
...
+    )
+
+    const undelegatedAt = new BigNumber(delegationInfo.undelegatedAt).toNumber()
```
pdyraga, line 134: Why do we need to do this transformation? I am not fully following it.

lukasz-zimnoch: Yeah, we can directly compare to `"0"` because we are no longer checking the `undelegatedAt` against interval start timestamp.

staker-rewards/lib/rewards-calculator.js
lukasz-zimnoch: Sure

staker-rewards/test/rewards-calculator.test.js
```diff
@@ -301,6 +320,37 @@ describe("rewards calculator", async () => {
...
+      )
+
+      const rewards = rewardsCalculator.getOperatorRewards(
+        "0x88Ed51f84c21Ae246c23137D090cdF441009D916"
+      )
+
+      assert.equal(rewards.totalRewards.isEqualTo(new BigNumber(0)), false)
+    }
+  )
+
```
pdyraga, line 353: Let's add a test where we'll cover this logic from `rewards-calculator.js`:
```
        operatorParameters.operatorAssets.ethTotal =
          operatorParameters.operatorAssets.ethBonded
```

pdyraga (commented):
How hard is it to add `undelegated` column to the summary table, just to make it clear for stakers why not all of their ETH is taken into account?

staker-rewards/lib/rewards-calculator.js
```diff
@@ -26,12 +26,25 @@ export default class RewardsCalculator {
...
+          requirementsViolations.includes("poolRequirementFulfilledAtStart")
+        ) {
+          requirementsViolations = requirementsViolations.filter(
+            (item) => item !== "poolRequirementFulfilledAtStart"
+          )
+        }
+
+        operatorParameters.operatorAssets.ethTotal =
+          operatorParameters.operatorAssets.ethBonded
+      }
```
pdyraga, line 44: I do not think it's correct as it does not include ETH that was withdrawn. When calculating assets, we do:
https://github.com/keep-network/keep-ecdsa/blob/e5132c4f0235649196af6b5d9fe7ef0bb59c839f/staker-rewards/lib/assets-calculator.js#L58

I even do wonder if the check for undelegating operator shouldn't be done in `assets-calculator.js` as well, so that we keep all the logic about `ethTotal` calculations there.

lukasz-zimnoch: But withdraws are about the unbonded value. I don't think we should subtract the unbonded value withdrawals from the bonded value sum. Am I right?

lukasz-zimnoch: If we move this to the `assets-calculator.js` we will have some duplications because we have to check undelegations in two places:
- in the `assets-calculator` to do `ethTotal = ethBonded`
- in the `rewards-calculator` to remove the sortition pool condition violation if present

If it is acceptable we can do it.

pdyraga:
I opened a PR to reflect this change in the README: https://github.com/keep-network/keep-ecdsa/pull/668.

staker-rewards/lib/rewards-calculator.js
pdyraga: > But withdraws are about the unbonded value. I don't think we should subtract the unbonded value withdrawals from the bonded value sum. Am I right?

Correct, though bonded could become unbonded and then withdrawn during the interval. That's why we always need to subtract.


> If we move this to the `assets-calculator.js` we will have some duplications because we have to check undelegations in two places

I think this is acceptable. We'll have to repeat `isUndelegating` check so there is a cost of an additional request but the structure will be cleaner. To me, `assets-calculator.js` reflects *Operator Parameters* section in the readme while `rewards-calculator.js` validates everything in *Requirements* section.

lukasz-zimnoch:
@pdyraga all comments addressed. I think we could make final testing on mainnet.

pdyraga:
Everything looks good, I'd change just one thing: rename `isUndelegating` to `undelegated` in column headers. We may have a situation when someone undelegated and recovered. We may also have a situation when someone undelegated but not yet recovered. The first one is still undelegating, the second one completed the process by recovering the stake. `undelegated` would better describe the state, especially that we display both cases in the summary table.

pdyraga (approved):

PR #668, opened 2021-01-13
Update staker rewards doc about undelegating operators

pdyraga:
Operators who are undelegating should still get rewards for their ETH
but only for the portion of ETH that is currently bonded.

This is just a documentation update, no changes in the script generating
the allocation. Those changes are being implemented in https://github.com/keep-network/keep-ecdsa/pull/667.


pdyraga:
Let's hold on with merging this PR until #667 is approved.

lukasz-zimnoch (approved):

PR #670, opened 2021-01-15
RFC 2: tBTC liquidation recovery functionality

Shadowfiend:
The outline and required inputs are all clearly listed out, but the juiciest
parts are still stubs, of course; namely, how to compute all the inputs that
require computation.


pdyraga:
So far it looks very good, all inputs are very well explained. 👏 As stated in the PR description, the juiciest parts are still stubs, so just adding some early thoughts on those juicy details.

I like the RFC does not require operators to run a BTC node. That would be a significant effort for some of them. Since this is the low-effort mechanism, can we just output the signed transaction to logs (or to some file, if really needed) and let operators do the broadcast on their side? If we document a way of how to do that in the RFC, this should be enough in my opinion.

Regarding the action triggering this whole mechanism, I was first thinking about having a simple CLI command in the client, similar to the signing or key material share decryption action, where the operator would put beneficiaries and keep address as an input. However, this seems wrong for a variety of reasons. First, there is no clear way of how operators should coordinate on when to trigger this action. Secondly, signing requires peer discovery, so this command should send a request to a running client instance instead of triggering some action as a completely separate process. Finally, there is no proven track of inputs operators provided.

Given that we need to get some data from tBTC deposit contract, then get some data from ECDSA keep contract, each operator needs to provide some data, and all of it should be triggerable only for terminated (or also closed if we want to handle misfunds) keeps, we can implement an on-chain contract for coordinating all of it. Operators would submit their beneficiary addresses and as soon as all beneficiaries are known, the contract would emit an event, application-specific actions in the client would react on. 

One may argue that implementing a contract adds some work but I do not think it's a significant effort and getting some data on-chain can be easier than computing them in the client. Secondly, we are not implementing any new mechanism for triggering client actions - it is just an event subscription so it's in fact less work on the client-side. Last but not least, operators would have a trustless way to coordinate and track the progress of recovery. Finally, there would be a proven track of BTC beneficiaries provided.

docs/rfc/rfc-2.adoc
```diff
@@ -0,0 +1,220 @@
...
+If an optional Bitcoin node connection is available, the client should
+additionally check a derived address for a balance before publishing. A fallback
+could be implemented that checks
+`https://blockstream.info/api/address/<address>` for a given derived address to
+ensure `tx_count` is 0.
+
+[[bitcoin-fee]]
+===== Dealing with the Bitcoin Fee
+
+// Each client sets a max fee per vbyte in config. Signing exchanges each
```
nkuba, line 179: Why some sections are commented out?

docs/rfc/rfc-2.adoc
```diff
@@ -0,0 +1,220 @@
...
+// client's max fee. The minimum of all the max fees is used. A default that
+// would overpay for confirmation in 100 (?) blocks under most circumstances
+// should be used if the operator does not specify another maximum. If an
+// optional Bitcoin node connection is present, the client should override the
+// max fee in config using the suggested fee from the Bitcoin node. Possible
+// default fallback if no Bitcoin node available: attempt to GET fee estimates
+// from `https://blockstream.info/api/fee-estimates`.
+
+[[signing]]
+===== Signing
```
nkuba, line 189: How is the process initiated? Is it enough that one operator executes a command on the client? Do we want the operator to call the contract? Is it starting automatically after keep termination?

docs/rfc/rfc-2.adoc
```diff
@@ -0,0 +1,220 @@
...
+ensure `tx_count` is 0.
+
+[[bitcoin-fee]]
+===== Dealing with the Bitcoin Fee
+
+// Each client sets a max fee per vbyte in config. Signing exchanges each
+// client's max fee. The minimum of all the max fees is used. A default that
+// would overpay for confirmation in 100 (?) blocks under most circumstances
+// should be used if the operator does not specify another maximum. If an
+// optional Bitcoin node connection is present, the client should override the
```
nkuba, line 183: Regarding the Bitcoin node connection, we should think of APIs we want to support, e.g. electrum, blockstream, blockcypher, ....

docs/rfc/rfc-2.adoc
```diff
@@ -0,0 +1,220 @@
...
+// max fee in config using the suggested fee from the Bitcoin node. Possible
+// default fallback if no Bitcoin node available: attempt to GET fee estimates
+// from `https://blockstream.info/api/fee-estimates`.
+
+[[signing]]
+===== Signing 
+
+// Round 1: exchange addresses and max fees. Reminder that underlying transport
```
nkuba, line 191: As clients exchange addresses they should also validate its correctness.

nkuba (commented):
I like the presented ideas! I left some comments/suggestions.

Having a BTC beneficiary address configurable in a contract would be favourable for transparency. Configuring such an important value in a config file makes me feel anxious.

docs/rfc/rfc-2.adoc
nkuba: It should be fairly easy for an operator to use publicly accessible APIs like blockcypher or blockstream.

docs/rfc/rfc-2.adoc
```diff
@@ -0,0 +1,220 @@
...
+===== Computing the Deposit Output Script
+
+[[recovery-address]]
+===== Specifying an Operator's Bitcoin Recovery Address
+
+To specify a given operator's Bitcoin Recovery Address, this proposal suggests
+adding a single key to the `Extensions.TBTC` config entry, `BTCRefundTarget`.
+The value for this key should be one of:
+
+- A simple BTC address.
```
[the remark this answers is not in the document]
Shadowfiend: We should be flexible here; since we don't need to prove the transaction on the Ethereum side, we don't need to worry as much about type.

docs/rfc/rfc-2.adoc
```diff
@@ -0,0 +1,220 @@
...
+- Have a *pub that is only meant for use by the client, and have the client
+  track the latest address generated (as an index starting at 0).
+
+To avoid _strictly requiring_ a connection to a Bitcoin node, this RFC proposes
+tracking the latest address generated, and allowing an additional config key,
+`BTCRefundTargetDerivationStartIndex`, that sets a starting point. The client
+should write the updated index to the `current/` directory that tracks
+active key shares, to an encrypted file named
+`current-btc-refund-derivation-index`. The operator should then seek to obtain
+from the staker an extended public key that is only used for this purpose.
```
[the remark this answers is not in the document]
Shadowfiend: It's complicated, but the stakers who will care about this should be able to do it. I don't consider it a primary concern of ours to handle this, though digging up a tutorial or two and linking them from our staker docs would be a good idea once this is out the door.

We do already know several stakers who have provided such extended keys when coordinating recovery right now (others have opted for a single address).

docs/rfc/rfc-2.adoc
Shadowfiend: These are the “I need to fill in more detail, but here are my high-level thoughts” bits. Also why this is still a draft 😅

I'm ~80% of the way through filling in a lot of these, should be able to push later this morning.

docs/rfc/rfc-2.adoc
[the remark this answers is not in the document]
nkuba: Just to confirm, a transaction can have multiple outputs with whichever type of address, right?

docs/rfc/rfc-2.adoc
Shadowfiend: It's going to be solely blockstream. I have more detail on this in my push this morning.

docs/rfc/rfc-2.adoc
Shadowfiend: Realized after the fact that I missed this, but yes, the idea is automatically after termination with an extended timeout window and probably a less aggressive retry timer than regular operations.

Shadowfiend:
Actually just pushed what I had last night. More detail later today, but what I've already added is important to clarify a few edges.

lukasz-zimnoch (commented):
Existing parts are clear and consistent. I think we have to make it explicit how the process will start. Using the Piotr's on-chain contract idea I think we have to:

1. Create a new tBTC extension listening on `Liquidation` (and maybe some other events to handle misfunds) from the `TBTCSystem` contract
2. Once such an event arrives, interested signers begin to exchange fee rates and post them along with their beneficiary addresses to the new contract.
3. The tBTC extension waits for an event from the new contract. 
4. Once all signers post their fees and addresses, the new contract emits an event as a signal to begin the signing process

Am I right?

docs/rfc/rfc-2.adoc
```diff
@@ -0,0 +1,261 @@
...
+  track the latest address generated (as an index starting at 0).
+
+To avoid _strictly requiring_ a connection to a Bitcoin node, this RFC
+proposes tracking the latest address generated, and allowing an additional
+config key, `TargetDerivationStartIndex`, that sets a starting point for such
+derivations. The client should write the updated index to the `current/`
+directory that tracks active key shares, to an encrypted file named
+`current-btc-refund-derivation-index`. The operator should then seek to
+obtain from the staker an extended public key that is only used for this
+purpose.
```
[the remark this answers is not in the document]
Shadowfiend: Yeah, I didn't get a chance to address that piece in @pdyraga's notes above: I'm fairly opposed to having a contract in the mix here. We have a clear trigger for when to start this with existing contracts, and contracts introduce cost, adaptability, and maintainability concerns that I suggest we avoid unless we have a security motivation to do otherwise. The contracts can't enforce anything in this scenario anyway, so using them strictly for data coordination seems unnecessary. Using them for operator-specific data tracking is even less desirable IMO---that info should remain private.

Lastly, we probably shouldn't put a *pub on-chain, as it's considered less secure than an address by a fair margin. If we don't have the *pub on-chain, having the derivation index on-chain seems problematic, e.g. if the staker decides to give the operator a different address.

Shadowfiend:
(Quick note: that last commit is messy and in-flight to make sure it's not sitting on one hard drive ready to be lost. Will be cleaning it up soon.)

pdyraga:
> Regarding the action triggering this whole mechanism, I was first thinking about having a simple CLI command in the client, similar to the signing or key material share decryption action, where the operator would put beneficiaries and keep address as an input. However, this seems wrong for a variety of reasons. First, there is no clear way of how operators should coordinate on when to trigger this action. Secondly, signing requires peer discovery, so this command should send a request to a running client instance instead of triggering some action as a completely separate process. Finally, there is no proven track of inputs operators provided.
> 
> Given that we need to get some data from tBTC deposit contract, then get some data from ECDSA keep contract, each operator needs to provide some data, and all of it should be triggerable only for terminated (or also closed if we want to handle misfunds) keeps, we can implement an on-chain contract for coordinating all of it. Operators would submit their beneficiary addresses and as soon as all beneficiaries are known, the contract would emit an event, application-specific actions in the client would react on.
> 
> One may argue that implementing a contract adds some work but I do not think it's a significant effort and getting some data on-chain can be easier than computing them in the client. Secondly, we are not implementing any new mechanism for triggering client actions - it is just an event subscription so it's in fact less work on the client-side. Last but not least, operators would have a trustless way to coordinate and track the progress of recovery. Finally, there would be a proven track of BTC beneficiaries provided.

Thinking more about it and reading @Shadowfiend's comments I am not longer sure what I proposed is an optimal long-term solution.

An advantage of having a contract and building a UI around it would be to allow stakers (owners) to track the progress. We have some stakers in the network with 10+ liquidated deposits and keeping an eye on them requires communication with the team and having an operator willing to track those deposits and doing this communication.

Another advantage would be that we can probably reuse some `tbtc.js` parts to construct BTC transactions without having to implement it in the client.

Finally, querying for pending recoveries upon client restart could be simpler. Instead of having to go through all keeps and checking if they are terminated, we could be calling one contract with a function optimized for this purpose.

However...

 With application-specific actions in the client, monitoring scripts ran by the community, and the development of external collateralization management services, having a lot of liquidated deposits for a single staker should hopefully no longer be a problem in the future. 

Querying keeps one by one needs to be done anyway because we are checking for pending signatures.

This contract would not provide any additional security guarantees as it is not enforcing where tokens are finally transferred.

Finally, what's the most important here IMO: having it all in the client does not require any action to be taken by operators. Ether cost is one thing but more importantly, other signing group members will not suffer from one operator not executing their on-chain part. The fact we eliminate manual actions is a big win.

Shadowfiend:
Ok, open season folks, officially marking this Ready For Review™. The Implementation Details section now includes several mostly-complete functions to do the most important Bitcoin manipulations needed to make this work. The Implementation Behaviors section steps through higher-level bits like key material archiving changes, startup behavior, Bitcoin fee resolution, adjustments to the signing protocol, and the method to submit signed transaction hex.

All sections that could benefit from a connection to a Bitcoin node specify optional behaviors in cases where a Blockstream Electrs fork is available for usage. Usage defaults to Blockstream's public API unless explicitly disabled in configuration*.

Some high-level bits of feedback I'm looking for specifically:

- Thoughts on everything under “Implementation Behaviors”.
- Thoughts on the optional Bitcoin node functionality.
- Anything that feels like it's missing or unclear in the RFC.
- Any missing code or other detail in the Implementation Details section.

As always, any other thoughtful feedback is also welcome :)

*- do we want to default this to on, or default it to off and strongly encourage it in the sample config?

docs/rfc/rfc-2.adoc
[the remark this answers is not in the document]
Shadowfiend: In liquidation recoveries they should always have 3 outputs, and each output can be a different type of address.

docs/rfc/rfc-2.adoc
Shadowfiend: These are now done.

docs/rfc/rfc-2.adoc
Shadowfiend: This is now filled in.

docs/rfc/rfc-2.adoc
Shadowfiend: I missed this when I fleshed it out, will add it now.

docs/rfc/rfc-2.adoc
```diff
@@ -0,0 +1,789 @@
...
+- Be able to check, on-chain, whether a given address has been used; or,
+- Have a *pub that is only meant for use by the client, and have the client
+  track the latest address generated (as an index starting at 0).
+
+To avoid _strictly requiring_ a connection to a Bitcoin node, this RFC
+proposes tracking the latest address generated, and allowing an additional
+config key, `TargetDerivationStartIndex`, that sets a starting point for such
+derivations. The client should write the updated index to the `current/`
+directory that tracks active key shares, to an encrypted file named
+`current-btc-refund-derivation-index`. The operator should then seek to
```
pdyraga, line 165: If the storage contains updated index, is `TargetDerivationStartIndex` in config ignored? Can we run into a situation when the value held by the storage is not up to date? For example, the situation when the receiving party transferred BTC they obtained somewhere else.

docs/rfc/rfc-2.adoc
```diff
@@ -0,0 +1,795 @@
...
+obtain from the staker an extended public key that is only used for this
+purpose.
+
+If an optional Bitcoin node connection is available (see <<electrs>>), the
+client should additionally check a derived address for a balance before
+publishing. To do this, a GET request to `/address/<address>` should be issued,
+and the JSON response inspected for a `tx_count` of 0.
```
pdyraga, line 172: What if they do not match? Is the protocol getting aborted?

docs/rfc/rfc-2.adoc
```diff
@@ -0,0 +1,795 @@
...
+2020-02-01T12:12:12.12-0800    WARN   keep-ecdsa   Please broadcast Bitcoin transaction 0010010101000828181124af234gbawed000
+```
+
+The first three components are the standard components emitted by the ECDSA
+client's logging library, while the message itself begins with `Please broadcast
+Bitcoin transaction`, followed by the hexadecimal string representing the signed
+transaction. The signed transaction can be broadcast via manual submission to a
+known node, via API access to a service provider like BlockCypher or
+Blockstream, or via a manual form submission on a page like
+https://live.blockcypher.com/btc/pushtx/ .
```
pdyraga, line 300: Is there no way to broadcast this TX automatically even if `Extensions.TBTC.ElectrsURL` is set, correct?

docs/rfc/rfc-2.adoc
Shadowfiend: Done.

docs/rfc/rfc-2.adoc
Shadowfiend: I'll clarify that the way to disambiguate between storage value and config value is to use the higher of the two.

I'm not sure I follow the scenario you're describing for the second question. The value held by the storage could be ”out of date“ in the sense of pointing to an already-used derived address if the recipient is using the same extended pubkey chain for more than just this purpose, or if there's a concurrency issue (feels solvable by an atomically incrementing int). Otherwise I don't see a clear way for that to happen. The optional Bitcoin node connection can help for the former scenario (subject to concurrency there as well, ofc).

docs/rfc/rfc-2.adoc
Shadowfiend: I can clarify; in that case the client should increment the derivation index and try again.

docs/rfc/rfc-2.adoc
Shadowfiend: The tx can be broadcast automatically if there's an optional Bitcoin node connection, next paragraph describes it.

docs/rfc/rfc-2.adoc
pdyraga: > if the recipient is using the same extended pubkey chain for more than just this purpose

This is the scenario that I meant.

Shadowfiend: Yeah, there's only so much we can do there. In my mind it would make the most sense to recommend usage of an extended pubkey chain that is not used elsewhere, or to provide a dedicated branch of the derivation tree (e.g. `/0` is generally for external transactions and `/1` for internal ones, but you could provide `/2` if you wanted a dedicated chain). The function in the derivation implementation details does allow for this:

```
// In cases where the extended public key is at depth 4, meaning the external or
// internal chain is already included, DeriveAddress will directly derive the
// address index at the existing depth.
```

docs/rfc/rfc-2.adoc
Shadowfiend: This is now clearly articulated.

docs/rfc/rfc-2.adoc
Shadowfiend: Clarified this.

docs/rfc/rfc-2.adoc
pdyraga: Cool. Just to clarify: if the operator provides a simple BTC address as `BeneficiaryAddress`, receive there BTC from a liquidated deposit, and transfer this BTC out of this account to some other account. Do they have to change `BeneficiaryAddress` in the config for further recovery of BTC from a liquidated deposit? My understanding is that the answer is no, they do not have to change anything. I am not sure about *pub though.

docs/rfc/rfc-2.adoc
```diff
@@ -0,0 +1,799 @@
...
+Bitcoin node connection is available (see <<electrs>>). In this case, the
+client should check the keep's associated deposit for an unspent balance by
+issuing a GET request to `/tx/:txid/outspend/:vout`, using the deposit's
+`fundingInfo().utxoOutpoint` to derive the transaction id and vout, and only
+start the signing process if the JSON result has a `spent` value of false. If
+`spent` is `true` and `status.confirmed` is `false`, a retry should occur
+after an hour until `spent` is `true`, `status.confirmed` is `true`, and
+`status.block_height` is at least 6 blocks past the value returned by issuing
+a GET request to `/blocks/tip/height`. If this occurs, the key material
+should be archived.
```
pdyraga, line 204: There is a scenario possible when one node opts out from recovery and two other report errors because they do not have electrs configured. To address it, we could be logging a warning that electrs is not configured and the client attempts signing without executing the check.

docs/rfc/rfc-2.adoc
```diff
@@ -0,0 +1,799 @@
...
+
+If an optional Bitcoin connection is available (see <<electrs>>), the client
+should look up the 25-block suggested fee per vByte and, if it is available
+and nonzero, should replace the hardcoded max fee with this value. 25 blocks
+means confirmation within ~4 hours. The default of 75 sats per vByte is aimed
+to be slightly higher than a reasonable value for 25 blocks, though of course
+fee market fluctuation can change this at a moment's notice. The 25-block
+suggested fee per vByte should be available at the `/fee-estimates` endpoint
+of the REST API, which returns a JSON object with the key `"25"`
+corresponding to this fee.
```
pdyraga, line 229: Should there be a maximum fee set, for example, 5% of the deposit? Thinking about a scenario where electrs returns very high value and client blindly accepts it. Even if we assume we can trust electrs, having such maximum could make sense for smaller lot sizes where the value returned can be a significant portion of the deposit.

docs/rfc/rfc-2.adoc
```diff
@@ -0,0 +1,799 @@
...
+// corresponding address.
+//
+// [BIP141]: https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki
+func PublicKeyToP2WPKHScriptCode(
+  publicKey *ecdsa.PublicKey,
+  chainParams *chaincfg.Params,
+) ([]byte, error) {
+  // ecdsa.PublicKey and btcec.PublicKey are both method attachments to
+  // Go's crypto/ecdsa.PublicKey, so we can cast.
+	publicKeyBytes := (*btcec.PublicKey)(publicKey).SerializeCompressed()
```
pdyraga, line 349: Been there, done that. 🔨 Still hurts.

pdyraga: It doesn't mean I have any better idea, just expressing my support.

docs/rfc/rfc-2.adoc
Shadowfiend: No, all of this derivation behavior only applies to extended pubkeys.

For extended pubkeys, we optionally (if there is a connection) check that there are no other transactions to the derived address, and derive a new one if there are existing transactions. The transaction count should be nonzero even if the address balance has been transferred out (that's why we check transaction count rather than balance or unspent outputs).

docs/rfc/rfc-2.adoc
Shadowfiend: Not sure I understand this scenario all the way. If one node opts out from recovery altogether (meaning it either has an older version of the client, hasn't configured a return address, or is on a custom build that has omitted the recovery functionality), that will lead to errors when the signing protocol starts, rather than here.

If electrs isn't configured (or, in the current RFC's state, if it is _explicitly disabled_), then we won't try to use it. It could be misconfigured, of course. Either way though, we should log errors if there are errors interacting with electrs.

docs/rfc/rfc-2.adoc
Shadowfiend: My default thought is that this will ultimately prove unnecessary. This is part of why I'm suggesting we look for 25-block confirmation fees. However, liquidations are more likely when BTC is high, and Bitcoin fees are generally higher when BTC is high, so that's not crazy.

Would we treat this as a `min`, basically? `min(first_available(fee_estimate, configured_fee), 5%*utxo value)` or so?

docs/rfc/rfc-2.adoc
Shadowfiend: Yeah it's monstrous.

Shadowfiend: I did steal it from you though :grin:

docs/rfc/rfc-2.adoc
pdyraga: Makes sense, thanks for the explanation!

docs/rfc/rfc-2.adoc
pdyraga: apes together strong

docs/rfc/rfc-2.adoc
pdyraga: Right, that's what I meant. Say we have three clients, two of them do not have electrs configured and hence do not perform the check. One client has electrs configured, it performs the check and discovers BTC has been spent and there is no point in joining the signing process.

In this scenario, we have two clients attempting to sign the transaction and reporting errors that the third client is not responding. 

My suggestion was to print a warning to logs on those two clients before they enter the signing process saying  that electrs is not configured, they can not perform the preliminary check of the deposit, and they enter the process regardless. This could help explain why the process is failing.

docs/rfc/rfc-2.adoc
pdyraga: > Would we treat this as a min, basically? min(first_available(fee_estimate, configured_fee), 5%*utxo value) or so?

Yeah, so it's basically: I am fine with any fee as long as it's no more than 5%*utxo. 

A separate question is, what if it's more than 5%*utxo. They should probably re-attempt the protocol after some time.

Shadowfiend: I kind of hate that, tbh. Adding a different retry in that scenario is going to add a boatload of complexity IMO---both to code and to understanding.

Shadowfiend: Actually if we just adhere to the existing retry strategy that seems sufficient, no? Even though fees don't change every 2 minutes, the protocol will abort early in that scenario, which seems fine. We'll hit the fee endpoint more frequently than might be “ideal”, but it shouldn't be an overall problem, I don't think. Thoughts?

pdyraga: Yeah, I like this approach. 👍

docs/rfc/rfc-2.adoc
Shadowfiend: WARNing that electrs isn't configured makes sense, but I wonder if we should just do it once at startup? Right now we [recommend](https://docs.keep.network/monitoring-and-alerting.html) that more than 10 WARN logs in 15 minutes should trigger an alert. If there's a situation where 10 liquidations trigger around the same time due to price movement, we would trigger an alert, which doesn't feel like the correct thing in this scenario.

docs/rfc/rfc-2.adoc
Shadowfiend: Ok, added this, with one small change: if estimated fee yields >5%*utxo outcome, we fall back on `min(estimated, configured)`. There's a world where the estimated fee is smaller than the configured fee and is still >5%*utxo, and I think in this case it makes sense to continue using the estimated fee.

docs/rfc/rfc-2.adoc
pdyraga: Yeah, emitting them once at startup makes sense. We just need to make sure they clearly articulate what are the implications.

docs/rfc/rfc-2.adoc
```diff
@@ -0,0 +1,811 @@
...
+- F, the desired fee.
+- L, the deposit lot size.
+
+Separately from the amount, the Bitcoin sighash must be computed in order to
+construct a transaction that can spend the tBTC deposit's UTXO; see the tBTC
+spec's section on https://docs.keep.network/tbtc/#sighash[Standardized
+Sighash Construction] for an introduction to this process. Computing the sighash
+requires two additional pieces of data:
+
+- The previous outpoint, which is the concatenation of the funding UTXO's
```
beaushinkle, line 91: Is outpoint a keep term? a quick google didn't turn up anything for me

docs/rfc/rfc-2.adoc
```diff
@@ -0,0 +1,811 @@
...
+  32-byte little-endian transaction hash followed by the 4-byte little-endian
+  index of the output pointing to the funding UTXO.
+- The previous output script, which is to say the address of the deposit.
```
beaushinkle, line 94: In this context, what is a script?

docs/rfc/rfc-2.adoc
```diff
@@ -0,0 +1,811 @@
...
+
+[[recovery-address]]
+===== Specifying an Operator's Bitcoin Recovery Address
+
+To specify a given operator's Bitcoin Recovery Address, this proposal
+suggests adding a single key to the `Extensions.TBTC.BTCRefunds` config
+entry, `BeneficiaryAddress`. The value for this key should be one of:
+
+- A simple BTC address.
+- An xpub, ypub, or zpub that allows the client to derive new addresses for
```
beaushinkle, line 150: Why do we need to support both BTC address and *pubs?

docs/rfc/rfc-2.adoc
Shadowfiend: Outpoint is a Bitcoin term, pulled from the C++ code in bitcoind I believe. You can do a find at https://developer.bitcoin.org/glossary.html mentions it, for example. It's defined as exactly the above, and is the way to reference previous output(s) for funding a transaction.

docs/rfc/rfc-2.adoc
Shadowfiend: There is… A lot of depth here. Bitcoin has a limited stack-based programming language, called Script, that is used to define how a transaction output can be spent. What it actually does depends on the conditions the sender wants to place on that output.

To really dig in on this front, https://github.com/jimmysong/programmingbitcoin is a great resource.

docs/rfc/rfc-2.adoc
Shadowfiend: *pubs are best practice for multiple transactions in the Bitcoin world for privacy purposes, because they allow addresses that are publicly unrelated but privately controlled by the same key, making it less obvious who owns what. Some light explanation of hierarchical deterministic wallets, etc at https://coinguides.org/bitcoin-address-keep-changing/ .

Some folks don't care, and a simple BTC address shortcuts the address derivation process here.

docs/rfc/rfc-2.adoc
beaushinkle: Got it - for the purposes of understanding this RFC, a script in this context is some data/code that lives in the bitcoin blockchain. We can associate it to the deposit, and we can derive the address of the deposit from it?

docs/rfc/rfc-2.adoc
beaushinkle: Okay - eventually the money needs to be sent to an address, and so the signer can either supply the address directly, or the signer can supply an address-generator (a *pub) to add a layer of abstraction, and we want to be able to support both in our first pass. :100:

docs/rfc/rfc-2.adoc
```diff
@@ -0,0 +1,811 @@
...
+purpose. In cases where in-memory/on-disk derivation index and config file index
+both exist, the highest index should be used.
+
+If an optional Bitcoin node connection is available (see <<electrs>>), the
+client should additionally check a derived address for a balance before
+publishing. To do this, a GET request to `/address/<address>` should be issued,
+and the JSON response inspected for a `tx_count` of 0. When such a connection is
+available, if the client finds an address is already used, it should increment
+the derivation index and check again. If the GET request fails, the index should
+be assumed to be valid, as if the Bitcoin node connection were not available.
```
beaushinkle, line 176: checking for understanding:

If the signer wants to use a *pub to generate wallet addresses, the way those generators work is that they have some sort of index (in the same way that a psuedo random number generator has a seed). Since the user would supply no other information besides the generator, we can either:

* generates addresses and then check to see if they've been used before
* not care
* start at 0, assume they've never been used, and maintain internal state

(and other schemes that are more complicated)

We can't just not-care, because if they're bothering to send us a *pub instead of an address, they probably have a large preference toward remaining anonymous, and sending the same wallet multiple transactions over time creates patterns and bleeds information.

We can't check to see if they've been used before without actually connecting to a blockchain, which we're trying to avoid.

We also don't have a great way to manage state other than writing to files so we do it in-memory and read/write to an encrypted file

docs/rfc/rfc-2.adoc
```diff
@@ -0,0 +1,811 @@
...
+
+==== Modification of key material archiving
+
+In the current client, key material is archived as soon as the ECDSA keep
```
beaushinkle, line 180: what is archiving key material?

docs/rfc/rfc-2.adoc
```diff
@@ -0,0 +1,811 @@
...
+issuing a GET request to `/tx/:txid/outspend/:vout`, using the deposit's
+`fundingInfo().utxoOutpoint` to derive the transaction id and vout, and only
+start the signing process if the JSON result has a `spent` value of false. If
+`spent` is `true` and `status.confirmed` is `false`, a retry should occur
+after an hour until `spent` is `true`, `status.confirmed` is `true`, and
+`status.block_height` is at least 6 blocks past the value returned by issuing
+a GET request to `/blocks/tip/height`. If this occurs, the key material
+should be archived.
+
+Note that the output spend may go from being `spent` and unconfirmed to being
```
beaushinkle, line 206: What does the output spend refer to here? The UTXOs from the deposit?

docs/rfc/rfc-2.adoc
Shadowfiend: Yep, you nailed it. And we provide the config option to skip “start at 0”.

docs/rfc/rfc-2.adoc
Shadowfiend: The client keeps key material in two places: the set of `current` keys and the set of `archive` keys. These are represented as directories on the filesystem. At startup, the client loads up the `current` keys, while `archive` keys are ignored. Archiving key material basically means moving keys from `current` to `archive`.

docs/rfc/rfc-2.adoc
Shadowfiend: I should update that to be `outspend`. I'm just referring to the data returned from the API at `/outspend/:index`.

beaushinkle:
On a more meta level, what do you think about putting some description in the filename of the rfc, like `rfc-2-liquidation-recovery.adoc`, that way when someone is trying to pull up the RFC in the future, they're not having to open the actual files and can scan the `/rfc` directory

pdyraga (dismissed):
LGTM, leaving this RFC in @beaushinkle's hands.

docs/rfc/rfc-2.adoc
pdyraga: Adding some code references...

Checking keep status on client start and removing keep from the registry if needed:
https://github.com/keep-network/keep-ecdsa/blob/f240e81002a7b26eb6326be068dbf80b53bc0a18/pkg/client/client.go#L123-L127

On a (confirmed) keep closed event:
https://github.com/keep-network/keep-ecdsa/blob/f240e81002a7b26eb6326be068dbf80b53bc0a18/pkg/client/client.go#L829-L830

On a (confirmed) keep terminated event:
https://github.com/keep-network/keep-ecdsa/blob/f240e81002a7b26eb6326be068dbf80b53bc0a18/pkg/client/client.go#L908-L909

It goes down to the keep registry:
https://github.com/keep-network/keep-ecdsa/blob/f240e81002a7b26eb6326be068dbf80b53bc0a18/pkg/registry/keeps.go#L69-L80

And finally, to the on-disk storage implementation:
https://github.com/keep-network/keep-common/blob/34905d2fe019015fb16dd6edc37da16efa988b87/pkg/persistence/disk_persistence.go#L136-L149

docs/rfc/rfc-2.adoc
Shadowfiend: I would say that we can derive it from the deposit, and we can derive the deposit's associated Bitcoin address from it. The deposit actually has an address on both the Ethereum and Bitcoin chains, and they are not the same or even rooted in the same information.

docs/rfc/rfc-2.adoc
```diff
@@ -0,0 +1,811 @@
...
+}
+
+func doubleSha256(bytes []byte) []byte {
+  hash := sha256.Sum256(bytes)
+  secondHash := sha256.Sum256(hash[:])
+
+  return secondHash[:]
+}
+
+func ConstructSighash(
```
nkuba, line 546: Could we first assemble unsigned TX as per `Construction of Final Transaction`, then get its' `sigHash` with btcd's function (instead of constructing it manually here), sign it, and add the signature to the unsigned TX to get the final one?

Unfortunately, I don't remember where in btcd was the function to pull sigHash for unsigned tx :(

nkuba (commented):
I'm lovin' it! <3 I'm just nitpicking in most of my comments.

docs/rfc/rfc-2.adoc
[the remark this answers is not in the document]
Shadowfiend: :100: just ran into this in [Beau's question above](https://github.com/keep-network/keep-ecdsa/pull/670#discussion_r572426436), will roll this in.

docs/rfc/rfc-2.adoc
```diff
@@ -0,0 +1,811 @@
...
+  from the `getMembers` view method, but also tracked locally on the client
+  alongside the members of the keep.
+- L, the deposit lot size, available on the tBTC deposit contract from the
+  `lotSizeSatoshis` view method.
+- The previous outpoint describing the deposit UTXO, available on the tBTC
+  deposit contract from the `fundingInfo` view method, which returns it as
+  the third `utxoOutpoint` entry in its 3-part return tuple.
+- The keep's public key, available on the Keep contract from the
+  `getPublicKey` view method.
+- The deposit redemption script, straightforwardly computable from the keep's
```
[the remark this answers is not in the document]
Shadowfiend: Yeah a fair point, but the thing is we use “redemption script” to refer specifically to the script used to spend the BTC throughout the system. But let me see if I can do something to make this clearer.

docs/rfc/rfc-2.adoc
```diff
@@ -0,0 +1,811 @@
...
+transaction. The signed transaction can be broadcast via manual submission to a
+known node, via API access to a service provider like BlockCypher or
+Blockstream, or via a manual form submission on a page like
+https://live.blockcypher.com/btc/pushtx/ .
+
+If an optional Bitcoin node connection is available (see <<electrs>>), the
+client should attempt to publish the transaction by performing an HTTP POST
+request to path `/tx`. The POST request body should be the signed transaction
+hex, as would be logged in the WARN logs above. If the POST fails, the client
+should log an error and then fall back on the WARN logging above.
```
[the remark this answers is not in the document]
Shadowfiend: We fall back to the WARN logging so that the signed transaction is available for publishing, rather than to call operator attention. I think it's simpler to consider it “the fallback transaction publishing path”, which _happens_ to be WARN logging. To that end, I think it's simpler to always have that path be the same, irrespective of what else is going on. I may be able to rephrase to make that clearer though.

docs/rfc/rfc-2.adoc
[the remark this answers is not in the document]
nkuba: In this document, we also mentioned `refund transaction` or `liquidation refund transaction`. Maybe this is a way?

docs/rfc/rfc-2.adoc
```diff
@@ -0,0 +1,811 @@
...
+
+=== Implementation Details
+
+There are a few implementation details that are worth calling out; in
+particular, the *pub address derivation, the deposit output script
```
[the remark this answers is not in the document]
Shadowfiend: You flagged this [earlier in the RFC cycle](https://github.com/keep-network/keep-ecdsa/pull/670#discussion_r561866187) :grin: I think it's beyond the scope of this RFC since it's not about the technical implementation, but I think the RFC can explicitly note that docs should be provided for stakers who want to do this.

docs/rfc/rfc-2.adoc
```diff
@@ -0,0 +1,811 @@
...
+
+```go
+// import "github.com/btcsuite/btcd/btcec"
+// import "github.com/btcsuite/btcutil"
+
+// PublicKeyToP2WPKHOutputScript converts a public key to a Bitcion p2wpkh
+// witness scriptCode that can spend an output sent to that public key's
+// corresponding address.
+//
+// [BIP141]: https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki
```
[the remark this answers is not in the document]
Shadowfiend: … Yep, you right. This function went through a few iterations 😅

docs/rfc/rfc-2.adoc
```diff
@@ -0,0 +1,811 @@
...
+    )
+  }
+  if (len(script) > 255) {
+    return fmt.Errorf(
+      "error deriving p2wpkh scriptCode from public key: [scriptCode too long]",
+      len(script)
+    )
+  }
+
+  return append([]byte{byte(len(script))}, script)
```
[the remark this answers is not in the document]
Shadowfiend: That's fair, I can do that.

docs/rfc/rfc-2.adoc
```diff
@@ -0,0 +1,811 @@
...
+that allows spending that UTXO. The signing process does this by including a
+`scriptCode` as part of the the sighash (see <<sighash-construction>>). The
+`scriptCode`, for Witness PubKeyHash transactions, is defined in BIP143.
+
+This function constructs the relevant `scriptCode` bytes from the deposit's
+public key:
+
+```go
+// import "github.com/btcsuite/btcd/btcec"
+// import "github.com/btcsuite/btcutil"
```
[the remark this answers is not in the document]
Shadowfiend: Yeah I actually haven't cross-checked the imports against the function bodies. I may drop the `// import` instead, tbh.

docs/rfc/rfc-2.adoc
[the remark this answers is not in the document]
nkuba: Oh man, I'm sorry! I was going to remove this comment before submitting the review 😆

docs/rfc/rfc-2.adoc
Shadowfiend: I couldn't find one… Which is why I did this. This actually flags something I didn't notice that is kind of confusing though. The current approach [recomputes the fee once the tx is fully assembled](https://github.com/keep-network/keep-ecdsa/pull/670/files/29e6d6d266b23e56ac172bb286ca0b1349bfecc4#diff-e19b702a5d320598532e72f46551a840ee1bda9df2e4415f4daf7295441ae354R689), but that shouldn't work given that the sighash includes the output values. I'm thinking I fumbled something when I was testing these against real examples.

beaushinkle (dismissed):
Lookin' great to me!

docs/rfc/rfc-2.adoc
Shadowfiend: @beaushinkle @pdyraga I do want to fix this before we merge here. I think the changes will be:
- Compute the fee assuming 175 vBytes (as already [noted for the 5% calculation](https://github.com/keep-network/keep-ecdsa/pull/670/files#diff-e19b702a5d320598532e72f46551a840ee1bda9df2e4415f4daf7295441ae354R237)).
- Update the transaction construction to skip the fee calculation and instead take the fee as a parameter.

The other option is to leave transaction construction as-is and follow @nkuba's approach as a pre step to the sighash by constructing the unsigned transaction, then either finding the secret function to pull the sighash or pulling the fee information from the unsigned transaction. Since this is p2wpkh, I believe the witness script length should be the same no matter what, so the unsigned transaction with a placeholder for signature data should be computable once we have the addresses and desired fees for everyone.

docs/rfc/rfc-2.adoc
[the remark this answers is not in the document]
beaushinkle: Check for understanding - the intent of `append([]byte{byte(len(script))}, script)` is to create a new byte array where the head is the length of `script` and the tail is `script`?

If so, i think we need: `append([]byte{byte(len(script))}, script...)`

```
	script := []byte{byte(11), byte(12)}
	b := append([]byte{byte(len(script))}, script...)
	fmt.Println(b)
```
prints out
```
[2 11 12]
```

beaushinkle:
I'm a little confused about how we're collecting the BTC addresses and who is running the command. Is this something that would plug into the ecdsa cli client? If so, does support run the command or would the signer?

The implementation needs to have a btc address for each of the signers, but in the implementation details, we're putting a single btc address in the configuration file. How does each client collect the other addresses?

Shadowfiend:
BTC addresses are provided by each respective client for their own memberships in various deposits, as specified in the section on [Specifying an Operator's Bitcoin Recovery Address](https://github.com/keep-network/keep-ecdsa/pull/670/files#diff-e19b702a5d320598532e72f46551a840ee1bda9df2e4415f4daf7295441ae354R143). The client itself automatically starts the process for a given deposit when its keep transitions to the TERMINATED state, as described in the section on [Modification of key material archiving](https://github.com/keep-network/keep-ecdsa/pull/670/files#diff-e19b702a5d320598532e72f46551a840ee1bda9df2e4415f4daf7295441ae354R183).

docs/rfc/rfc-2.adoc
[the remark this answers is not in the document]
Shadowfiend: Will switch this to “Bitcoin redeem script”, as this is the common way to describe the relevant data in Bitcoin-land.

docs/rfc/rfc-2.adoc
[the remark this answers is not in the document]
Shadowfiend: Rephrased this a bit.

docs/rfc/rfc-2.adoc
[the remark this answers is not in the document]
Shadowfiend: Yep, you're right. The above will cause a compile failure :)

docs/rfc/rfc-2.adoc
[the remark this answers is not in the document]
Shadowfiend: Eh fine, I added that one in. This list will be incomplete still, but I think making sure the btcsuite imports are flagged is worth it.

docs/rfc/rfc-2.adoc
```diff
@@ -0,0 +1,811 @@
...
+  case "xpub", "ypub", "zpub":
+    chainParams = &chaincfg.MainNetParams
+  case "tpub", "upub", "vpub":
+    chainParams = &chaincfg.TestNet3Params
+  }
+
+  requestedAddress, err := requestedPublicKey.Address(chainParams)
+
+  var finalAddress btcutil.Address = requestedAddress
+  switch extendedPublicKey[0:4] {
```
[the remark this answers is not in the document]
Shadowfiend: I love that this is the only reason for that variable yet I forgot to use it :ok_man:

docs/rfc/rfc-2.adoc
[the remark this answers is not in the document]
nkuba: I was too picky with the missing import. I'm glad you decided not to remove the imports completely. It's always super helpful to have the imports in code samples. Better one missing import than no imports at all.

docs/rfc/rfc-2.adoc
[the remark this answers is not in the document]
nkuba: That's why I left this comment. I didn't want the variable to feel unwanted!

docs/rfc/rfc-2.adoc
Shadowfiend: https://pkg.go.dev/github.com/btcsuite/btcd/txscript#CalcWitnessSigHash

You beautiful beautiful man @nkuba . Still need to do some fiddling about to get it done with the data we have, but this is going to feel much cleaner.

nkuba: I knew I've seen it somewhere before (and probably used it too)!

Shadowfiend: This is now handled 🎊

docs/rfc/rfc-2.adoc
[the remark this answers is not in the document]
Shadowfiend: Fix't.

docs/rfc/rfc-2.adoc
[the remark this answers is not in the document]
Shadowfiend: This should be fixed.

docs/rfc/rfc-2.adoc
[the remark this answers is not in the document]
Shadowfiend: Fix't.

docs/rfc/rfc-2.adoc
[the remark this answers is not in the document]
Shadowfiend: Done.

Shadowfiend:
Ok, I believe all comments are handled here. I also reworked the whole process of going from ready message -> sighash -> signed transaction. The process now, as suggested by @nkuba, starts by constructing an unsigned transaction. The unsigned transaction has placeholder dummy signature and pubkeyhash data in order to properly estimate the fee, since that data is not used to generate the sighash. Once the signature is available, the witness script is attached to the existing unsigned transaction and then the whole thing is turned into a hex string.

Everything about this approach is better, especially the part where the fee calculation results in a signature that is actually valid :grimacing:

docs/rfc/rfc-2.adoc
```diff
@@ -689,24 +625,96 @@ func BuildTransactionHexString(
...
+	hashType := uint8(txscript.SigHashAll)
+
+	// The witness signature field is the DER signature followed by the hash type.
+	signatureForWitnessBuffer := &bytes.Buffer{}
+	signatureForWitnessBuffer.Write(signatureDERBytes)
+	binary.Write(signatureForWitnessBuffer, binary.LittleEndian, hashType)
+
+  // The witness is for the first input, since this is known to be a
+  // single-input transaction.
+  signedTransaction.TxIn[0].Witness = [][]byte{
```
nkuba, line 696: Could construct a `wire.TxWitness` object instead of `[][]byte`? For reference please see our old code: 
https://github.com/keep-network/keep-ecdsa/blob/bef6601428605a61d26d98c1e66cda751a049f6c/pkg/btc/witness.go#L15-L32

docs/rfc/rfc-2.adoc
```diff
@@ -0,0 +1,819 @@
...
+transaction is available from above the function in
+<<transaction-construction>>. This sighash is the data then signed by the
+signer group to produce a valid transaction (see the section on
+<<final-transaction-construction>> for the rest of the process once the
+signature has been computed):
+
+```go
+// import "github.com/btcsuite/btcd/txscript"
+
+// unsignedTransaction := ConstructUnsignedTransaction(...)
```
nkuba, line 655: Out of curiosity, why are these line and import lines prepended with `//`?

docs/rfc/rfc-2.adoc
```diff
@@ -689,24 +625,96 @@ func BuildTransactionHexString(
...
+```go
+func BuildSignedTransactionHexString(
+  unsignedTransaction *wire.MsgTx, // <1>
+	signature *ecdsa.Signature, // <2>
+	publicKey *ecdsa.PublicKey, // <3>
+) (string, error) {
+  // For safety's sake, work on a deep copy, as mutations follow.
+  signedTransaction := unsignedTransaction.Copy()
+
+	// Construct witness script.
```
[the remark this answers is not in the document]
Shadowfiend: There are def problems with indentation all over, as some of this was run through the Go formatter and some of it wasn't. I'll see if I can just replace all tabs.

docs/rfc/rfc-2.adoc
Shadowfiend: Going to need to dig into the type juggling going on in there. I guess you can assign structs to slices somehow?

Anyway, yes, this does seem Clearer™.

docs/rfc/rfc-2.adoc
Shadowfiend: For imports, they're not designed to be standalone files. For this line, it's not described here and would be run elsewhere based on various inputs.

docs/rfc/rfc-2.adoc
```diff
@@ -0,0 +1,819 @@
...
+  dummyCompressedPublicKeyForWitness := []byte{
+		0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+		0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+		0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+    0, 0, 0,
+  }
+
+	tx := wire.NewMsgTx(wire.TxVersion)
+	txIn := wire.NewTxIn(
+		wire.NewOutPoint(previousOutputTransactionHash, previousOutputIndex),
```
beaushinkle, line 583: I'm getting an error saying:
```
(type *"github.com/decred/dcrd/chaincfg/chainhash".Hash) as type *"github.com/btcsuite/btcd/chaincfg/chainhash".Hash in argument to wire.NewOutPoint
```

Are we able to typecast  `dcrd/chaincfg/chainhash.Hash` into `btcd/chaincfg/chainhash.Hash`? (it looks like they're both just `[]byte`)?

beaushinkle: I think this is actually because of some import nonsense that I'm working through - false alarm

docs/rfc/rfc-2.adoc
```diff
@@ -0,0 +1,819 @@
...
+	// The witness signature field is the DER signature followed by the hash type.
+	signatureForWitnessBuffer := &bytes.Buffer{}
+	signatureForWitnessBuffer.Write(signatureDERBytes)
+	binary.Write(signatureForWitnessBuffer, binary.LittleEndian, hashType)
+
+  // The witness is for the first input, since this is known to be a
+  // single-input transaction.
+  signedTransaction.TxIn[0].Witness = [][]byte{
+    signatureForWitnessBuffer.Bytes(),
+    compressedPublicKeyBytes((publicKey)),
```
beaushinkle, line 698: what does `compressedPublicKeyBytes` refer to?

docs/rfc/rfc-2.adoc
Shadowfiend: Oh. This is actually the exact same thing. `wire.TxWitness` is [a type alias for `[][]byte`](https://github.com/btcsuite/btcd/blob/v0.21.0-beta/wire/msgtx.go#L241). The rest does effectively the same thing as I was doing already, just with a slightly modified set of function calls. I'll still switch over to it as I think it's more concise, and `wire.TxWitness` makes the intent clearer than `[][]byte`.

Shadowfiend: Done.

docs/rfc/rfc-2.adoc
Shadowfiend: The tweak to witness construction that I did based on [Kuba's note](https://github.com/keep-network/keep-ecdsa/pull/670#discussion_r577527529) substitutes a real implementation here. I think I initially thought this function would need to be written separately, but never wrote it.

Shadowfiend:
Merged master, believe this one is ready to go.

docs/rfc/rfc-2.adoc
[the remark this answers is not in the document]
Shadowfiend: This is done.

nkuba (approved):

pdyraga (approved):

PR #671, opened 2021-01-15
Incorporating the new Ethereum event subscription API and background event pull loop

pdyraga:

Depends on https://github.com/keep-network/keep-ecdsa/pull/663
Depends on https://github.com/keep-network/keep-common/pull/63

# Overview 

There are two major changes to Ethereum subscriptions proposed here, as a result of the changes implemented in https://github.com/keep-network/keep-common/pull/63:
- new subscription API with `OnEvent` and `Pipe` functions,
- background monitoring loop pulling past events from the chain.

The first change should allow implementing some handler logic easier and to avoid complex logic leading to bugs such as keep-network/keep-core#1333 or keep-network/keep-core#2052. ECDSA keep client is currently not that much affected by this chain but this may change when proper event deduplication will be implemented.

The second change should improve client responsiveness for operators running their nodes against Ethereum deployments that are not very reliable on the event delivery front. This should hopefully improve SLA of some mainnet operators of ECDSA client.

# New API

Event subscription API has been refactored in https://github.com/keep-network/keep-common/pull/63 to resemble the proposition from https://github.com/keep-network/keep-core/issues/491. The new event subscription mechanism allows installing event callback handler function with `OnEvent` function as well as piping events from a subscription to a channel with `Pipe` function.

Example usage of `OnEvent`:
```
handlerFn := func(
    submittingMember common.Address,
    conflictingPublicKey []byte,
    blockNumber uint64,
) {
   // (...)
}
subscription := keepContract.ConflictingPublicKeySubmitted(
	nil, // default SubscribeOpts
	nil, // no filtering on submitting member
).OnEvent(handlerFn)
```

The same subscription but with a `Pipe`:
```
sink := make(chan *abi.BondedECDSAKeepConflictingPublicKeySubmitted)

subscription := keepContract.ConflictingPublicKeySubmitted(
	nil, // default SubscribeOpts
	nil, // no filtering on submitting member
).Pipe(sink)
```

Currently, all our event subscriptions in ECDSA client use function handlers and the code has been adjusted to the new API but is still using function handlers. This may or may not change in the future depending on an individual use case.

# Background monitoring loop

Some nodes in the network are running against Ethereum setups that are not particularly reliable in delivering events. Events are not delivered, nodes are not starting key-generation, or are not participating in redemption signing. Another problem is the stability of the event subscription mechanism (see https://github.com/keep-network/keep-ecdsa/pull/663). If the web socket connection is dropped too often, the resubscription mechanism is not enough to receive events emitted when the connection was in a weird, stale state.

To address this problem, https://github.com/keep-network/keep-common/pull/63 introduces a background loop periodically pulling past events from the chain next to a regular `watchLogs` subscription. How often events are pulled and how many blocks are taken into account can be configured with `SubscribeOpts` parameters. 

This way, even if the event was lost by `watchLogs` subscription for whatever reason, it should be pulled by a background monitoring loop later. An extremely important implication of this change is that handlers should have logic in place allowing them to de-duplicate received events even if a lot of time passed between receiving the original event and the duplicate. 

This part has been implemented in `event_deduplicator.go` for four events:
- opened keep (key generation requested),
- redemption signature requested,
- keep closed,
- keep archived.

tBTC-specific events are covered separately in https://github.com/keep-network/keep-ecdsa/pull/679.

The only event that is not covered with deduplication is conflicting public key submitted event but it does not look to me as something that needs to be deduplicated as the subscription is canceled immediately after this event is received.


pkg/client/event_deduplicator.go
```diff
@@ -0,0 +1,188 @@
...
+// time. It is eventDeduplicator's responsibility to decide whether the given
+// event is a duplicate and should be ignored or if it is not a duplicate and
+// should be handled.
+//
+// Four events are supported:
+// - key generation request for a new keep,
+// - signature request for a keep,
+// - keep close request,
+// - keep terminate request.
+type eventDeduplicator struct {
```
lukasz-zimnoch, line 26: Thoughts on extracting `event_deduplicator.go` and `event_tracking.go` to a separate package `event`? This way we could remove the `event` prefix from file and struct names and achieve better encapsulation.

pkg/client/event_tracking.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+// terminateKeepTrack is used to track keep terminate events. It is used to
+// ensure that the process execution is not duplicated, e.g. when the client
+// receives the same event multiple times. When event is received, it should be
+// noted in this struct. When the signer generation process completes (no matter
+// if it succeeded or failed), it should be removed from this struct.
+type terminateKeepTrack struct {
+	*keepEventTrack
+}
```
lukasz-zimnoch, line 36: Why do we need those specialized structs? Wouldn't be enough to have just three instances of `keepEventTrack` in the deduplicator?

pkg/client/event_tracking.go
```diff
@@ -0,0 +1,138 @@
...
+// - keep getting closed
+// - keep getting archived
+//
+// Those events do not require any special tracking or additional fields to
+// distinguish them. It is used to ensure that the process execution is not
+// duplicated, e.g. when the client receives the same event multiple times. When
+// event is received, it should be noted in this struct. When the signer
+// generation process completes (no matter if it succeeded or failed), it should
+// be removed from this struct.
+type keepEventTrack struct {
```
lukasz-zimnoch, line 50: I'm wondering whether we could just use `string` as type of `keepAddress` param in all methods. This could be a big win as we could get rid of the `go-ethereum` dependency. May be useful in the context of future multi-chain support.

pkg/client/event_deduplicator.go
lukasz-zimnoch, line 26: If we decide to use `string` instead of `common.Address` in the event tracker, we should also do it here. I see we will have a problem with the `keepRegistry` interface but we can replace it by a type `type signerExistsFn func(keepAddress string) bool` or something similiar and let the caller play with address conversion.

pkg/client/event_tracking.go
```diff
@@ -0,0 +1,138 @@
...
+	data  map[string]bool // <keep, bool>
+	mutex *sync.Mutex
```
lukasz-zimnoch, line 52: Why do we use a mutex pointer here?

pkg/client/event_tracking.go
```diff
@@ -0,0 +1,138 @@
...
+
+// requestedSignaturesTrack is used to track signature calculation started after
+// signature request event is received. It is used to ensure that the process execution
+// is not duplicated, e.g. when the client receives the same event multiple times.
+// When event is received, it should be noted in this struct. When the signature
+// calculation process completes (no matter if it succeeded or failed), it should
+// be removed from this struct.
+type requestedSignaturesTrack struct {
+	data  map[string]map[string]bool // <keep, <digest, bool>>
+	mutex *sync.Mutex
```
lukasz-zimnoch, line 90: Why do we use a mutex pointer here?

pkg/client/event_deduplicator.go
lukasz-zimnoch, line 26: Wouldn't `eventApprover` (or something similar) reflect better what are we doing here? The real deduplication occurs in `keepEventTrack` instances. Here we are one level upper as we're also checking things against the chain.

pdyraga: That's not a bad idea but I'd do it as `pkg/client/event` and not as `pkg/event`. This code is about how the client reacts to those events and not about a generic concept of an event.

pdyraga: Please see 7f6dd64fe65abe7f2f343dce1d873ba585cf22c3.

pkg/client/event_tracking.go
pdyraga: Good point. We do not need them, it is definitely too much. Removed in c7e3ce66e74ec2c4411d8d5189e66879dc4f2372.

pkg/client/event_tracking.go
pdyraga: I just copied this code without changing this piece but... let's take this as an opportunity and fix it.

pdyraga: Done in 7588bc22432c9a969c5a37a81a51a305dbf062eb.

pkg/client/event_tracking.go
pdyraga: Explained [in the previous comment](https://github.com/keep-network/keep-ecdsa/pull/671/files/df9623df0bba07ebf4119a55cfaaefb3dc455f38#diff-eed7bc28eaadcd6dba03386d92e97ad8a4a42c524220bc645fc9135d4c37508c).

pkg/client/event_deduplicator.go
pdyraga: I'd say that real deduplication happens in `Deduplicator` (former `eventDeduplicator`). `keepEventTrack` is filtering out *immediate* duplicates but duplicates that will be pulled by subscription's past events call, will be deduplicated in `Deduplicator`.

pkg/utils/wrappers.go
```diff
@@ -0,0 +1,126 @@
+package utils
```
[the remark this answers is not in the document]
pdyraga: Yeah, was thinking about it but I'd leave it for the future when we'll be wanting to reuse this code from `keep-core`. It may happen that we do something in the meantime that will invalidate this idea.

pkg/client/client.go
```diff
@@ -543,44 +544,74 @@ func monitorSigningRequests(
...
-						keepAddress.String(),
-						event.Digest,
-					)
-					return
-				}
-
-				generateSignatureForKeep(clientConfig, tssNode, keepAddress, signer, event.Digest)
+				utils.DoWithDefaultRetry(
+					clientConfig.GetSigningTimeout(),
+					func(ctx context.Context) error {
```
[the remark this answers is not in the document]
pdyraga: Agreed it is needed but I am concerned about touching this code in this PR as the diff is already pretty significant. I would rather address it when working on support for multiple chains. We'll have to clean up the relationship between `client.go` and `node.go` there anyway. WDYT?

pkg/client/event_deduplicator.go
lukasz-zimnoch: Ok, that makes sense now. Thanks for the clarification.

pkg/utils/wrappers.go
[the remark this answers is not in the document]
lukasz-zimnoch: Sure. Let's leave this for now then.

pkg/client/client.go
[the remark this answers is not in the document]
lukasz-zimnoch: Sure, let's just leave some TODOs here.

pkg/utils/wrappers.go
```diff
@@ -0,0 +1,126 @@
...
+			backoffTime *= 2
+			if backoffTime > backoffMax {
+				backoffTime = backoffMax
+			}
+		}
+	}
+}
+
+func backoffWait(ctx context.Context, waitTime time.Duration) bool {
+	timer := time.NewTimer(waitTime)
```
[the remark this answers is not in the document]
pdyraga: Very good point. I assumed this is not important because handlers should not be invoked at the same time but given that this code is generic and can be used anywhere in the future, jitter absolutely makes sense. Implemented it in 3e7c210004ed9cadf6082a2bdea79b05ce5a761c.

pkg/client/event_deduplicator.go
pdyraga: Yesh... I was struggling with it, took another attempt and... I am not sure if it's the right moment to address it. 

`keepRegistry` is not the only problem. The entire `chain.Handle` interface is organized around `common.Address`. So it is `SubmitSignature`, `SubmitKeepPublicKey`, also `RequestSignature`, `AddSigner` and `OpenKeep` from `local` chain implementation.

As much as I do agree we should not let the `common.Address` propagate any further and try to write a new code in a chain-agnostic way I do wonder if this situation is not an exception. We'll end up with a bunch of transformations back and forward.

Instead of doing it now, I'd suggest addressing `common.Address` problem separately in another PR while preparing the client for proper multi-chain support where we'll update all the layers. What do you think?

pdyraga: Actually, I think we can try to address it partially. Leave `keepRegistry` with `common.Address` and `common.HexToAddress` before calling it. The rest of the code are tests, so there might be some acceptable level of boilerplate code to achieve our goal. Let me try, I am not satisfied with what's here currently.

lukasz-zimnoch: Sure that's a reasonable proposition.

pdyraga: Here is the proposition, I intentionally pushed it to a separate branch in case we decide not to go with it here: https://github.com/keep-network/keep-ecdsa/commit/bdcfb776b35216ffb6cfc78516916657bdd2c399

I am not sure... there is a lot of going back and forward with transformation. I feel that the first comment I made in this thread makes sense and maybe we should wait with this change and swallow `common.Address` for now. There are too many dependencies with other components in the deduplicator.

pkg/client/client.go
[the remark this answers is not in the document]
pdyraga: Added it in ee39b6c82272cca7259ce396d2aa32e9e5945fa9.

pkg/client/event_tracking.go
pdyraga: Discussion about this [here](https://github.com/keep-network/keep-ecdsa/pull/671#discussion_r562597150).

pkg/client/event_deduplicator.go
lukasz-zimnoch: Ok, after looking at the proposition I agree we should rather leave `common.Address` for now. It's safer and more readable for now. Trying to remove that dependency doesn't give us any benefits for now. We will have to revisit that part while implementing chain-agnostic approach anyway.

lukasz-zimnoch:
No more comments! Code looks good!

lukasz-zimnoch (approved):

pkg/client/client.go
```diff
@@ -180,35 +179,31 @@ func Initialize(
...
 			event.Members,
+			event.BlockNumber,
 		)
 
 		if event.IsMember(ethereumChain.Address()) {
 			go func(event *eth.BondedECDSAKeepCreatedEvent) {
-				if ok := requestedSigners.add(event.KeepAddress); !ok {
-					logger.Warningf(
-						"keep creation event for keep [%s] already registered",
+				if ok := eventDeduplicator.NotifyKeyGenStarted(event.KeepAddress); !ok {
```
nkuba, line 196: I prefer the `shouldHandle` name we used for the signing notification, it describes what's happening well.

```suggestion
				if shouldHandle := eventDeduplicator.NotifyKeyGenStarted(event.KeepAddress); !shouldHandle {
```

pkg/client/client.go
```diff
@@ -728,30 +776,44 @@ func monitorKeepClosedEvents(
...
-					return ethereumChain.IsActive(keepAddress)
-				},
-			)
-			if err != nil {
-				logger.Errorf(
-					"failed to confirm keep [%s] closure: [%v]",
-					keepAddress.String(),
-					err,
+			go func(event *eth.KeepClosedEvent) {
+				if ok := eventDeduplicator.NotifyClosingStarted(keepAddress); !ok {
```
nkuba, line 780:
```suggestion
				if shouldHandle := eventDeduplicator.NotifyClosingStarted(keepAddress); !shouldHandle {
```

pkg/client/client.go
```diff
@@ -792,30 +855,44 @@ func monitorKeepTerminatedEvent(
...
-					return ethereumChain.IsActive(keepAddress)
-				},
-			)
-			if err != nil {
-				logger.Errorf(
-					"failed to confirm keep [%s] termination: [%v]",
-					keepAddress.String(),
-					err,
+			go func(event *eth.KeepTerminatedEvent) {
+				if ok := eventDeduplicator.NotifyTerminatingStarted(keepAddress); !ok {
```
nkuba, line 859:
```suggestion
				if shouldHandle := eventDeduplicator.NotifyTerminatingStarted(keepAddress); !shouldHandle {
```

pkg/chain/ethereum/ethereum.go
```diff
@@ -106,11 +107,14 @@ func (ec *EthereumChain) OnKeepClosed(
...
-			handler(&eth.KeepClosedEvent{BlockNumber: blockNumber})
-		},
-	), nil
+
+	onEvent := func(blockNumber uint64) {
+		handler(&eth.KeepClosedEvent{BlockNumber: blockNumber})
+	}
+	return keepContract.KeepClosed(&ethutil.SubscribeOpts{
+		Tick:       4 * time.Hour,
+		PastBlocks: 2000,
```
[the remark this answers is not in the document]
pdyraga: I have mixed feelings about it and that's why I decided on a safer option. Those two parameters are quite fragile, when someone is altering them, they really need to know what they are doing. Past blocks value depends on tick, tick depends on timeouts and the expected time of executing the protocol. Timeouts and time of executing the protocol are not configurable. I suggest we leave it as-is for now and perhaps make it configurable in future versions. It's always easier to make something configurable than to remove this possibility.

pkg/client/event/tracking.go
```diff
@@ -7,35 +7,48 @@ import (
...
+// - keep getting archived
+//
+// Those events do not require any special tracking or additional fields to
+// distinguish them. It is used to ensure that the process execution is not
+// duplicated, e.g. when the client receives the same event multiple times. When
+// event is received, it should be noted in this struct. When the signer
 // generation process completes (no matter if it succeeded or failed), it should
 // be removed from this struct.
-type requestedSignersTrack struct {
+type keepEventTrack struct {
```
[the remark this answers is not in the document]
pdyraga: `uniqueEventTrack` makes sense, I'll rename it.

pkg/utils/wrappers.go
```diff
@@ -0,0 +1,149 @@
...
+
+const (
+	// DefaultDoBackoffTime is the default value of backoff time used by
+	// DoWithDefaultRetry function.
+	DefaultDoBackoffTime = 1 * time.Second
+
+	// DefaultDoMaxBackoffTime is the default value of max backoff time used by
+	// DoWithDefaultRetry function.
+	DefaultDoMaxBackoffTime = 120 * time.Second
+)
```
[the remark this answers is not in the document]
pdyraga: That's what I started with but... they are used in `DoWithDefaultRetry` so the context is better.

pkg/client/event/deduplicator.go
```diff
@@ -0,0 +1,193 @@
...
+	keepAddress common.Address,
+	digest [32]byte,
+) (bool, error) {
+	if d.requestedSignatures.has(keepAddress, digest) {
+		return false, nil
+	}
+
+	// repeat the check in case of a small chain reorg or if chain nodes
+	// are out of sync
+	isAwaitingSignature, err := utils.ConfirmWithTimeout(
```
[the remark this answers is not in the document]
pdyraga: This is something different. This check covers the situation when we received an event, we are not handling this event at the moment (not an immediate duplicate) and we need to validate whether this is something recent or some event from the past. For this reason, we execute on-chain call and we retry this call only to address situations when Ethereum nodes behind a load balancer are not yet in sync. If we can not confirm the signing is pending, we end up with `shouldHandle = false` and we consider it a normal situation.

The referenced check in `client.go` addresses the situation of chain reorg when the event was emitted, signing request was confirmed on-chain but it disappeared from there as a result of a chain reorg and we do not want to proceed with signing because adversary would be able to report fraud.

I do agree these two checks could probably be combined assuming we do not want to log warning for the second one but I am not sure if it would be the right call. The context is different, basically.

pkg/utils/wrappers.go
```diff
@@ -0,0 +1,149 @@
...
+
+			timedOut := backoffWait(ctx, backoffTime)
+			if timedOut {
+				return fmt.Errorf("retry timeout [%v] exceeded", timeout)
+			}
+
+			backoffTime = calculateBackoff(
+				backoffTime,
+				backoffMax,
+			)
```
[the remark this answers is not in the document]
pdyraga: I was thinking about it but the current solution seems a bit cleaner. We do not have to return two values from `calculateBackoff` and we only calculate `backoffWait` if there is a chance we'll use it, that is, no timeout on the context.

pkg/client/event/deduplicator.go
```diff
@@ -0,0 +1,193 @@
...
+	chain        chain.Handle
+
+	keyGenKeeps         *keepEventTrack
+	requestedSignatures *requestedSignaturesTrack
+	closingKeeps        *keepEventTrack
+	terminatingKeeps    *keepEventTrack
+}
+
+type keepRegistry interface {
+	HasSigner(keepAddress common.Address) bool
```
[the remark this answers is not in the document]
pdyraga: See f57fe591a1f14ff95bd22241e5934f48f83c8d0a please.

pkg/client/event/tracking.go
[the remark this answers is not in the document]
pdyraga: See ec75b4362401041bfc083a5d2245b0acf39f5b5a please.

pkg/client/event/deduplicator.go
```diff
@@ -0,0 +1,193 @@
...
+//
+// In case the client proceeds with the key generation, it should call
+// notifyKeyGenCompleted once the protocol completes, no matter if it failed or
+// succeeded.
+func (d *Deduplicator) NotifyKeyGenStarted(keepAddress common.Address) bool {
+	if d.keyGenKeeps.has(keepAddress) {
+		return false
+	}
+
+	if d.keepRegistry.HasSigner(keepAddress) {
```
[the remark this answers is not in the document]
pdyraga: See f57fe591a1f14ff95bd22241e5934f48f83c8d0a please.

pkg/client/event/deduplicator.go
```diff
@@ -0,0 +1,193 @@
...
+	d.keyGenKeeps.remove(keepAddress)
+}
+
+// NotifySigningStarted notifies the client wants to start signature generation
+// for the given keep and digest upon receiving an event. It returns boolean
+// indicating whether the client should proceed with the execution or ignore the
+// event as a duplicate.
+//
+// In case the client proceeds with signing, it should call
+// notifySigningCompleted once the protocol completes, no matter if it failed or
```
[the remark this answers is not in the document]
pdyraga: Fixed in 030bad40773ca3bbe70fb439c13005c2144f2fcf.

pkg/client/event/deduplicator.go
```diff
@@ -0,0 +1,193 @@
...
+		terminatingKeeps:    terminatingKeeps,
+	}
+}
+
+// NotifyKeyGenStarted notifies the client wants to start key generation for
+// a keep upon receiving an event. It returns boolean indicating whether the
+// client should proceed with the execution or ignore the event as a duplicate.
+//
+// In case the client proceeds with the key generation, it should call
+// notifyKeyGenCompleted once the protocol completes, no matter if it failed or
```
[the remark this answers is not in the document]
pdyraga: Fixed in 030bad40773ca3bbe70fb439c13005c2144f2fcf.

pkg/client/event/deduplicator.go
```diff
@@ -0,0 +1,193 @@
...
+) {
+	d.requestedSignatures.remove(keepAddress, digest)
+}
+
+// NotifyClosingStarted notifies the client wants to close a keep upon receiving
+// an event. It returns boolean indicating whether the client should proceed
+// with the execution or ignore the event as a duplicate.
+//
+// In case the client proceeds with closing the keep, it should call
+// notifyClosingCompleted once the protocol completes, no matter if it failed or
```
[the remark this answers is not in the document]
pdyraga: Fixed in 030bad40773ca3bbe70fb439c13005c2144f2fcf.

pkg/client/event/deduplicator.go
```diff
@@ -0,0 +1,193 @@
...
+func (d *Deduplicator) NotifyClosingCompleted(keepAddress common.Address) {
+	d.closingKeeps.remove(keepAddress)
+}
+
+// NotifyTerminatingStarted notifies the client wants to terminate a keep upon
+// receiving an event. It returns boolean indicating whether the client should
+// proceed with the execution or ignore the event as a duplicate.
+//
+// In case the client proceeds with terminating the keep, it should call
+// notifyTerminatingCompleted once the protocol completes, no matter if it
```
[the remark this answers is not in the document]
pdyraga: Fixed in 030bad40773ca3bbe70fb439c13005c2144f2fcf.

pkg/client/client.go
pdyraga: Please see e194a9c36f2b28fc5b78968f16cc32fffeeed953.

pkg/client/client.go
pdyraga: Please see e194a9c36f2b28fc5b78968f16cc32fffeeed953.

pkg/client/client.go
pdyraga: Please see e194a9c36f2b28fc5b78968f16cc32fffeeed953.

pkg/utils/wrappers.go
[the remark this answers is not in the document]
nkuba: Up to you.

nkuba (approved):

PR #674, opened 2021-01-18
Staker rewards: Decrease the required keygen SLA to 80%

lukasz-zimnoch:
So far, an operator had to have no more than 10% of failed key generations for >=10 of total key generations count. That requirement turned out to be a little too restrictive for big operators. Because of that, we are increasing that threshold to 20%.


pdyraga:
Tested and looks good. @lukasz-zimnoch please update `staker-rewards/README.md` as it is currently out of date.

pdyraga:
Additional reasoning behind this change:

Some operators in the network complained about SLA for key-generation being too aggressive. If you think about key-generation as a kind of a filter not allowing poorly operated nodes to enter into the deposit, it is obvious that we will have more key-generation failures than redemption signature failures and it may happen that a poorly operated node affects SLA of another, properly operating, node. For this reason, we are making SLA for key-generation more operator-friendly. 

The previous version was:
> For less than 10 key generations, one of them can fail. For 10 or more key generations, no more than 10% can fail.

What translated to:
- [0-9] key generations - one could fail
- [10-19] key generations - one could fail
- [20-29] key generations - two could fail
- [30-39] key generations - three could fail

The new version is:
> For less than 10 key generations, one of them can fail. For 10 or more key generations, no more than 20% can fail.

What translates to:

- [0-9] key generations - one can fail
- [10-14] key generations - two can fail
- [15-19] key generations - three can fail
- [20-24] key generations - four can fail
- [24-29] key generations - five can fail
- [30-34] key generations - six can fail
- [35-39] key generations - seven can fail

pdyraga (approved):

PR #677, opened 2021-01-21
Bump up sortition pools to the latest

nkuba:
Bumped up sortition-pools version to the latest pre release to fix
contracts deployment.


r-czajkowski (approved):
Tested contracts deployment locally and works! 👍

PR #679, opened 2021-01-26
tBTC chain extension: Incporporating the new Ethereum event subscription API and background event pull loop

pdyraga:

Depends on https://github.com/keep-network/keep-ecdsa/pull/671
Depends on https://github.com/keep-network/tbtc/pull/750

# Overview 

There are two major changes to Ethereum subscriptions proposed here, as a result of the changes implemented in https://github.com/keep-network/keep-common/pull/63:
- new subscription API with `OnEvent` and `Pipe` functions,
- background monitoring loop pulling past events from the chain.

The first change should allow implementing some handler logic easier and to avoid complex logic leading to bugs such as keep-network/keep-core#1333 or keep-network/keep-core#2052. ECDSA keep client is currently not that much affected by this chain but this may change when proper event deduplication will be implemented.

The second change should improve client responsiveness for operators running their nodes against Ethereum deployments that are not very reliable on the event delivery front. This should hopefully improve SLA of some mainnet operators of ECDSA client.

# New API

Event subscription API has been refactored in https://github.com/keep-network/keep-common/pull/63 to resemble the proposition from https://github.com/keep-network/keep-core/issues/491. The new event subscription mechanism allows installing event callback handler function with `OnEvent` function as well as piping events from a subscription to a channel with `Pipe` function.

Example usage of `OnEvent`:
```
handlerFn := func(
    submittingMember common.Address,
    conflictingPublicKey []byte,
    blockNumber uint64,
) {
   // (...)
}
subscription := keepContract.ConflictingPublicKeySubmitted(
	nil, // default SubscribeOpts
	nil, // no filtering on submitting member
).OnEvent(handlerFn)
```

The same subscription but with a `Pipe`:
```
sink := make(chan *abi.BondedECDSAKeepConflictingPublicKeySubmitted)

subscription := keepContract.ConflictingPublicKeySubmitted(
	nil, // default SubscribeOpts
	nil, // no filtering on submitting member
).Pipe(sink)
```

Currently, all our event subscriptions in ECDSA client use function handlers and the code has been adjusted to the new API but is still using function handlers. This may or may not change in the future depending on an individual use case.

# Background monitoring loop

Some nodes in the network are running against Ethereum setups that are not particularly reliable in delivering events. Events are not delivered, nodes are not starting key-generation, or are not participating in redemption signing. Another problem is the stability of the event subscription mechanism (see https://github.com/keep-network/keep-ecdsa/pull/663). If the web socket connection is dropped too often, the resubscription mechanism is not enough to receive events emitted when the connection was in a weird, stale state.

To address this problem, https://github.com/keep-network/keep-common/pull/63 introduces a background loop periodically pulling past events from the chain next to a regular `watchLogs` subscription. How often events are pulled and how many blocks are taken into account can be configured with `SubscribeOpts` parameters. 

This way, even if the event was lost by `watchLogs` subscription for whatever reason, it should be pulled by a background monitoring loop later. An extremely important implication of this change is that handlers should have logic in place allowing them to de-duplicate received events even if a lot of time passed between receiving the original event and the duplicate. 

The original version of tBTC deposit monitoring code - without the modification here - was mostly safe. There is 24 hours monitoring period eliminating duplicate events within this time frame.

Just to be absolutely safe, in case 24 hours time period is exceeded for a duplicate, I added an initial state check in this PR (see ff611ec). Before the monitoring starts on start function signal, we check the initial deposit state whether
it's an expected one. If not, we do not start the monitoring.


pkg/extensions/tbtc/tbtc.go
```diff
@@ -780,7 +813,39 @@ func (t *tbtc) shouldMonitorDeposit(depositAddress string) bool {
...
 			err,
 		)
-		return false // return false but don't cache the result in case of error
+		// return false but don't cache the result in case of an error;
+		// it gives a chance to retry the monitoring later but sooner
+		// than notMonitoredDepositsCache caching period.
+		return false
+	}
+
+	hasInitialState, err := utils.ConfirmWithTimeoutDefaultBackoff(
```
lukasz-zimnoch, line 822: Why not just increase the `monitoringCachePeriod`?

pdyraga: Because this value would depend on subscription settings - specifically the values in `SubscribeOpts` - and blocks mining time in the particular chain/network. The current value of 24 hours is probably enough for Ethereum mainnet and default `SubscribeOpts` fwiw.

pkg/client/client.go
```diff
@@ -654,7 +658,11 @@ func checkAwaitingSignature(
 		utils.DoWithDefaultRetry(
 			clientConfig.GetSigningTimeout(),
 			func(ctx context.Context) error {
-				shouldHandle, err := eventDeduplicator.notifySigningStarted(keepAddress, latestDigest)
+				shouldHandle, err := eventDeduplicator.notifySigningStarted(
+					60*time.Second,
```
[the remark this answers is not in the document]
pdyraga: Please see https://github.com/keep-network/keep-ecdsa/pull/679/commits/6fb4c2996c20195ca373e0b8f86f3fed5593fe95.

pkg/client/client.go
```diff
@@ -547,7 +547,11 @@ func monitorSigningRequests(
 				utils.DoWithDefaultRetry(
 					clientConfig.GetSigningTimeout(),
 					func(ctx context.Context) error {
-						shouldHandle, err := eventDeduplicator.notifySigningStarted(keepAddress, event.Digest)
+						shouldHandle, err := eventDeduplicator.notifySigningStarted(
+							60*time.Second,
```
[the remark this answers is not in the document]
pdyraga: Please see https://github.com/keep-network/keep-ecdsa/pull/679/commits/6fb4c2996c20195ca373e0b8f86f3fed5593fe95.

pkg/utils/wrappers.go
```diff
@@ -113,6 +113,38 @@ func ConfirmWithTimeout(
...
+	// ConfirmWithDefaultTimeout function.
+	DefaultConfirmBackoffTime = 5 * time.Second
+
+	// DefaultConfirmMaxBackoffTime is the default value of max backoff time
+	// used by ConfirmWithDefaultTimeout function.
+	DefaultConfirmMaxBackoffTime = 10 * time.Second
+)
+
+// ConfirmWithTimeoutDefaultBackoff executed the provided confirmFn until it
+//  returnstrue or until it fails or until timeout is hit. It applies
```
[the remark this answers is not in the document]
pdyraga: Done in https://github.com/keep-network/keep-ecdsa/pull/679/commits/13b0501d32747bf46e855d1943b5ed5462d5d8bc.

pkg/client/event_deduplicator.go
```diff
@@ -111,6 +111,7 @@ func (ed *eventDeduplicator) notifyKeyGenCompleted(keepAddress common.Address) {
 // notifySigningCompleted once the protocol completes, no matter if it failed or
 // succeeded.
 func (ed *eventDeduplicator) notifySigningStarted(
+	timeout time.Duration,
```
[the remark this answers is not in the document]
pdyraga: See https://github.com/keep-network/keep-ecdsa/pull/679/commits/a5a289a30a6f127d0d3875236eb167a01c04db9e please.

pkg/extensions/tbtc/tbtc.go
```diff
@@ -839,7 +813,39 @@ func (t *tbtc) shouldMonitorDeposit(depositAddress string) bool {
...
```
lukasz-zimnoch, line 822: I performed several tests and I think we have a problem here. Suppose a test scenario:

1. A deposit is opened and clients start `retrieve pubkey` monitoring. All of them cache this deposit in their `monitoredDepositsCache`.
2. The public key is retrieved by a third-party. All clients stop their `retrieve pubkey` monitoring instances.
3. Before `24 hours` pass, clients lose their ETH connections for a while. After restoring, past events are checked. This causes an occurrence of a duplicated `Created` event telling about the deposit creation once again.
4. Clients try to start a new instance of the `retrieve pubkey` monitoring for the same deposit. Because they all have that deposit in their `monitoredDepositsCache`, they all decide to start a new monitoring instance without checking the actual deposit state.

pdyraga: Good catch, addressed in 77c355b. I will quote here explanation from the commit message:


The previous approach assumed that if the given deposit address is cached in `monitoredDepositsCache`, `shouldMonitorDeposit` returns `true` without executing any further checks.

Together with `acquireMonitoringLock` and `releaseMonitoringLock`, it worked fine for most of the cases but not for all of them. Specifically, for the case when the deposit has already transitioned to the next state, we could run into a situation when the monitoring lock has already been released and `shouldMonitorDeposit` returns `true` just because the client has a member in that deposit, without executing the initial state check.

The order of cache checks has been changed to address it.

Also, renamed `monitoredDepositsCache` to `memberDepositsCache`, and `notMonitoredDepositsCache` to `notMemberDepositsCache` to state it clearly those caches hold information whether the client has or has not a member in the given deposit.

The new approach is:
1. Check if the given deposit is in `notMonitoredDepositsCache`. If so, we are sure the client is not interested in monitoring this deposit no matter what - it has no member in it.

2. Check the initial state of the deposit to do not start monitoring for duplicate events from the past. If the initial state is not as expected, the client is not interested in monitoring this deposit as the event triggering the monitoring is something from the past that does not match the current chain state.

3. Check if the given deposit is in `memberDepositCache`. If so, start the monitoring. This check allows to do not repeat `getSignerIndex` call for the same deposit more than one time. If the deposit is not in `memberDepositCache`, execute `getSignerIndex` call, check members, update caches and return `true` or `false` depending on if the client has a member in the deposit.

lukasz-zimnoch: Ok, that change looks good. Tested and the problem has gone.

lukasz-zimnoch:
LGTM! I've tested and haven't found any problems. Specifically, I've checked:

- starting all the monitorings (`retrieve pubkey`, `provide redemption signature`, `provide redemption proof`) on appropriate events
- stopping the `retrieve pubkey` monitoring when the key has been retrieved by a third-party
- stopping the `provide redemption signature` monitoring when the signature has been provided by a third-party 
- stopping the `provide redemption proof` monitoring when the proof has been provided by a third-party
- retrieving the public key on `retrieve pubkey` timeout
- submitting redemption signature on `provide redemption signature` timeout
- increasing redemption fee on `provide redemption proof` timeout
- resubscription and monitoring behavior after Ethereum node crash
- reaction on event duplicates
- usual keygen and signing

lukasz-zimnoch (approved):

PR #681, opened 2021-01-27
Keep ECDSA on Celo - development environment setup scripts

dimpar:
This PR refactors existing scripts for Ethereum connection and also adds 2 new installation scripts for Celo chain: local & alfajores. 

**Ethereum**
_Local_
Ethereum connection should work as it worked before and the invocation with the default params should remain the same. 

- `./scripts/install.sh`
- `./scripts/initialize.sh`

For available options add `--help` flag.

**Celo**
_Local_
Local connection and the default values are the same as for Ethereum local connection. For Celo installation use:
- `./scripts/install-celo.sh`
- `./scripts/initialize-celo.sh`

For available options add `--help` flag.

_Alfajores_

```
Celo Alfajores host connection info.
  URL    = "wss://alfajores-forno.celo-testnet.org/ws"
  URLRPC = "https://alfajores-forno.celo-testnet.org"
```

Example usage:
```
CONTRACT_OWNER_CELO_ACCOUNT_PRIVATE_KEY=f84e15953c7478f0853d03d4d580d2317f42c55fda572ff986fd70b3c4e3eae1 ./scripts/install-celo.sh --keep-core-path /Users/dp/go/src/github.com/keep-network/keep-core-celo/keep-core --network alfajores
```

```
CONTRACT_OWNER_CELO_ACCOUNT_PRIVATE_KEY=f84e15953c7478f0853d03d4d580d2317f42c55fda572ff986fd70b3c4e3eae1 ./scripts/initialize-celo.sh --network alfajores --application-address 521d24b3c9d6280454c456e832d9b720b2c685e8
```


solidity/scripts/lcl-initialize.js
```diff
@@ -16,10 +16,32 @@ module.exports = async function () {
...
+      iter = 1
+    } else if (accounts.length > 5) {
+      numberOfAccounts = 5
+      iter = 1
+    }
+
+    const operators = []
+    for (let i = iter; i < numberOfAccounts; i++) {
+      operators.push(accounts[i])
+    }
```
pdyraga, line 43: I do not understand this change. The assumption in `solidity/integration/smoke_test.js` is that accounts [1,4] are operators, account 5 is a keep owner, and account 6 is the application. This will work for the second case where we have at least 6 accounts but the first case will fail to complete the smoke test. I also do not think there is a chance this script will work with Alfajores connection, as defined in `truffle.js` - `web3.eth.getAccounts()` will always return only one account (`process.env.CONTRACT_OWNER_CELO_ACCOUNT_PRIVATE_KEY)`.

Instead of adding these conditions and modifying bonding value earlier, I'd recommend exiting this script if the funding account does not have enough funds or if there are no at least 6 accounts in the system.

At the same time, we should update celo Docker image introduced in https://github.com/keep-network/keep-core/pull/2324 to create one more account.

With this logic, we'll be able to work with Celo locally the same way as we do with Ethereum. Alfajores will not be fully initialized but the same issue is for Ropsten for which we use separate provisioning scripts.

solidity/scripts/get-default-application-account.js
```diff
@@ -1,5 +1,11 @@
 module.exports = async function () {
   const accounts = await web3.eth.getAccounts()
-  console.log(`${accounts[6]}`)
+  // arbitrary chosen account for default application
+  if (accounts.length < 7) {
+    // Assign the last account address as application account
+    console.log(`${accounts[accounts.length - 1]}`)
+  } else {
+    console.log(`${accounts[6]}`)
+  }
```
pdyraga, line 9: In the context of this script, `7` is a magic number. 
I think we should consider altering the logic in `solidity/integration/smoke_test.js` and `stress_test.js` to take `accounts[1]` as the application. Leave a comment in this script why it's `accounts[1]` and if `accounts.length == 1` take `accounts[0]` - just one special case. WDYT?

scripts/initialize-celo.sh
```diff
@@ -0,0 +1,107 @@
...
+if [ ! -z ${client_app_address+x} ]; then
+  # Read user app when --application-address is set
+  CLIENT_APP_ADDRESS=$client_app_address
+fi
+
+# Read application address in case it was not set from the command line
+if [ -z ${CLIENT_APP_ADDRESS+x} ]; then
+  read -p "Enter client application address: " client_app_address
+  CLIENT_APP_ADDRESS=${client_app_address}
+fi
```
pdyraga, line 77: Is this possible to happen? We are setting it in `get-default-application-account.js`.

solidity/scripts/lcl-initialize.js
dimpar: Right, `smoke_test.js` won't work neither on Alfajores nor Ropsten with the current config. But isn't `smoke_test.js` is just one of our use cases? I mean, if you'd want to test some other functionality on testnet without having all 6 accounts at our disposal then running this script might still be helpful? This script authorizes, deposits and creates a new sortition pool for an application. With the current setup `initialize-celo.sh` will pass and replace all the addresses needed to run the client. 

On the other hand I see your point that if we don't have a full support for testnets yet, then anything between might just be confusing. If we'd rather not deal with testnets for now in these scripts, then I can certainly revert changes in `lcl-initialze.js` but also will need to modify `initialize-celo.sh` to inform that we can initialize only local network.

> At the same time, we should update celo Docker image introduced in keep-network/keep-core#2324 to create one more account

Sure, I will create a new account.

pdyraga: To open a new keep on any network, local or testnet, we need at least 3 accounts for operators. On Alfajores, we'll always have one when running this script. Agreed, sortition pool initialization is something we can do but as you say - let's either fully execute the initialization or not execute it at all and leave it for separate scripts, just like we do for Ropsten.
I like your idea about executing this script only for `local`. This is cleaner than exiting from it.

solidity/scripts/get-default-application-account.js
dimpar: `7` was magic because the default application was at `6` which I take was also chosen arbitrary. 

If we introduce more accounts in [docker for Celo](https://github.com/keep-network/keep-core/pull/2324) then maybe we should just revert the current change and leave `smoke_test.js` & `stress_test.js` as is?

scripts/initialize-celo.sh
dimpar: Yes, if you're on `Alfajores` network, then `get-default-application-account.js` won't be executed. If you won't provide an application address in a command line, then it will ask you to enter it here. But if we are leaning towards handling only `local` network as part of initialisation, then this `if` block won't be necessary.

solidity/scripts/get-default-application-account.js
pdyraga: We will probably still have a problem with Alfajores where we'll always have only one account. We can't skip `get-default-application-account.js` execution for non-local, because then, we'd also have to skip `lcl-client-config.js`.

Having a special case for just one account in the system in this particular case is fine IMO.

scripts/initialize-celo.sh
pdyraga: Ah! I was looking at `initialize.sh` and it behaves differently from `initialize-celo.sh` when it comes to `get-default-application-account.js`. If the changes we discussed allows keeping the behavior aligned, that's only better.

solidity/scripts/get-default-application-account.js
dimpar: Both `get-default-application-account.js` & `lcl-client-config.js` are called only from initialize scripts. If we limit the initialize scripts only to `local` then I think we don't have this problem anymore.

pdyraga: I'd love to have `lcl-client-config.js` called for non-local. For Alfrajores we can assume contract owner is the application and client configuration such as bonding or staking needs to be done from dashboard.

dimpar: 22f7d3c 

Given that we will create more Celo accounts running on docker, I left a default application account as it was implemented originally. IMO current implementation reflects a real scenario a bit better, when a `member` doesn't have the same address as `default application`, ie. both accounts[1]. 

If you think it is better to have them both assigned to `accounts[1]` I can change it.

solidity/scripts/lcl-initialize.js
dimpar: 221eba0 & 2ccc305

solidity/scripts/get-default-application-account.js
dimpar: PR with 3 extra accounts: https://github.com/keep-network/keep-core/pull/2332

scripts/initialize-celo.sh
pdyraga: This block can now be removed, correct?

dimpar: b553d44 yes, we can. In this case an application account on Alfajores will be either accounts[0] or an address passed in `--application-address`

scripts/initialize.sh
```diff
@@ -1,32 +1,76 @@
...
-# Dafault config files directory.
 CONFIG_DIR_PATH_DEFAULT=$(realpath -m $(dirname $0)/../configs)
+KEEP_ECDSA_PATH=$(realpath $(dirname $0)/../)
+KEEP_ECDSA_SOL_PATH=$(realpath $KEEP_ECDSA_PATH/solidity)
 
-# Read user config file path.
-read -p "Enter path to keep-ecdsa config files directory [$CONFIG_DIR_PATH_DEFAULT]: " config_dir_path
-CONFIG_DIR_PATH=${config_dir_path:-$CONFIG_DIR_PATH_DEFAULT}
+# Defaults, can be overwritten by env variables/input parameters
+NETWORK_DEFAULT="local"
```
pdyraga, line 14: Shouldn't `CONFIG_DIR_PATH_DEFAULT` be placed here, under `# Defaults, can be overwritten...`?

scripts/initialize.sh
```diff
@@ -1,32 +1,76 @@
...
 
-KEEP_ECDSA_PATH=$(realpath $(dirname $0)/../)
+help()
+{
+   echo -e "\nUsage: $0"\
+           "--keep-ecdsa-config-path <path>"\
```
pdyraga, line 19: Why not `--config-dir` [as we do](https://github.com/keep-network/keep-core/blob/f99e6e9f9283a7d794d9e8d72c04b6338eeae5cd/scripts/install.sh#L19) for the same parameter in `keep-core/scripts/install.sh` and in `keep-escda/scripts/start.sh`?

scripts/initialize.sh
```diff
@@ -1,32 +1,76 @@
...
+           "--application-address <address>"\
+           "--network <network>"
+   echo -e "\nCommand line arguments:\n"
+   echo -e "\t--keep-ecdsa-config-path : Path to keep-ecdsa client configuration file(s)"
```
pdyraga, line 23: Nitpick: whitespace before `:`

scripts/initialize-celo.sh
```diff
@@ -0,0 +1,101 @@
+#!/bin/bash
+set -euo pipefail
+
+KEEP_CORE_PATH_DEFAULT=$(realpath -m $(dirname $0)/../../keep-core)
```
pdyraga, line 4: Seems to be unused.

scripts/initialize-celo.sh
```diff
@@ -0,0 +1,101 @@
...
+CONFIG_DIR_PATH_DEFAULT=$(realpath -m $(dirname $0)/../configs)
```
pdyraga, line 5: Same question as for `initialize.sh` - shouldn't we place it in `# Defaults, can be overwritten...` section?

scripts/initialize-celo.sh
```diff
@@ -0,0 +1,101 @@
...
+KEEP_ECDSA_PATH=$(realpath $(dirname $0)/../)
+KEEP_ECDSA_SOL_PATH=$(realpath $KEEP_ECDSA_PATH/solidity)
+
+# Defaults, can be overwritten by env variables/input parameters
+KEEP_CELO_PASSWORD=${KEEP_CELO_PASSWORD:-"password"}
```
pdyraga, line 10: Is it used anywhere?

scripts/initialize-celo.sh
```diff
@@ -0,0 +1,101 @@
...
+NETWORK_DEFAULT="local"
+CONTRACT_OWNER_CELO_ACCOUNT_PRIVATE_KEY=${CONTRACT_OWNER_CELO_ACCOUNT_PRIVATE_KEY:-""}
```
pdyraga, line 12: Is it used anywhere? I see it is passed as `CONTRACT_OWNER_ACCOUNT_PRIVATE_KEY` environment variable when executing `scripts/lcl-initialize.js` and `lcl-client-config.js` but those two places in `initialize-celo.sh` are the only occurrences of `CONTRACT_OWNER_ACCOUNT_PRIVATE_KEY` in the entire `keep-ecdsa` I can find.

scripts/initialize-celo.sh
```diff
@@ -0,0 +1,101 @@
...
+help()
+{
+   echo -e "\nUsage: ENV_VAR(S) $0"\
+           "--keep-ecdsa-config-path <path>"\
+           "--application-address <address>"\
+           "--network <network>"
+   echo -e "\nEnvironment variables:\n"
+   echo -e "\tCONTRACT_OWNER_CELO_ACCOUNT_PRIVATE_KEY: Contracts owner private key on Celo"
+   echo -e "\nCommand line arguments:\n"
+   echo -e "\t--keep-ecdsa-config-path : Path to keep-ecdsa client configuration file(s)"
```
pdyraga, line 23: Maybe just `--config-dir`? We use `--config-dir` for the same purpose in `keep-ecdsa/scripts/start.sh` and in `keep-core/scripts/install.sh`.

scripts/initialize-celo.sh
```diff
@@ -0,0 +1,101 @@
...
+if [ ! -z ${client_app_address+x} ]; then
+  # Read user app when --application-address is set
+  CLIENT_APP_ADDRESS=$client_app_address
+fi
+
+# Run script.
+LOG_START='\n\e[1;36m'  # new line + bold + cyan
+LOG_END='\n\e[0m'       # new line + reset
+DONE_START='\n\e[1;32m' # new line + bold + green
+DONE_END='\n\n\e[0m'    # new line + reset
```
pdyraga, line 75: Can we define them at the beginning of the file just like we do in `scripts/initialize.sh`? Alternatively, update `scripts/initialize.sh` to define them at the same place as this script.

scripts/install-celo.sh
```diff
@@ -0,0 +1,100 @@
...
+
+help()
+{
+   echo -e "\nUsage: ENV_VAR(S) $0"\
+           "--keep-core-path <path>"\
+           "--network <network>"
+   echo -e "\nEnvironment variables:\n"
+   echo -e "\tKEEP_CELO_PASSWORD: The password to unlock local Celo accounts to set up delegations."\
+           "Required only for 'local' network. Default value is 'password'"
+   echo -e "\tCONTRACT_OWNER_CELO_ACCOUNT_PRIVATE_KEY: Contracts owner private key on Celo"
```
pdyraga, line 26: We could note here and in `scripts/install.sh` that it is required only for networks other than 'local'. We missed it in `keep-core` PR. 😬

scripts/initialize.sh
dimpar: 045db58

scripts/initialize-celo.sh
dimpar: 045db58

scripts/initialize.sh
dimpar: removed

scripts/initialize-celo.sh
dimpar: I think it was, but not now: f77cf6e

scripts/initialize-celo.sh
dimpar: f77cf6e

scripts/initialize-celo.sh
dimpar: Right, these scripts don't use it but if we want to run it against `Alfajores`, then we still need to set `CONTRACT_OWNER_ACCOUNT_PRIVATE_KEY` when executing this script. That's why we need to have in our `help()` function.

scripts/initialize-celo.sh
dimpar: e7e1fd6

scripts/install-celo.sh
dimpar: d0c6d91 if the wording is okay for you, I'll create a PR in `keep-core` updating the description as well.

scripts/initialize-celo.sh
dimpar: 6dc60d8

scripts/install-celo.sh
pdyraga: Looks good!

scripts/initialize-celo.sh
```diff
@@ -0,0 +1,96 @@
...
+CONFIG_DIR_PATH_DEFAULT=$(realpath -m $(dirname $0)/../configs)
+
+help()
+{
+   echo -e "\nUsage: ENV_VAR(S) $0"\
+           "--config-dir <path>"\
+           "--application-address <address>"\
+           "--network <network>"
+   echo -e "\nEnvironment variables:\n"
+   echo -e "\tCONTRACT_OWNER_CELO_ACCOUNT_PRIVATE_KEY: Contracts owner private key on Celo"
```
pdyraga, line 23: Should we note it is required only for non-local?

dimpar:
Related PR around priv key for non-local networks in `keep-core` https://github.com/keep-network/keep-core/pull/2335

pdyraga (approved):

PR #684, opened 2021-02-09
Deploy LPRewardsTBTCSaddle to local/Ropsten network

r-czajkowski:
This PR deploys the `LPRewardsTBTCSaddle` contract to the local/Ropsten network. It also fixes an issue that the KEEP token dashboard doesn't work because imports the `LPRewardsTBTCSaddle` that doesn't exist in the network.

### DIFF
--- a/solidity/scripts/liquidity-rewards-starter.js
+++ b/solidity/scripts/liquidity-rewards-starter.js
@@ -1,6 +1,8 @@
 const LPRewardsTBTCETH = artifacts.require("LPRewardsTBTCETH")
 const LPRewardsKEEPETH = artifacts.require("LPRewardsKEEPETH")
 const LPRewardsKEEPTBTC = artifacts.require("LPRewardsKEEPTBTC")
+const LPRewardsTBTCSaddle = artifacts.require("LPRewardsTBTCSaddle")
+
 const TestToken = artifacts.require("./test/TestToken")
 const KeepToken = artifacts.require(
   "@keep-network/keep-core/build/truffle/KeepToken"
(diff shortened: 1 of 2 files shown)


solidity/scripts/liquidity-rewards-starter.js
```diff
@@ -39,105 +41,72 @@ const mintAndApproveLPReward = async (
   })
 }
 
+const getWrappedTokenContract = async (LPRewardsContract) => {
```
nkuba, line 44: We're expecting a contract instance here so maybe name it starting with lowercase as we do elsewhere?
```suggestion
const getWrappedTokenContract = async (lpRewardsContract) => {
```

solidity/scripts/liquidity-rewards-starter.js
```diff
@@ -39,105 +41,72 @@ const mintAndApproveLPReward = async (
...
-
-    await lpRewardsTBTCETH.stake(staker1WrappedTokenBalance, { from: staker1 })
-    await lpRewardsTBTCETH.stake(staker2WrappedTokenBalance, { from: staker2 })
-
-    await lpRewardsKEEPTBTC.stake(staker1WrappedTokenBalance, { from: staker1 })
-    await lpRewardsKEEPTBTC.stake(staker2WrappedTokenBalance, { from: staker2 })
+    const LP_REWARDS = {
+      KEEP_ETH: {
+        contract: lpRewardsKEEPETH,
+        wrappedTokenContract: null,
```
nkuba, line 64: Do we have to define this property? Wouldn't it be enough to do `LP_REWARDS[key].wrappedTokenContract = await getWrappedTokenContract(...`?

solidity/scripts/liquidity-rewards-starter.js
```diff
@@ -39,105 +41,72 @@ const mintAndApproveLPReward = async (
...
+
+      await initLPRewardContract(
+        LP_REWARDS[key].contract,
+        keepToken,
+        lpRewardsOwner,
+        reward
+      )
+    }
+
+    for (let i = 8; i < 10; i++) {
```
nkuba, line 93: Could we include this loop in the one above so we won't need to do `for (const values of Object.values(LP_REWARDS))` again?

solidity/scripts/liquidity-rewards-starter.js
```diff
@@ -39,105 +41,72 @@ const mintAndApproveLPReward = async (
...
+        contract: lpRewardsKEEPTBTC,
+        wrappedTokenContract: null,
+      },
+      TBTC_SADDLE: {
+        contract: lpRewardsTBTCSaddle,
+        wrappedTokenContract: null,
+      },
+    }
+
+    for (const key of Object.keys(LP_REWARDS)) {
```
nkuba, line 80: Could this be `const value of Object.values(LP_REWARDS)`, so we replace `LP_REWARDS[key]` with `value`?

After we combine loops and do the above maybe we could simplify it to:
```js
for (const value of Object.values(LP_REWARDS)) {
  const wrappedTokenContract = await getWrappedTokenContract(value.contract)
  
  await initLPRewardContract(value.contract, keepToken, lpRewardsOwner, reward)

  for (let i = 8; i < 10; i++) {
    const staker = accounts[i]
    const amount = web3.utils.toWei(`${i * 100}`)

    await mintAndApproveLPReward(wrappedTokenContract, value.contract.address, staker, amount)
  }
}
       
```

solidity/scripts/liquidity-rewards-starter.js
```diff
@@ -39,105 +41,72 @@ const mintAndApproveLPReward = async (
...
-
-    await lpRewardsKEEPETH.stake(staker1WrappedTokenBalance, { from: staker1 })
-    await lpRewardsKEEPETH.stake(staker2WrappedTokenBalance, { from: staker2 })
-
-    await lpRewardsTBTCETH.stake(staker1WrappedTokenBalance, { from: staker1 })
-    await lpRewardsTBTCETH.stake(staker2WrappedTokenBalance, { from: staker2 })
-
-    await lpRewardsKEEPTBTC.stake(staker1WrappedTokenBalance, { from: staker1 })
-    await lpRewardsKEEPTBTC.stake(staker2WrappedTokenBalance, { from: staker2 })
+    const LP_REWARDS = {
```
nkuba, line 61: Maybe we don't need a map? Could it be a simple array? We could also get contracts instances in the loop.

```js
const LPRewardsContracts = [LPRewardsKEEPETH, LPRewardsTBTCETH, LPRewardsKEEPTBTC, LPRewardsTBTCSaddle]

for (const LPRewardsContract of LPRewardsContracts){
  const lpRewardsContract = await LPRewardsContract.deployed()
  ...
}
```

nkuba (changes requested):

r-czajkowski:
@nkuba ready for the next round.

nkuba (approved):

PR #686, opened 2021-02-10
Celo integration

lukasz-zimnoch:
Depends on: https://github.com/keep-network/keep-common/pull/66
Depends on: https://github.com/keep-network/keep-core/pull/2338
Depends on: https://github.com/keep-network/keep-ecdsa/pull/685
Depends on: https://github.com/keep-network/tbtc/pull/756
Depends on: https://github.com/keep-network/keep-core/pull/2344
Depends on: https://github.com/keep-network/keep-common/pull/69

This pull request introduces support for the Celo blockchain. The whole changeset can be divided into separated parts as described below.

First, a new `pkg/chain/celo` package has been created. It mirrors the structure of the existing `pkg/chain/ethereum` package. This package introduces a new `celo.CeloChain` struct which implements the `eth.Handle` interface. However, because the `eth.Handle` interface is tied to `go-ethereum` types (like `common.Address`), some temporary ducktape code was needed. That code converts `celo-blockchain` types used by `celo.CeloChain` to `go-ethereum` types used by `eth.Handle` to implement its method set. Once `eth.Handle` is decoupled from `go-ethereum`, all ducktape code will be removed.

The second part was introducing Celo contract bindings. To achieve that, the current content of the `pkg/chain/gen` directory has been moved to `pkg/chain/gen/ethereum` package. Next, a new `pkg/chain/gen/celo` package has been created. Its `Makefile` leverages the new contract generator params introduced by https://github.com/keep-network/keep-common/pull/66. After that change, both Ethereum and Celo contract bindings are generated during the usual `go generate ./...` command.

Last but not least, in order to make the `keep-ecdsa` client compiling with Celo, Go build tags have been used. The initial plan assumed we will integrate Celo to the codebase and will be still able to use a single `keep-ecdsa` binary for both chains. Unfortunatelly, `go-ethereum/crypto` and `celo-blockchain/crypto` packages can't be used simoultaneously at build time. Both packages use the same native C code which declares identical symbols. This causes a C linker error during Go build. To address that problem, all usages of the `go-ethereum/crypto` package in `keep-ecdsa` codebase have been identified. Then, each file received its counterpart for Celo and build tag rules assure the correct version is selected during the build time. The same has been done in `keep-core` (https://github.com/keep-network/keep-core/pull/2338) as this module is a direct dependency of `keep-ecdsa`. 

After all those changes, the `keep-ecdsa` client can work with Celo. To build the correct binary, `go build -tags celo` command should be used.  The client is started as usual. The config file structure also remains the same but all chain-specific properties should be placed under `celo` section instead of the usual `ethereum`.

Worth noting that nothing changed for the Ethereum version. The client should be built with the same `go build` command without any tags. There are no changes on the run and config side as well.


pdyraga:
> First, a new pkg/chain/celo package has been created. It mirrors the structure of the existing pkg/chain/ethereum package. This package introduces a new celo.CeloChain struct which implements the eth.Handle interface. However, because the eth.Handle interface is tied to go-ethereum types (like common.Address), some temporary ducktape code was needed

There is a possibility that we may first pull some subset of changes from https://github.com/keep-network/keep-ecdsa/pull/693, merge it to `master` and then go back to this PR eliminating the need for a ducktape FWIW.

lukasz-zimnoch:
> > First, a new pkg/chain/celo package has been created. It mirrors the structure of the existing pkg/chain/ethereum package. This package introduces a new celo.CeloChain struct which implements the eth.Handle interface. However, because the eth.Handle interface is tied to go-ethereum types (like common.Address), some temporary ducktape code was needed
> 
> There is a possibility that we may first pull some subset of changes from #693, merge it to `master` and then go back to this PR eliminating the need for a ducktape FWIW.

Yeah, this is the preferable way. The ducktape code is in case this PR will go first. It also makes testing possible.

Dockerfile
```diff
@@ -49,15 +49,25 @@ COPY ./solidity $APP_DIR/solidity
...
-RUN go generate ./.../gen
+RUN go generate ./...
 
 # Build the application.
 COPY ./ $APP_DIR/
 
+# Cleanup the `pkg/chain/gen` dir from unused chains bindings. Leave only
+# the ones which are currently in use. This helps reducing the size of
+# resulting binary and can prevent unexpected errors which may occur due to
+# transitive dependencies conflicts.
```
pdyraga, line 64: Does it mean that both celo and ethereum contract bindings are generated? I do not follow why we can have problems with transitive dependencies.

pkg/ecdsa/key_ethereum.go
```diff
@@ -0,0 +1,18 @@
...
+)
+
+// PublicKey holds a public key of the Signer.
+type PublicKey cecdsa.PublicKey
+
+// Marshal serializes Public Key to bytes in uncompressed form as described in
+// [SEC 1] section 2.3.3: `04 + <x coordinate> + <y coordinate>`
+func (pk *PublicKey) Marshal() []byte {
+	return crypto.FromECDSAPub((*cecdsa.PublicKey)(pk))
+}
```
pdyraga, line 18: Is this function used anywhere? I commented it out and I was able to build the client. Maybe we can avoid having two versions of this file.

If this function is no longer used I'd also love to get rid of our own `ecdsa.PublicKey` and use one from `crypto/ecdsa`.

pkg/chain/gen/celo/abi/BondedECDSAKeep.go
```diff
@@ -0,0 +1,2154 @@
+// Code generated - DO NOT EDIT.
```
pdyraga, line 1: I am confused why do we add ABI for `BondedECDSAKeep.go` but we do not for `BondedECDSAKeepFactory.go` and everything works. 🤔

Dockerfile
lukasz-zimnoch: Both `pkg/chain/gen/ethereum` and `pkg/chain/gen/celo` are generated during `go generate ./...`. During `go build` only one directory is included in the resulting binary due to build tags. If we use, for example, the ethereum connector, celo bindings are not imported at all so there are not included.

However, there is a problem during testing. According to the [docs](https://golang.org/pkg/cmd/go/internal/test/), `go test` builds a testing binary for each package. That means for `pkg/chain` tests both `gen/ethereum` and `gen/celo` are included in the test binary. In that case they not conflict but themselves but `gen/celo` depends on `celo-blockchain` parts which import `celo-bls-go`. As I observed that binary has problems on Alpine and cause errors while [running tests](https://app.circleci.com/pipelines/github/keep-network/keep-ecdsa/3462/workflows/0c91fe7c-4028-48e3-a4fa-3701ea5b0d92/jobs/13585). This is why I added this line in the Dockerfile. It acts as a workaround. I plan to revisit this in the near future and create a proper Dockerfile for Celo or refactor this one to support multiple chains if possible.

pkg/chain/gen/celo/abi/BondedECDSAKeep.go
lukasz-zimnoch: It's just to make `gosec` pass. We do the same for `pkg/chain/gen/ethereum/abi/BondedECDSAKeep.go`.

pkg/ecdsa/key_ethereum.go
lukasz-zimnoch: Looks you're right but this can be used for implicit marshaling somewhere. I'll make sure we are good with removing them and if so I'll do it.

cmd/connector_celo.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	address := celokey.Address.Hex()
+
+	balanceMonitor.Observe(
+		ctx,
+		address,
+		alertThreshold,
+		defaultBalanceMonitoringTick,
+	)
```
pdyraga, line 88: In [unchainded refactor](https://github.com/keep-network/keep-ecdsa/pull/693#discussion_r576783818) PR we reached the conclusion that the balance monitor can be Ethereum (or Ethereum-like) specific concept. If we get rid of it from `keep-core/pkg/chain/chain.go` interface, we can have it called in connectors with a specific currency passed to `Observe` function.

lukasz-zimnoch: Addressed in https://github.com/keep-network/keep-ecdsa/pull/686/commits/dd5a78dfd2b8fb27e1e14a3908966d560d23a949

Shadowfiend:
> The second part was introducing Celo contract bindings. To achieve that, the current content of the pkg/chain/gen directory has been moved to pkg/chain/gen/ethereum package. Next, a new pkg/chain/gen/celo package has been created. Its Makefile leverages the new contract generator params introduced by keep-network/keep-common#66. After that change, both Ethereum and Celo contract bindings are generated during the usual go generate ./... command.

Whoops, just a flag that this is not how this is meant to work. `go generate ./.../gen` is how we generally run `go generate`, and in this model, the `go:generate` directive in `pkg/chain/gen/gen.go` should trigger any downstream generation (which can be done---just have it run `sh -c go generate ./ethereum ./celo` or better yet, adjust the Makefile to recurse). See https://github.com/keep-network/keep-core#directory-structure and https://github.com/keep-network/keep-ecdsa#directory-structure for this note, and https://github.com/keep-network/keep-core/commit/0ac41c78b53377c1953234d350f47d2a0b321ecc for why: namely, to allow docker layer invalidation to only invalidate code-generation stages if code-generation directives have changed.

I don't think that should block this merge, but unless we have a very strong reason to diverge from this approach, I'd like to move back to it in a follow-up PR.

Shadowfiend:
… Also running `go generate` in the current situation seems to generate contract bindings for a lot more contracts than we need. Not entirely sure where the makefile filtering is going sideways, but another thing that can wait for a follow-on PR. I may even try to fix it in my first branch.

go.mod
```diff
@@ -14,13 +14,14 @@ require (
...
 	github.com/btcsuite/btcd v0.20.1-beta
+	github.com/celo-org/celo-blockchain v0.0.0-20210222234634-f8c8f6744526
 	github.com/ethereum/go-ethereum v1.9.10
 	github.com/gogo/protobuf v1.3.1
 	github.com/google/gofuzz v1.1.0
 	github.com/ipfs/go-log v1.0.4
-	github.com/keep-network/keep-common v1.3.1-0.20210217111137-6949bf58ebd8
-	github.com/keep-network/keep-core v1.3.2-0.20210217115335-806bf9a58b01
-	github.com/keep-network/tbtc v1.1.1-0.20210217115559-4a6a60e5274a
+	github.com/keep-network/keep-common v1.3.1-0.20210224153316-e59366b9ba03
```
pdyraga, line 22: We can take the opportunity and update the dependency to include the change from https://github.com/keep-network/keep-common/pull/68.

lukasz-zimnoch: See https://github.com/keep-network/keep-ecdsa/pull/686/commits/3e2b29eb7a83e8cb3b8e38c0dd3cdad6f65cba4d. It bumps `keep-common` to the head of https://github.com/keep-network/keep-common/pull/69 which also embeds changes from https://github.com/keep-network/keep-common/pull/68

cmd/connector_celo.go
```diff
@@ -0,0 +1,101 @@
...
+// the alert should be triggered.
+var defaultBalanceAlertThreshold = commoncelo.WrapWei(
+	big.NewInt(500000000000000000),
+)
+
+// defaultBalanceMonitoringTick determines how often the monitoring
+// check should be triggered.
+const defaultBalanceMonitoringTick = 10 * time.Minute
+
+func connectChain(
```
pdyraga, line 32: All of this code, including defaults, can be hidden behind `ethereum.Connect` (see https://github.com/keep-network/keep-core/pull/2344#discussion_r582923438). This will let us minimize the code inside the build-tag-dependent connector and we will no longer have problems with `chain.Handle` interface and how to instantiate Ethereum-specific balance monitor.

Shadowfiend: Since this is now a blocker for the chain refactor stuff, thoughts on merging this into master as-is (I'm making the assumption that it already works) and iterating on it on master?

pdyraga: I haven't had a chance to review code generator changes today but I am fine with merging the rest of the code as-is and addressing comments in a follow-up PR. If you can look at the code generator changes @Shadowfiend and they look good to you, please go ahead and merge.

Shadowfiend (approved):
Latest set of differences looks good, approving per [@pdyraga's note](https://github.com/keep-network/keep-ecdsa/pull/686#discussion_r583060821).

PR #688, opened 2021-02-12
Removing GOPRIVATE configuration from Dockerfile

michalinacienciala:
Go dependencies are no longer stored in private repositories,
making the step with configuration of GOPRIVATE in the Dockerfile
obsolete.


nkuba:
Can we also update Circle CI config not to pass this argument? https://github.com/keep-network/keep-ecdsa/blob/2863280a24411b17163c26b99aa77d9372c92d49/.circleci/config.yml#L103

nkuba (approved):

michalinacienciala:
> Can we also update Circle CI config not to pass this argument?
> 
> https://github.com/keep-network/keep-ecdsa/blob/2863280a24411b17163c26b99aa77d9372c92d49/.circleci/config.yml#L103

Done, #700.

PR #690, opened 2021-02-12
Updated .dockerignore file

nkuba:
Updated file with files and directories that we don't want to copy to
Docker building context.

We can ignore all hidden directories and files, infrastructure, scripts
and staker-rewards directories, more files and adocs as they are not
used by the Docker image.

We also ignored content of solidity directory with an exclusion for
contracts and NPM package configuration that is needed for Go code
generation.

These changes reduced data sent to the docker building context on local
machine from ~1.5 GB to ~2.5 MB.


.dockerignore
```diff
@@ -1,8 +1,29 @@
...
+# NPM stuff.
+**/node_modules/*
+
+# Solidity stuff.
+# We want to include only bare contracts and NPM package configuration for Go code
+# generator.
+solidity/
+!solidity/contracts
+!solidity/package.json
+!solidity/package-lock.json
```
Shadowfiend, line 25: How much does this gain? It feels like most of the extra heft from the solidity dir should be hit by the node_modules exclusion above, is that not the case?

.dockerignore
```diff
@@ -1,8 +1,29 @@
...
+scripts/
+staker-rewards/
 
-solidity/node_modules/
-solidity/build/
+# Top-level files unrealted to the build.
+CODEOWNERS
+crytic-config.json
+Dockerfile
+*.adoc
```
Shadowfiend, line 14: This isn't really top-level, eh?

Shadowfiend (approved):
This is glorious, and despite having a couple of small comments in the mix I'm going to let it roll on in as they're extra-minor.

Shadowfiend:
One small note worth flagging: that 1.5GB context is on the local env. If we look at the [Circle build for the docker image](https://app.circleci.com/pipelines/github/keep-network/keep-ecdsa/3475/workflows/457fc983-e0f2-4e35-847e-0c107ed5e8a1/jobs/13654/parallel-runs/0/steps/0-104), we see that on a fresh checkout the docker build context is ~21MB. This branch's [Circle build](https://app.circleci.com/pipelines/github/keep-network/keep-ecdsa/3378/workflows/80510b95-642b-4d83-ad2e-defbd74207ae/jobs/13168/parallel-runs/0/steps/0-104) does show 2.3MB though, so it's still a factor of 10 savings :muscle::muscle::muscle:

.dockerignore
nkuba: Yep, you're right. Maybe I was trying to make it too kosher. From the CI perspective, there are mostly files that are not heavy but unnecessary in the image. From the local env perspective, there's also `build` directory which is ~37MB. The filters are not about size but more about purity.

```
✗ du -hs solidity/* | sort -rh | head -10
715M    solidity/node_modules
 37M    solidity/build
1.1M    solidity/package-lock.json
412K    solidity/test
328K    solidity/contracts
 60K    solidity/scripts
 24K    solidity/migrations
 16K    solidity/integration
4.0K    solidity/truffle.js
4.0K    solidity/test-environment.config.js
```

.dockerignore
nkuba: With `.dockerignore` it will ignore only `*.adoc` files in the root directory. It works in a different way than `.gitignore` which would ignore files recursively. Was that your concern?

Shadowfiend: Of course it works differently, why wouldn't it work differently? 🙃 Every once in a while I remember that software engineering is really just a game of how many slightly different models to do the exact same thing you can remember at the same time.

nkuba: 🤷🏻‍♂️ 🤷🏻‍♀️

PR #691, opened 2021-02-12
Automatically Broadcast BTC Recovery Address

beaushinkle:
This PR marks the beginning of Liquidation Recovery Saga!

relevant threads: https://www.flowdock.com/app/cardforcoin/tech/threads/IME_6wBhIH6rYl2qUC_t0rQYggl https://www.flowdock.com/app/cardforcoin/tech/threads/yfCgVoCZf3BOyDm7_w3Q6kL9QXL

The aim of this PR is demonstrate a mechanism that broadcasts a BTC address to each other client in a signing group in the case of a `Terminated` event. The code draws heavily from `ready_protocol`, and I expect that the code in `tss.go` will find a new home elsewhere. The changes in `client.go` are all about exposing the data up the call-stack to `monitorKeepTerminatedEvent`.

#### To Test

Set up a whole local environment (geth, bitcoin, core, 3x ecdsa), then create a deposit (either through the dApp or through new my personal fav, 

```
$ bin/tbtc.js --rpc ws://127.0.0.1:8546 deposit new 10000000
```

Then, crash the collateralization to under 110% by running a script like:
```
const ETHBTCPriceFeedMock = artifacts.require("ETHBTCPriceFeedMock")

async function run() {
  try {
    const ethBtcPriceFeedMock = await ETHBTCPriceFeedMock.deployed()
    await ethBtcPriceFeedMock.setValue("03428571428602040")

    process.exit(0)
  } catch (ex) {
    console.error(ex)
    throw ex
  }
}

module.exports = run
```
in the `tbtc/solidity/scripts` directory via `$ truffle exec <script-name>`

After which, you can check the collateralization by running
```
$ bin/tbtc.js --rpc ws://127.0.0.1:8546 deposit list --vending-machine
```
and seeing an output like:
```
0x55B9684bF537FBCd50B9551195b7b4bC66d29f6c	ACTIVE	100000000	24
0xaBBd372A8891E39372b5b8DcC28B1065Dc052C9f	LIQUIDATION_IN_PROGRESS	10000000	0
```

Grab the address of one of of the `ACTIVE` deposits (in my case, that would be `0x55B9684bF537FBCd50B9551195b7b4bC66d29f6c`), and then liquidate it via:
```
$ bin/tbtc.js --rpc ws://127.0.0.1:8546 deposit 0x55B9684bF537FBCd50B9551195b7b4bC66d29f6c liquidate
```

Then watch your keep-ecdsa logs for messages that look like:
```
2021-02-11T17:02:50.006-0500    INFO    keep-tss    member [0x28bcfad75ee4c87b9089ba4c03a79b8148104da4] from keep [0xb2805515d0B9A58D8fE7D38eD71C97AF9393658F] announced supplied btc address [abc123] for liquidation recovery
2021-02-11T17:02:50.007-0500    INFO    keep-tss    member [0xb7021acafecf289fa345ecc95ecab18077530841] from keep [0xb2805515d0B9A58D8fE7D38eD71C97AF9393658F] announced supplied btc address [abc123] for liquidation recovery
2021-02-11T17:02:50.007-0500    INFO    keep-tss    member [0x65849fc91f8289b401e9d8f6d306b814b7f055c3] from keep [0xb2805515d0B9A58D8fE7D38eD71C97AF9393658F] announced supplied btc address [abc123] for liquidation recovery
```

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -18,10 +18,10 @@ import (
 	eth "github.com/keep-network/keep-ecdsa/pkg/chain"
 	"github.com/keep-network/keep-ecdsa/pkg/client/event"
 	"github.com/keep-network/keep-ecdsa/pkg/ecdsa/tss"
+	"github.com/keep-network/keep-ecdsa/pkg/ecdsa/tss/params"
 	"github.com/keep-network/keep-ecdsa/pkg/node"
 	"github.com/keep-network/keep-ecdsa/pkg/registry"
 	"github.com/keep-network/keep-ecdsa/pkg/utils"
-	"github.com/keep-network/keep-ecdsa/pkg/ecdsa/tss/params"
 )
 
 var logger = log.Logger("keep-ecdsa")
@@ -173,13 +173,13 @@ func Initialize(
 				ctx,
 				ethereumChain,
 				networkProvider,
-				keepAddress,
-				operatorPublicKey,
 				clientConfig,
 				tssNode,
+				operatorPublicKey,
 				keepsRegistry,
-				subscriptionOnSignatureRequested,
 				eventDeduplicator,
+				keepAddress,
+				subscriptionOnSignatureRequested,
 			)
 
 		}(keepAddress)
@@ -505,17 +505,18 @@ func generateKeyForKeep(
 		subscriptionOnSignatureRequested,
 		eventDeduplicator,
 	)
+
 	go monitorKeepTerminatedEvent(
 		ctx,
 		ethereumChain,
 		networkProvider,
-		keepAddress,
-		operatorPublicKey,
 		clientConfig,
 		tssNode,
+		operatorPublicKey,
 		keepsRegistry,
-		subscriptionOnSignatureRequested,
 		eventDeduplicator,
+		keepAddress,
+		subscriptionOnSignatureRequested,
 	)
 }
 
@@ -874,13 +875,13 @@ func monitorKeepTerminatedEvent(
 	ctx context.Context,
 	ethereumChain eth.Handle,
 	networkProvider net.Provider,
-	keepAddress common.Address,
-	operatorPublicKey *operator.PublicKey,
 	clientConfig *Config,
 	tssNode *node.Node,
+	operatorPublicKey *operator.PublicKey,
 	keepsRegistry *registry.Keeps,
-	subscriptionOnSignatureRequested subscription.EventSubscription,
 	eventDeduplicator *event.Deduplicator,
+	keepAddress common.Address,
+	subscriptionOnSignatureRequested subscription.EventSubscription,
 ) {
 	keepTerminated := make(chan *eth.KeepTerminatedEvent)
 
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -21,6 +21,7 @@ import (
 	"github.com/keep-network/keep-ecdsa/pkg/node"
 	"github.com/keep-network/keep-ecdsa/pkg/registry"
 	"github.com/keep-network/keep-ecdsa/pkg/utils"
+	"github.com/keep-network/keep-ecdsa/pkg/ecdsa/tss/params"
 )
 
 var logger = log.Logger("keep-ecdsa")
@@ -169,8 +170,13 @@ func Initialize(
 				eventDeduplicator,
 			)
 			go monitorKeepTerminatedEvent(
+				ctx,
 				ethereumChain,
+				networkProvider,
 				keepAddress,
+				operatorPublicKey,
+				clientConfig,
+				tssNode,
 				keepsRegistry,
 				subscriptionOnSignatureRequested,
 				eventDeduplicator,
@@ -182,6 +188,7 @@ func Initialize(
 	go checkAwaitingKeyGeneration(
 		ctx,
 		ethereumChain,
+		networkProvider,
 		clientConfig,
 		tssNode,
 		operatorPublicKey,
@@ -215,6 +222,7 @@ func Initialize(
 				generateKeyForKeep(
 					ctx,
 					ethereumChain,
+					networkProvider,
 					clientConfig,
 					tssNode,
 					operatorPublicKey,
@@ -245,6 +253,7 @@ func Initialize(
 func checkAwaitingKeyGeneration(
 	ctx context.Context,
 	ethereumChain eth.Handle,
+	networkProvider net.Provider,
 	clientConfig *Config,
 	tssNode *node.Node,
 	operatorPublicKey *operator.PublicKey,
@@ -305,6 +314,7 @@ func checkAwaitingKeyGeneration(
 		err = checkAwaitingKeyGenerationForKeep(
 			ctx,
 			ethereumChain,
+			networkProvider,
 			clientConfig,
 			tssNode,
 			operatorPublicKey,
@@ -325,6 +335,7 @@ func checkAwaitingKeyGeneration(
 func checkAwaitingKeyGenerationForKeep(
 	ctx context.Context,
 	ethereumChain eth.Handle,
+	networkProvider net.Provider,
 	clientConfig *Config,
 	tssNode *node.Node,
 	operatorPublicKey *operator.PublicKey,
@@ -372,6 +383,7 @@ func checkAwaitingKeyGenerationForKeep(
 			go generateKeyForKeep(
 				ctx,
 				ethereumChain,
+				networkProvider,
 				clientConfig,
 				tssNode,
 				operatorPublicKey,
@@ -392,6 +404,7 @@ func checkAwaitingKeyGenerationForKeep(
 func generateKeyForKeep(
 	ctx context.Context,
 	ethereumChain eth.Handle,
+	networkProvider net.Provider,
 	clientConfig *Config,
 	tssNode *node.Node,
 	operatorPublicKey *operator.PublicKey,
@@ -493,8 +506,13 @@ func generateKeyForKeep(
 		eventDeduplicator,
 	)
 	go monitorKeepTerminatedEvent(
+		ctx,
 		ethereumChain,
+		networkProvider,
 		keepAddress,
+		operatorPublicKey,
+		clientConfig,
+		tssNode,
 		keepsRegistry,
 		subscriptionOnSignatureRequested,
 		eventDeduplicator,
@@ -853,8 +871,13 @@ func monitorKeepClosedEvents(
 // happens unsubscribes from signing event for the given keep and unregisters it
 // from the keep registry.
 func monitorKeepTerminatedEvent(
+	ctx context.Context,
 	ethereumChain eth.Handle,
+	networkProvider net.Provider,
 	keepAddress common.Address,
+	operatorPublicKey *operator.PublicKey,
+	clientConfig *Config,
+	tssNode *node.Node,
 	keepsRegistry *registry.Keeps,
 	subscriptionOnSignatureRequested subscription.EventSubscription,
 	eventDeduplicator *event.Deduplicator,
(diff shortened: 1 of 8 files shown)


pkg/ecdsa/tss/marshaling.go
```diff
@@ -272,3 +272,25 @@ func (m *AnnounceMessage) Unmarshal(bytes []byte) error {
...
+	pbMsg := &pb.LiquidationRecoveryMessage{}
+	if err := pbMsg.Unmarshal(bytes); err != nil {
+		return err
+	}
+
+	m.SenderID = pbMsg.SenderID
+	m.BtcRecoveryAddress = pbMsg.BtcRecoveryAddress
+
+	return nil
+}
```
pdyraga, line 296: It would be good to add a simple marshal-unmarshal roundtrip unit test for this message. Please see `pbutils.RoundTrip` and how it is used in tests.

pkg/client/client.go
```diff
@@ -169,8 +170,13 @@ func Initialize(
 				eventDeduplicator,
 			)
 			go monitorKeepTerminatedEvent(
+				ctx,
 				ethereumChain,
+				networkProvider,
 				keepAddress,
+				operatorPublicKey,
+				clientConfig,
+				tssNode,
```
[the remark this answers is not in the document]
Shadowfiend: Yeah, we have a few places where stuff like this is happening that I ran into in #693 as well. Some of the work there may help, but not all of it.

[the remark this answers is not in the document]
beaushinkle: should be good to go here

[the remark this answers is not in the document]
beaushinkle: My vote is that we backlog some explicit work to re-org this after 693 settles

beaushinkle:
anything else ya'll want to see here?

pkg/ecdsa/tss/tss.go
```diff
@@ -169,3 +169,139 @@ func (s *ThresholdSigner) CalculateSignature(
...
+	memberBTCAddresses := make(map[string]string)
+
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case msg := <-msgInChan:
+				for _, memberID := range group.groupMemberIDs {
+					if msg.SenderID.Equal(memberID) {
```
pdyraga, line 215: Do we need to execute this check? Broadcast channel filters guarantee only signing group members can write to the channel.

pkg/ecdsa/tss/tss.go
```diff
@@ -169,3 +169,139 @@ func (s *ThresholdSigner) CalculateSignature(
...
+			"waiting for btc recovery addresses timed out after: [%v]", protocolReadyTimeout,
+		)
+	case context.Canceled:
+		logger.Infof("successfully gathered all btc addresses")
+
+		return nil
+	default:
+		return fmt.Errorf("unexpected context error: [%v]", ctx.Err())
+	}
+	return nil
```
pdyraga, line 306: I think this line is unreachable and we can safely remove it.

pkg/ecdsa/tss/marshaling_test.go
```diff
@@ -176,3 +176,37 @@ func TestFuzzAnnounceMessageRoundtrip(t *testing.T) {
 func TestFuzzAnnounceMessageUnmarshaler(t *testing.T) {
 	pbutils.FuzzUnmarshaler(&AnnounceMessage{})
 }
+
+func TestLiquidationRecoveryMessageRoundtrip(t *testing.T) {
```
pdyraga, line 180: Maybe `TestFuzzLiquidationRecoveryMessageRoundtrip` to keep it aligned with the rest of fuzz tests?

pkg/ecdsa/tss/tss.go
```diff
@@ -169,3 +169,139 @@ func (s *ThresholdSigner) CalculateSignature(
 
 	return signature, err
 }
+
+func BroadcastRecoveryAddress(
+	parentCtx context.Context,
+	groupID string,
+	memberID MemberID,
+	groupMemberIDs []MemberID,
+	dishonestThreshold uint,
```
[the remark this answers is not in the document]
Shadowfiend: This def feels squirmy to me. I wonder if we should be instantiating an extension type of `ThresholdSigner` that embeds `ThresholdSigner`, that is specific to the application in question. In general I think we need to make applications more first-class citizens rather than concepts that are strapped to the side of the system (you can see a lot of that at play in the refactorings in #693).

pkg/ecdsa/tss/tss.go
Shadowfiend: Feels like a good sign that `go vet` isn't running here.

@beaushinkle depending on how you went through setup here, it's possible you may not have pre-commit hooks set up (not sure if local-setup handles this step). If so, you may want to make sure to run `pre-commit install` in the repo root for keep-ecdsa. `brew install pre-commit` if it isn't already.

pkg/ecdsa/tss/tss.go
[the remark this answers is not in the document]
pdyraga: That's a very good point, the additional logic is application-specific. 👌

pkg/ecdsa/tss/tss.go
beaushinkle: What's the difference between here and `readyProtocol` (which is where I stole the group inclusion checking from)?

beaushinkle:
Okay - I think I covered everything except converting to `func (s *ThresholdSigner) BroadcastRecoveryAddress(`, since that looked still up in the air. Good for another look!

pkg/ecdsa/tss/tss.go
pdyraga: I think ready protocol was implemented before we had broadcast channel filters so we had to do this check everywhere. Need to confirm though.

pkg/ecdsa/tss/tss.go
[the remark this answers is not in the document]
Shadowfiend: Flagging that this is part of the refactoring we're punting as per the note [above](https://github.com/keep-network/keep-ecdsa/pull/691#discussion_r577748928).

pkg/ecdsa/tss/tss.go
Shadowfiend: Let's make a flag to check this and not block merging.

pkg/client/client.go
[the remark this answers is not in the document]
Shadowfiend: Now that I'm doing a deep dive I nearly fainted when I saw the list of parameters lololol.

pkg/client/client.go
```diff
@@ -492,12 +504,18 @@ func generateKeyForKeep(
...
-		keepAddress,
+		networkProvider,
+		clientConfig,
+		tssNode,
+		operatorPublicKey,
 		keepsRegistry,
-		subscriptionOnSignatureRequested,
 		eventDeduplicator,
+		keepAddress,
+		subscriptionOnSignatureRequested,
```
Shadowfiend, line 518: Another refactoring note here (cc @pdyraga ) is something else I noticed when doing the cross-chain refactor, which is that we've leaned pretty hard on functions where we could lean on methods to carry around shared state. In particular, we could probably use a `Keep` struct that things like `generateKey` and `monitorKeepTerminatedEvent` are defined on, and that contains the ThresholdSigner in turn.

pkg/client/client.go
```diff
@@ -905,6 +928,25 @@ func monitorKeepTerminatedEvent(
...
 				}
 
+				members, err := ethereumChain.GetMembers(keepAddress)
+				memberID := tss.MemberIDFromPublicKey(operatorPublicKey)
+				memberIDs, err := tssNode.AnnounceSignerPresence(
+					ctx,
+					operatorPublicKey,
+					keepAddress,
+					members,
+				)
```
Shadowfiend, line 938: Yet Another Refactoring Note™: I don't love that the signer presence, which is a prereq to I think all of our interactive TSS stuff, is an explicit step for the caller. It feels like BroadcastRecoveryAddress, public key generation, etc, should internally start with signer presence, and the caller just initiates the relevant protocol from the outside with a single method call. Open to feedback here though, esp from some of the folks who did the work :grimacing:

pkg/client/client.go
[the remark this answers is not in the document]
beaushinkle: yeah, it's not pretty 💃

pkg/ecdsa/tss/message.go
```diff
@@ -39,6 +39,15 @@ func (m *AnnounceMessage) Type() string {
 	return "ecdsa/announce_message"
 }
 
+type LiquidationRecoveryMessage struct {
```
Shadowfiend, line 42: Let's add a docstring to this and its exported methods in the next PR.

pkg/client/client.go
Shadowfiend: Actually, what happened to our state machine abstraction from keep-core's gjkr beacon protocol? Is it not a good fit here? Feels like unifying these under a common state machine concept and related test infrastructure could be useful (though might not be worth it).

pkg/ecdsa/tss/protocol_recovery.go
```diff
@@ -0,0 +1,147 @@
...
+	"fmt"
+	"time"
+
+	"github.com/keep-network/keep-core/pkg/net"
+)
+
+// BroadcastRecoveryAddress broadcasts and receives the BTC recovery addresses
+// of each client so that each client can retrieve the underlying bitcoin in
+// the case that a keep is terminated.
+func broadcastRecoveryAddress(
```
Shadowfiend, line 15: You can just make this exported in-place here and drop the exported passthrough stub in tss.go, fwiw. The package's exported functions are pulled from all the files in the package.

pkg/ecdsa/tss/protocol_recovery.go
```diff
@@ -0,0 +1,147 @@
...
+						if err != nil {
+							logger.Errorf(
+								"could not convert member ID to address for "+
+									"a member of keep [%s]: [%v]",
+								group.groupID,
+								err,
+							)
+							break
+						}
+						memberBTCAddresses[memberAddress] = msg.BtcRecoveryAddress
```
Shadowfiend, line 70: As we move towards the multichain refactoring, it will probably make sense to make `memberAddress` be a `chain.OperatorID` and have the map key be that type as well.

pkg/ecdsa/tss/protocol_recovery.go
```diff
@@ -0,0 +1,147 @@
...
+
+		// Send the message first time. It will be periodically retransmitted
+		// by the broadcast channel for the entire lifetime of the context.
+		sendMessage()
+
+		<-ctx.Done()
+		// Send the message once again as the member received messages
+		// from all peer members but not all peer members could receive
+		// the message from the member as some peer member could join
+		// the protocol after the member sent the last message.
```
Shadowfiend, line 107: Hmmm… Not sure this is entirely clear to me, but more to the point---this seems like something we should really wrap up in a helper if we're doing it in every protocol. Not necessarily something we want to tackle in liquidation recovery though.

pkg/ecdsa/tss/protocol_recovery.go
```diff
@@ -0,0 +1,147 @@
...
+		sendMessage()
+		return
```
Shadowfiend, line 109: Do we need to `return` here? 🤔

Shadowfiend (approved):
> Then, crash the collateralization to under 110% by running a script like

Would love to see a parameterized version of this script land in the tbtc repo as part of this work. This might look like e.g. taking a deposit address and target collateralization and doing the math on what to set the price to, or even passing a simple factor to reduce collateralization or price by.

Also left a few notes on refactorings and such, but none are blocking this PR or necessarily even this work path. Let's roll this PR on in!

PR #692, opened 2021-02-12
Moved arguments for versioning in Dockerfile

nkuba:
As we expect the arguments to be updating, especially REVISION to be set
to the current commit hash we need to define them closer to the place
they will be used. Currently, if we update a value for REVISION it causes
all subsequent layers to be executed and not fetched from the cache. There's
actually nothing changing besides the commit hash we pass to go build
command, so all the layers can be reused.


Shadowfiend (approved):
Lol oops.

PR #693, opened 2021-02-12
Unchained Refactor: Draft refactoring to pull chain-specific types out of pkg/chain

Shadowfiend:
This PR, while still in draft, is represents a handful of large refactorings stemming from
the goal of making `github.com/keep-network/keep-ecdsa/pkg/chain.Handle` independent
of the underlying chain implementation. This largely involved finding proper abstractions for
cases where concrete Ethereum-related types (such as `common.Address`) were being
used, as well as adapting the existing local and ethereum packages to implement these
abstractions.

In the process, a few FIXMEs were flagged for future refactoring that were not directly
related. Additionally, a few additional architectural changes are proposed here, which
are not strictly required for the cross-chain abstraction, but move the abstraction of the
`chain.Handle` interface to more of an application interface, rather than an interface that
is dictated by how the underlying implementation behaves. Here are several of the highest-
level changes, with some context on how they've played out in the diff:

- The `pkg/chain` package is now correctly named `chain`. It was, for historical reasons,
  still named `eth`, despite residing at `pkg/chain`.
- All instances of `github.com/ethereum/go-ethereum/common.Address` have been replaced
  by a series of `*ID` abstractions. `chain.OperatorID` now represents the chain-specific ID
  of an operator in the network. `chain.KeepID` represents the chain-specific ID of a keep in
  the network (for Ethereum, this _currently_ means a specific instance of a `BondedECDSAKeep`
  contract, but this is not a given for all chains; see more on architecture below).
  `chain.KeepMemberID` represents the ID of a single member of a keep. Currently on Ethereum,
  this is equivalent to an operator's address, but this is not a given for all chains or even for future
  Ethereum abstractions. Currently, the interface requires that a `KeepMemberID` be convertible
  to an `OperatorID`, and that an `OperatorID`, given an associated `KeepID`, be convertible to
  a `KeepMemberID`. Finally, `chain.KeepApplicationID` is the ID of a single application in the
  Keep network. Currently, only tBTC is considered such an application. All IDs can be converted
  to strings, and none have an `Equals` method for other IDs. That is, IDs are only comparable
  to each other as strings.
- The `chain.Handle` interface has gained a `PublicKeyToOperatorID` method, which is used
  in many cases where a bespoke `pubKeyToAddressFn` was being used.
- `BondedECDSAKeepFactory` and `BondedECDSAKeep` have been moved from embedded
  interfaces directly on `chain.Handle` to being separate abstract concepts, represented by the
  `chain.BondedECDSAKeepManager` and `chain.BondedECDSAKeepHandle` interfaces. These
  interfaces expose an API that allows for managing and interacting with the set of keeps as a
  whole and a particular keep, without having to pass around the KeepID everywhere. They also
  permit the passing around of a first-class handle to an already-resolved keep. This allows passing
  around the very restricted ability to interact with a particular keep or the set of keeps, without
  explicitly passing around access to the whole underlying chain apparatus. This lets us reason
  much more clearly about what a given function does and needs in isolation.
- `BondedECDSAKeepFactory` used to hold the functions that relate to application authorization
  and status. Instead, `chain.BondedECDSAKeepManager` exposes these on a specific application
  handle, described by a`chain.BondedECDSAKeepApplicationHandle`, and in particular for our
  purposes the specific implementation of `chain.TBTCHandle`. `chain.TBTCHandle` reworks the
  earlier `chain.TBTCHandle`. As with the other refactorings, `chain.TBTCHandle` now only
  exposes tBTC-related functionality, without pulling in the entirety of the `chain.Handle` API
  surface area.

Partially done or marked with FIXMEs are a few other potential changes:
- Block numbers are used to reference several things, but there is no guarantee that block
  numbers are a proper abstraction level for most of those things. As one example, the
  `chain.DepositRedemptionRequestedEvent` carried a block number that was used to look up
  timestamps in application code. This requires the chain handle to expose a way to map a block
  to a timestamp. Instead, this has now been folded in to the event construction process, so that
  the event itself already carries a timestamp. This has allowed the unexporting of `blockTimestamp`
  from the chain packages.
- Another example where block numbers are likely removable, that has not yet been dealt with,
  are past event lookups. These generally specify a start block, but exposing this as a start block
  rather than a different type of reference is likely a leak of the event abstraction.
- Finally, block numbers are used to determine finality. This should be moved to a chain-specific
  way of implementing such finality, tucked behind a method that expresses the intent of achieving
  sufficient finality.

The overall goal of this PR is to show a clear path towards chain abstractions that are as high-level
as possible, making for more expressive and straightforward keep-specific and application-specific
code. This PR by itself is currently incomplete; in particular, it asks several questions that it does
not provide the answers to.

In general, the principle that it introduces, and that isn't fully realized but that I feel we should
aim towards, is the reduction of the scope touched by various functions in the system. Tracking
specific entities instead of the whole chain handle is a start, but there are still too many places
that require the chain handle itself in order to deal with various specific details. Figuring out how
to wrap these so that the calling application is dealing with them less is a good goal. Figuring out
how to pull finality into the expectations that the chain interface expects is likewise a strong goal.
These are hinted in these changes, but they're more a direction than a complete idea at this point.

The questions we need to answer in this PR are:
- Is the direction correct? This PR can be split into smaller pieces to move to the same end goal
  if so, rather than trying to review a single monster PR.
- Which of the open suggestions (marked as FIXME or TODO) are good ideas?
- Are there any high-level architectural decisions proposed here that are a bad idea?
- What other questions arise, and how do we want to address them?

In this sense, this PR is a sort of RFC focused on code. The last question is whether the right
next step is to close this PR and write an RFC, rather than close this PR and start creating smaller
ones based on the work here.


Shadowfiend:
Oh good, a package-lock update somehow sneaked in :sob::sob::sob::sob:

pkg/chain/chain.go
```diff
@@ -1,27 +1,83 @@
...
+
+type OperatorID interface {
+	fmt.Stringer
+
+	// OperatorIDs should be convertible to an associated KeepMemberID given the
+	// id of the keep in question. This may be a 1-to-1 correspondence, i.e. a
+	// simple interface cast, or it may be an internally-tracked association on
+	// the host chain. The conversion _may but is not guaranteed to be_ nil if
+	// the operator is not in the specified keep.
+	KeepMemberID(keepID KeepID) KeepMemberID
```
pdyraga, line 37: I do not understand why do we need to pass `keepID` as a parameter. In all usages of `KeepMemberID` I looked at, this function assumes conversion succeeded.

pkg/chain/chain.go
```diff
@@ -1,27 +1,83 @@
...
+}
+
+// Should be chain-specific, unique across chains, and simple ([A-Za-z0-9-_.]).
+// This can generally mean prefixing the common name of the chain followed by a
+// dash followed by the chain-specific identifier (e.g., the contract address
+// for Ethereum). For legacy reasons, Ethereum KeepIDs should always be the bare
+// Keep contract address.
+//
+// Should not be expected to be comparable between types or amongst themselves.
```
pdyraga, line 46: Should not be expected to be comparable because the bare Ethereum-format address on Celo chain obtained from `fmt.Stringer` may look the same as the bare Ethereum-format address on Ethereum chain obtained from `fmt.Stringer` but they are IDs of two separate keeps?

pkg/chain/chain.go
```diff
@@ -1,27 +1,83 @@
...
+
+// package chain in keep-core -- or keep-common/chain?
+// type BaseHandle interface {
+// 	BlockCounter() (BlockCounter, error)
+// 	StakeMonitor() (StakeMonitor, error)
+// 	BalanceMonitor() (BalanceMonitor, error)
+// }
+
+// FIXME BalanceMonitor may be an Ethereum-specific concept, kick it
+// FIXME from the chain interface?
```
pdyraga, line 59: If this is more than 10 minutes of work, I'd consider leaving FIXME but not bothering about it for now. All the potential integrations I can think of have a concept of operator's account balance.

pkg/chain/chain.go
```diff
@@ -30,135 +86,162 @@ type Handle interface {
...
+	GetKeepAtIndex(keepIndex *big.Int) (BondedECDSAKeepHandle, error)
+	// GetKeepAtIndex returns a handle to the keep with the given identifier.
+	GetKeepWithID(keepID KeepID) (BondedECDSAKeepHandle, error)
+
+	// IsOperatorAuthorized checks if the factory has the authorization to
+	// operate on stake represented by the provided operator.
+	IsOperatorAuthorized(operator OperatorID) (bool, error)
+}
+
+type BondedECDSAKeepApplicationHandle interface {
```
pdyraga, line 124: Where/How can I obtain an instance of this interface?

pkg/chain/ethereum/bonded_ecdsa_keep_handle.go
```diff
@@ -0,0 +1,343 @@
...
+	"github.com/keep-network/keep-ecdsa/pkg/chain/gen/contract"
+	"github.com/keep-network/keep-ecdsa/pkg/ecdsa"
+	"github.com/keep-network/keep-ecdsa/pkg/utils/byteutils"
+)
+
+type bondedEcdsaKeep struct {
+	handle   *ethereumChain
+	contract *contract.BondedECDSAKeep
+
+	contractAddress common.Address // FIXME Expose address on contract.BondedECDSAKeep?
```
pdyraga, line 21: > // FIXME Expose address on contract.BondedECDSAKeep?

Not sure if I follow - do we want to create multiple instances of `*contract.BondedECDSAKeep`?

internal/testdata/config.toml
```diff
@@ -32,5 +32,5 @@
 	PreParamsGenerationTimeout = "6m37s"
 	PreParamsTargetPoolSize = 36
 
-[Extensions.TBTC]
+[ethereum.ContractAddresses.TBTC]
```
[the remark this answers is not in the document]
Shadowfiend: > in case we add another application

If we add another application, it'll still be an Ethereum contract if it's on Ethereum, and a different one if it's not. I'm not sure I follow, but the current structure didn't make a lot of sense to me---definitely open to revising though, I didn't get a chance to dig into the PRs and read the reasoning on this.

pkg/chain/chain.go
```diff
@@ -1,27 +1,83 @@
-// Package eth contains interface for interaction with an ethereum blockchain
-// along with structures reflecting events emitted on an ethereum blockchain.
-package eth // TODO: rename; this can be any host chain
+// Package chain contains interface for interaction with a host chain along with
+// structures reflecting events emitted on such a chain.
+package chain // TODO: rename; this can be any host chain
```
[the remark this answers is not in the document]
Shadowfiend: Oh. Snap. :grin:

cmd/start.go
```diff
@@ -140,12 +141,25 @@ func Start(c *cli.Context) error {
...
 	networkProvider, err := libp2p.Connect(
 		ctx,
 		config.LibP2P,
 		networkPrivateKey,
 		libp2p.ProtocolECDSA,
-		firewall.NewStakeOrActiveKeepPolicy(ethereumChain, stakeMonitor),
+		firewall.NewStakeOrActiveKeepPolicy(
+			ethereumChain.PublicKeyToOperatorID,
+			keepManager,
+			tbtcHandle,
```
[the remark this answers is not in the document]
Shadowfiend: Whoops, I think this is a leftover from an earlier rearrangement. Will drop it.

pkg/chain/chain.go
```diff
@@ -30,135 +86,162 @@ type Handle interface {
...
-	// to a keep.
-	RegisterAsMemberCandidate(application common.Address) error
+// BondedECDSAKeepManager is an interface whose implementations allow for
+// managing, monitoring, and creating bonded ECDSA keeps as well as the
+// operator's eligibility for participation in such keeps.
+type BondedECDSAKeepManager interface {
+	// TBTCApplicationHandle returns a handle for interacting with the tBTC
+	// application associated with this BondedECDSAKeepManager. Returns nil with
+	// an error if no tBTC application exists for this manager.
+	TBTCApplicationHandle() (TBTCHandle, error)
```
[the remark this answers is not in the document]
Shadowfiend: I could see it going either way. The tBTC application is a BondedECDSAKeep application. I originally played with having this return a `BondedECDSAKeepApplicationHandle`, but realized that just added complexity to consumers (suddenly you have to instantiate the tBTC stuff separately with the same stack of chain dependencies) and the gain in flexibility wasn't really clear to me. `TBTCHandle` embeds `BondedECDSAKeepApplicationHandle`, so any functions that should be generic across applications can be, while someone who is looking specifically for this application can get the specific information.

This also allows easier iteration on a new chain version, because the methods all descend from the `chain.Handle` interface's methods, so if `chain.Handle` is implemented you know you're good to go. That wasn't a formal requirement, it just sort of fell out informally as I went through the code.

pkg/chain/chain.go
Shadowfiend: Yeah, it doesn't matter for Ethereum, but my observation here is that there's no reason to assume that a Keep member's id is the same as their operator id (indeed, the split in types is meant to reflect that). If it isn't guaranteed they're the same, then the Keep member's id is likely specific to the keep, in which case if we're going from operator id to keep member id, then the function would need some way to connect the operator id to the keep member id for a particular keep---necessitating a keep ID.

This might be a little bit of premature design, but it feels much more correct in the new typing structure.

pkg/chain/chain.go
Shadowfiend: Yes, and meaning to say that `==` should not be expected to work “correctly”. Right now as these are structured, the only way folks should probably compare them is using `.String() == .String()`, or by casting internally to known internal types (which of course risks problems if someone manages to cross an id from one chain to another).

pkg/chain/chain.go
Shadowfiend: Yeah, I don't think this one's that important. I think BlockCounter is going to have to be reworked though.

pkg/chain/chain.go
Shadowfiend: TBTCApplicationHandle embeds it---I would expect the same to happen for any other application. So basically, off of the BondedECDSAKeepManager.

pkg/chain/ethereum/bonded_ecdsa_keep_handle.go
Shadowfiend: The open question is whether we should be exposing this address via an exported function. This note predates `chain.KeepID` and the `ID()` function (i.e., the answer was “yes” haha).

internal/testdata/config.toml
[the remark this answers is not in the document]
pdyraga: I think the current structure makes sense in case the application has more configurable properties than just Ethereum contract address or when it has more than one contract address to configure. I am not sure where to, for example, put BTC configuration properties for liquidated deposit recovery. This is something TBTC-specific, so we could have all those properties grouped together under `[Extensions.TBTC]`.

config/config.go
```diff
@@ -37,6 +37,8 @@ type SanctionedApplications struct {
 	AddressesStrings []string `toml:"Addresses"`
 }
 
+// TODO This should be IDs(chain.Handle) ([]chain.KeepApplicationID, error)
+// TODO It should also be per-chain.
```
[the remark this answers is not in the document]
pdyraga: Well... based on my [previous comment](https://github.com/keep-network/keep-ecdsa/pull/693#discussion_r577003611), `[ethereum.SanctionedApplications]` and `[ethereum.Extensions.TBTC]` are probably not a good idea.

internal/testdata/config.toml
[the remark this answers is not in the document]
Shadowfiend: The recovery stuff is a great point. I think it may make sense to consider that chain-centrically as well though---we haven't historically had a `bitcoin` namespace in the config, but that may make sense here...

On the other hand I wonder if an operator may prefer to use some recovery data for one chain and different recovery data for a different one… I could, for example, see them using different xpubs with different derivation paths in these cases to make it easier to reconcile accounting for liquidations. So I don't think we _can_ consider these the same thing. I think we should nest the configurations under the host chain, and consider providing a way to set “defaults” for those who don't want to configure per-chain.

Shadowfiend:
Added a bunch of changes here that leave tests compiling and running. I also dropped a bunch of FIXMEs and TODOs that had already been handled, and filled in a few remaining FIXMEs that are likely not blocking.

Tomorrow morning I'm going to synthesize the work in this branch into a more coherent PR description so we can have another couple of eyeballs on what's going on.

internal/testdata/config.toml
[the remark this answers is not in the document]
pdyraga: I like the idea about `bitcoin` namespace. My concern is how much the configuration properties we need to add are BTC-specific or TBTC-specific. For example, `ElectrsURL` should nicely fit into `bitcoin` namespace. `MaxFeePerVByte` should fit nicely into both `bitcoin` and `extensions.tbtc`. On the other hand, `BTCRefunds` fits better into `extensions.tbtc` and clearly describes what this property is for.

pkg/chain/chain.go
pdyraga: Yeah, totally makes sense. I am concerned about the potential `nil` conversion though because it means we probably need to do a post-conversion check for every call and report an error if something goes wrong. I also don't fully understand in what scenarios `nil` could be returned.

For example, if `KeepMemberID` is a mix of `KeepID` and `MemberID`, this function should always succeed. 

On the other hand, if the result depends on the on-chain state and this function is expected to validate if the given `MemberID` is really a member of a keep with the given `KeepID` before doing the conversion, we should let this function report `error`. But it feels wrong to execute such checks for conversion functions.

cmd/start.go
```diff
@@ -140,12 +141,21 @@ func Start(c *cli.Context) error {
...
+	}
+
 	networkProvider, err := libp2p.Connect(
 		ctx,
 		config.LibP2P,
 		networkPrivateKey,
 		libp2p.ProtocolECDSA,
-		firewall.NewStakeOrActiveKeepPolicy(ethereumChain, stakeMonitor),
+		firewall.NewStakeOrActiveKeepPolicy(
+			ethereumChain.PublicKeyToOperatorID,
```
pdyraga, line 155: This will be a fun spot for two or more p2p networks. Operator ID may be the same for Celo and Ethereum but we do not want Celo and Ethereum p2p client networks to mix with each other. We may, for example, expect that the string representation of Operator ID is prefixed with a network name... What do you think?

Shadowfiend: Yep, the [note on KeepID](https://github.com/keep-network/keep-ecdsa/blob/faa12a770f65b7fecd48ff66ce35a69d8ed21be6/pkg/chain/chain.go#L40-L46) will apply across ids. I expect Ethereum ones will remain unprefixed for historical reasons, unless/until we can confirm that we can prefix safely. In particular, I want to make sure the refactor is network-compatible with the previous version.

internal/testdata/config.toml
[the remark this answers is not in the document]
Shadowfiend: > I think we should nest the configurations under the host chain, and consider providing a way to set “defaults” for those who don't want to configure per-chain.

This is where I'm settling here, fwiw.

pkg/chain/chain.go
Shadowfiend: The `KeepMemberID` *could be nil* if the operator is not in the keep. Right now that won't be the case in any implementations. We could deal with this using an `error`, but that felt like too much with the current setup (though it may make sense to add it later). There are very few places where this conversion happens at the moment, and I think discouraging the conversion is the right thing---this is why `IsThisOperatorMember` exists on keeps now, for example.

Frankly, I would rather _not be able to convert between these types_ -- it just wasn't feasible the way things are currently wired up.

pkg/chain/chain.go
```diff
@@ -1,164 +1,239 @@
...
 type Handle interface {
-	// Address returns client's ethereum address.
-	Address() common.Address // TODO: use implementation-agnostic type
+	// Name returns the name of this chain handle. The name should be
+	// descriptive of the chain this handle represents; for example, for
+	// Ethereum it might return "Ethereum", for Celo "Celo". It can optionally
+	// include additional information if it would be useful in logs or debug
+	// output; for example, Ethereum handles might include network id
+	// information.
+	Name() string
```
nkuba, line 67: Would it be worth adding a `ChainType() string` if we decide to append additional information to the chain name in `Name()` function?

pkg/chain/chain.go
```diff
@@ -1,164 +1,239 @@
...
+	// an error if no tBTC application exists for this manager.
+	TBTCApplicationHandle() (TBTCHandle, error)
 
 	// OnBondedECDSAKeepCreated installs a callback that is invoked when an
 	// on-chain notification of a new bonded ECDSA keep creation is seen.
 	OnBondedECDSAKeepCreated(
 		handler func(event *BondedECDSAKeepCreatedEvent),
 	) subscription.EventSubscription
 
+	// GetKeepCount returns number of keeps.
```
nkuba, line 105: I know the scope of this PR is a higher level, but leaving the comment here for the future. As this is no longer a function on `BondedECDSAKeepFactory` interface I'd mention in the comment that it's a number of keeps created by the associated factory.

pkg/chain/chain.go
```diff
@@ -1,164 +1,239 @@
...
+	GetKeepCount() (*big.Int, error)
+
+	// GetKeepAtIndex returns a handle to the keep at the given index.
+	GetKeepAtIndex(keepIndex *big.Int) (BondedECDSAKeepHandle, error)
+	// GetKeepAtIndex returns a handle to the keep with the given identifier.
```
nkuba, line 110:
```suggestion
	// GetKeepWithID returns a handle to the keep with the given identifier.
```

pkg/chain/chain.go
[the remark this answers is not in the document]
nkuba: At first, I wasn't into defining the application-specific handle here, I was thinking about something more generic. But after considering the complexity of being generic here it sounds fine to be tBTC specific.

pkg/chain/chain.go
```diff
@@ -1,164 +1,239 @@
...
-	// IsOperatorAuthorized checks if the factory has the authorization to
-	// operate on stake represented by the provided operator.
-	IsOperatorAuthorized(operator common.Address) (bool, error)
-
-	// GetKeepCount returns number of keeps.
-	GetKeepCount() (*big.Int, error)
-
-	// GetKeepAtIndex returns the address of the keep at the given index.
-	GetKeepAtIndex(keepIndex *big.Int) (common.Address, error)
+	UpdateStatusForApplication() error
```
nkuba, line 141: I'm wondering if it would be worth updating the functions names as they are exposed on the application handle now.
`IsRegisteredForApplication` -> `IsOperatorRegistered` or `IsRegistered`
`IsEligibleForApplication` -> `IsOperatorEligible` or `IsEligible`
`IsStatusUpToDateForApplication` -> `IsOperatorStatusUpToDate` or `IsStatusUpToDate`
`UpdateStatusForApplication` -> `UpdateOperatorStatus` or `UpdateStatus`

pkg/chain/chain.go
```diff
@@ -1,164 +1,239 @@
...
-	GetOpenedTimestamp(keepAddress common.Address) (time.Time, error)
+	GetOpenedTimestamp() (time.Time, error)
 
 	// PastSignatureSubmittedEvents returns all signature submitted events
 	// for the given keep which occurred after the provided start block.
 	// All implementations should returns those events sorted by the
 	// block number in the ascending order.
 	PastSignatureSubmittedEvents(
-		keepAddress string,
+		// FIXME how can this be divorced from a block number?
```
nkuba, line 236: Maybe passing a flag describing the point of time we want to use, so e.g. keep creation or last redemption.

pkg/chain/ethereum/bonded_ecdsa_keep_handle.go
```diff
@@ -0,0 +1,335 @@
...
+
+type bondedEcdsaKeep struct {
+	handle   *ethereumChain
+	contract *contract.BondedECDSAKeep
+
+	contractAddress common.Address
+}
+
+// GetKeepWithID returns a handle to the BondedECDSAKeep with the provided id.
+func (bekm *bondedEcdsaKeepManager) GetKeepWithID(
```
nkuba, line 25: Wouldn't it be better to move this function to `bonded_ecdsa_keep_manager.go `?

pkg/chain/ethereum/connect.go
```diff
@@ -122,6 +137,35 @@ func Connect(accountKey *keystore.Key, config *ethereum.Config) (*EthereumChain,
...
+}
+
+func (ec *ethereumChain) OperatorID() chain.OperatorID {
+	return ec.PublicKeyToOperatorID(&ec.accountKey.PrivateKey.PublicKey)
+}
+
+// combinedChainID represents all chain ids in one type: KeepMemberID,
+// OperatorID, KeepApplicationID, and KeepID. For the Ethereum chain, this is an
+// appropriate abstraction since all are underlied by the common.Address type.
+type combinedChainID common.Address
```
nkuba, line 151: Maybe this could be named `commonChainID`? IMHO `combinedChainID` would be a good fit if only `KeepMemberID` and `OperatorID` are the same. Here we're defining a common type for **all** the IDs for the chain.

nkuba (commented):
Dropping the first part of the comments.

config/config.go
```diff
@@ -37,6 +37,10 @@ type SanctionedApplications struct {
 	AddressesStrings []string `toml:"Addresses"`
 }
 
+// FIXME Update to improved contract address structure that removes the need for
+// SanctionedApplications as an entry?
+// TODO This should be IDs(chain.Handle) ([]chain.KeepApplicationID, error)
```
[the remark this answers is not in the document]
Shadowfiend: Yeah, this is a good point. And considering this in the context of some of the other discussion about these being tucked under the actual chain config objects, they can be non-generic---that is, `config.ethereum.SanctionedApplications`---at which point the type can be non-generic as well.

pkg/chain/chain.go
Shadowfiend: I couldn't think of a good use case, but it's def a possibility.

I think if we went `ChainType` we might want to make it an enum/`iota`-based constant set, though.

pkg/chain/chain.go
Shadowfiend: One of the things I'm interested in thoughts on that I didn't call out explicitly is whether it makes sense to call this `BondedECDSAKeepManager` instead of sticking to `BondedECDSAKeepFactory`.

pkg/chain/chain.go
Shadowfiend: Love it.

pkg/chain/chain.go
Shadowfiend: I was actually thinking last night about using the event itself, so e.g. `sinceEvent Event`, where all the structs implement the `Event` interface. But… Due to how interfaces work in Go I'm not sure if there's a good way of doing this generically without re-exposing the block stuff. I haven't done a deep dive, but it feels like it could be a promising direction.

pkg/chain/ethereum/bonded_ecdsa_keep_handle.go
Shadowfiend: I tried something interesting across this PR, which is making the first function in each file answer the question “how do I get one of these?” This is related to something I'd like us to do in packages as well, which is making the first function in the package's eponymous Go file (e.g. `ethereum.go` for `pkg/ethereum`, `chain.go` for `pkg/chain`) answer the question “how do I get the main types this package exposes?”

Definitely open to doing it the other way though.

pkg/chain/ethereum/connect.go
Shadowfiend: `combinedChainID` refers to the fact that it combines representation of all of the ID types under one internal type. `commonChainID` would be fine by me as well, though.

pkg/ecdsa/tss/member.go
```diff
@@ -3,10 +3,13 @@ package tss
...
 	"bytes"
 	"encoding/hex"
-	"github.com/keep-network/keep-core/pkg/operator"
 	"math/big"
+
+	"github.com/keep-network/keep-core/pkg/operator"
 )
 
+// FIXME How does this differ from an operator id? Why is it a distinct concept?
+// Improve name to reflect this.
```
lukasz-zimnoch, line 12: `MemberID` must be convertible to the network ID (`net.TransportIdentifier`). We need to know network IDs to open unicast channels between members. To obtain a network ID we need the operator public key and this is why `MemberID` looks as follows. Regarding the new structure, we should probably rename it to `MemberPublicKey` or `KeepMemberPublicKey` to not confuse it with `*ID` types. Alternatively, we could drop it and operate on bare `operator.PublicKey` but this is probably not for this PR as the refactoring will be rather massive.

pkg/ecdsa/tss/protocol_announce.go
```diff
@@ -55,27 +57,16 @@ func AnnounceProtocol(
 			case <-ctx.Done():
 				return
 			case msg := <-announceInChan:
-				// Since broadcast channel has an address filter, we can
-				// assume each message come from a valid group member.
```
lukasz-zimnoch, line 59: Maybe we can leave that comment as it clarifies why we don't check the sender of the message.

pkg/chain/key.go
```diff
@@ -1,10 +1,11 @@
-package eth
+package chain
 
 import (
 	"github.com/keep-network/keep-ecdsa/pkg/ecdsa"
 	"github.com/keep-network/keep-ecdsa/pkg/utils/byteutils"
 )
 
+// FIXME Why is this in the chain package? 🤔
```
nkuba, line 8: The serialization here is specific for Keep ECDSA contract, tBTC, and Ethereum, a mix of these three. The idea was that it is chain specific, but probably will be the same for other Ethereum-related chains. 

E.g. https://github.com/keep-network/keep-ecdsa/blob/4f172f1cedd89b794c679382f8dbdf40d4234bce/solidity/contracts/AbstractBondedECDSAKeep.sol#L594-L606

I'm fine with moving it somewhere else.

pkg/ecdsa/tss/signer.go
```diff
@@ -13,6 +15,8 @@ type ThresholdSigner struct {
 	// thresholdKey contains a signer's key generated for a threshold signing
 	// scheme. This data should be persisted to a local storage.
 	thresholdKey ThresholdKey
+
+	pubKeyToAddressFn func(cecdsa.PublicKey) []byte
```
lukasz-zimnoch, line 19: Where do we set this one? Also, we should make sure it's set after client restart as `ThresholdSigner` is unmarshaled from disk storage.

pkg/chain/event.go
```diff
@@ -1,15 +1,17 @@
...
-	KeepAddress     common.Address   // keep contract address
-	Members         []common.Address // keep members addresses
-	HonestThreshold uint64
-	BlockNumber     uint64
+	// KeepAddress          KeepID // keep contract address
+	Keep                 BondedECDSAKeepHandle
+	Members              []KeepMemberID // keep members addresses
+	HonestThreshold      uint64
+	BlockNumber          uint64
+	ThisOperatorIsMember bool
```
nkuba, line 14: I like the idea of defining `ThisOperatorIsMember`. 
It changes our previous unwritten rule a little bit as `BondedECDSAKeepCreatedEvent` is no longer a 1:1 representation of the on-chain event as it has some additional fields. 

We also need to remember to set `ThisOperatorIsMember` once creating the event. I'm a little afraid that we will forget to.

pkg/chain/chain.go
nkuba: At first, I found the rename controversial. But if the idea of this diverges from the idea of reflecting the `BondedECDSAKeepFactory` contract very closely it makes sense to rename it.

pkg/chain/chain.go
nkuba: > I couldn't think of a good use case, but it's def a possibility.

It depends if we will end up appending additional information to the `Name`. If yes, then in most cases it won't be necessary to log the whole Name but just a chain type (I love the idea with enum).

pkg/chain/ethereum/bonded_ecdsa_keep_handle.go
nkuba: Interesting idea. It sounds good to me if we move `func (bekm *bondedEcdsaKeepManager) GetKeepWithID(` to the mentioned eponymous `ethereum.go` file.

pkg/chain/chain.go
Shadowfiend: Worth flagging, I'm always thinking of log messages as I receive them as someone who isn't running a given node. For example, if every log message included the chain id, I wouldn't have to ask if a given client is running on mainnet or Ropsten. This is part of why I don't want to make the interface overly broad here---I think including the full `Name` is potentially a win, and it's up to the chain to decide what the relevant details are for that name.

I agree we could end up overbloating the name field though (I'm looking at you, web browser user agents). It feels like the right move might be starting with `Name`, and adding `ChainType` if we find it's necessary.

pkg/chain/chain.go
Shadowfiend: It may or may not right now, but I think we don't want these to be attached to each other. The underlying concept being that `chain` is the interface, but it dictates as little as possible about the underlying implementation for a specific supported chain.

pkg/chain/ethereum/bonded_ecdsa_keep_handle.go
Shadowfiend: Hmmm… Not sure if I follow, but also don't want to go too hard on this topic in this particular PR. It feels like a good fit for a smaller PR targeted at this specific refactoring.

pkg/ecdsa/tss/member.go
Shadowfiend: > we should probably rename it to MemberPublicKey or KeepMemberPublicKey

This is a great idea. If it's a public key, let's be clear that it's a public key. If it ceases to be a public key later on, we can look at why and adjust accordingly.

pkg/chain/key.go
Shadowfiend: Sounds like it might be a good fit for the new `ethlike` package or similar, perhaps? Def feels like something that should be pulled up to keep-common.

Shadowfiend:
Discussed this a bit with @pdyraga @nkuba @lukasz-zimnoch . The decision here is to apply these changes in more structured, piecewise fashion to the full codebase. I'll start the work and hand it off if I have to take off. My intended order of operations here is:

- eth -> chain package rename
- chain.*ID refactor
- structural refactor to remove a lot of the explicit ids from the interfaces in favor of `BondedECDSAKeepHandle`.
- structural refactor to remove explicit ids from application stuff in favor of `BondedECDSAKeepApplicationHandle`.
- structural refactor to move tBTC application into an extension of `BondedECDSAKeepApplicationHandle` instead of a standalone `chain.Handle` extension.

Exploration of finality refactoring and other FIXMEs may follow, or may stay parked for the time being.

Closing this PR, though comments here are still important to reference.

pkg/chain/chain.go
nkuba: Oh, now I see what you meant in the code comment. It makes perfect sense!

PR #694, opened 2021-02-15
Create GH Actions workflow for build&test of Go client

michalinacienciala:
and tBTC release processes in order to reduce human involvement and make
the processes faster and less error prone. A part of this task is
migration from CircleCI jobs to GitHub Actions. This commit creates a
GitHub Action workflow for building and testing of Go client.

At this point, solution utilizes EnricoMi/publish-unit-test-result-action@v1.7
action for publishing of the test results, but this may be replaced in the
future by use of other action or use of problem matchers.


.github/workflows/client.yml
```diff
@@ -0,0 +1,81 @@
...
+on:
+  push:
+    branches:
+      # TODO: Run on master after we're fully migrated from Circle CI
+      # - master
+      - "rfc-18/**"
+  pull_request:
+      branches:
+      # TODO: Run on all branches after we're fully migrated from Circle CI
+      - "rfc-18/**"
```
nkuba, line 13: We should probably add exclude paths for dirs like `docs`, `infrastructure`, and `staker-rewards`. But let's do it in the future, not in this PR.

.github/workflows/client.yml
```diff
@@ -0,0 +1,81 @@
...
+          mkdir test-results
+      - name: Run Go tests
+        run: |
+          docker run \
+            --volume $GITHUB_WORKSPACE/test-results:/mnt/test-results \
+            --workdir /go/src/github.com/keep-network/keep-ecdsa \
+            go-build-env \
+            gotestsum --junitfile /mnt/test-results/unit-tests.xml
+      - name: Publish unit test results
+        uses: EnricoMi/publish-unit-test-result-action@v1.7
```
nkuba, line 63: There's `v1.9` tag available. We could define simply `@v1` and I believe it should pull the latest `v1.x`.

nkuba (changes requested):

.github/workflows/client.yml
```diff
@@ -0,0 +1,81 @@
...
+      # to have no less than 15 GB after the cache is restored.
+      - run: sudo df -h
+
+      - name: Build Docker Build Image
+        uses: docker/build-push-action@v2
+        with:
+          target: gobuild
+          tags: go-build-env
+          build-args: |
+            REVISION=${{ github.sha }}
```
nkuba, line 46: I know I personally added it in keep-core, but can we comment it out. After consideration I think we shouldn't pass this argument during work on feature branches but only after the merge to master when we want to publish an image. Updating revision on each commit on feature branch causes longer docker image builds as it won't restore go client layer from cache but will rebuild it every time.
```suggestion
            # REVISION=${{ github.sha }} TODO: set revision only when we want to publish the image
```

nkuba (approved):

nkuba (approved):

PR #699, opened 2021-02-19
Create GH Actions workflow for test and lint of staker rewards

michalinacienciala:
and tBTC release processes in order to reduce human involvement and make
the process less error prone. A part of this task is migration from
CircleCI to GitHub Actions. This commit creates a GitHub Action workflow
for linting and testing of staker rewards.


.github/workflows/staker-rewards.yml
```diff
@@ -0,0 +1,80 @@
+name: Staker Rewards
+
+#TODO: extend the conditions once workflow gets tested together with other workflows
+on:
+  push:
+    branches:
+      # TODO: Run on master after we're fully migrated from Circle CI
```
nkuba, line 7: We can switch completely to GH Actions for staker-rewards checks as it's independent and no other CircleCI jobs rely on them.

nkuba: So let's remove all the CircleCI config for staker-rewards.

.github/workflows/staker-rewards.yml
```diff
@@ -0,0 +1,80 @@
...
+          restore-keys: |
+            ${{ runner.os }}-build-${{ env.cache-name }}-
+            ${{ runner.os }}-build-
+            ${{ runner.os }}-
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Run tests
+        run: npm test # or npm run test?
```
nkuba, line 50: Any of these will do.

nkuba (changes requested):

.github/workflows/staker-rewards.yml
michalinacienciala: Right, done.

nkuba (approved):

PR #700, opened 2021-02-19
Removing GOPRIVATE configuration from CircleCI

michalinacienciala:
Go dependencies are no longer stored in private repositories,
hence gobuild image doesn't need to be built with GOPRIVATE parameter.
Configuration of this param was removed from Dockerfile in PR #688.


Shadowfiend (approved):
Love it :rocket:

PR #701, opened 2021-02-22
Appease the Pre-Commit Hook Checker

beaushinkle:
As I was working through the Liquidation Recovery stuff, we noticed that I didn't have the pre-commit hooks installed. After installing them, I ran into a lot of grief with the files I was touching with regards to pre-commit hook conformity.

Rather than trying to tackle those on a file-by-file basis, I opted to go nuclear and make sure the whole keep-ecdsa project conforms to `goimports`, `go vet` and `golint`. This is the result!


pdyraga (approved):
Three small notes. Nothing blocking, they can be addressed separately in another PR.

PR #704, opened 2021-02-25
Unchained ETH: Rename pkg/chain package to `chain`

Shadowfiend:
The pkg/chain package was named `package eth` throughout its files due
to a historical rename that wasn't fully executed. The package is now
correctly named `chain`, and all references to it have been revised to
correctly refer to it as `chain` instead of `eth`. This is in
preparation to remove from the interface the Ethereum-specific types and
make the package a true cross-chain abstraction.

An additional commit is included that adapts the pkg/chain/gen Makefiles
to reduce the number of calls to the Solidity compiler and abigen.

In draft until #686 lands on master.


pdyraga (dismissed):
Looks good to me :+1: Happy to merge as soon as #686 goes in.

lukasz-zimnoch (approved):
🚀

PR #481, opened 2020-06-11
Use package-specific version of abigen

ElderOrb:
Fixes: #439 

Running `go run github.com/ethereum/go-ethereum/cmd/abigen` should run the package-specific version of abigen and we can avoid problems with incompatible `abigen` versions this way.

PR #710, opened 2021-03-01
Multi-chain Dockerfile

lukasz-zimnoch:
Here we refactor the Dockerfile in order to support chains other than Ethereum. This is achieved by introducing a new `CHAIN` argument which is used to generate the right set of chain-specific contract bindings and is included on the build
tag list in order to build the `keep-ecdsa` client with the right host chain implementation.

This pull request also introduces several adjustments needed to make the `Dockerfile` work with Celo. The `celo-blockchain` module introduces a dependency on the `celo-bls-go` library. That library uses some native C code under the hood and links the proper target implementation according to the passed build tags. By default, the `glibc` implementation is used which doesn't work on the Alpine platform. Alpine uses the `musl` C standard library implementation so while using `keep-ecdsa` with Celo, the `musl` build tag must be explicitly passed. This is achieved by introducing the `LIBC` env variable which defines the correct C standard library implementation tag. The value of `LIBC` variable is passed to the Go compiler via `APP_BUILD_TAGS` and `ABIGEN_BUILD_TAGS` variables. As the last step of this part, several changes in tests have been made in order to run them with `celo` build tag as part of a future CI job.

By the way, this change also gets rid of the `go-ethereum`'s abigen installation. This is no longer needed due to changes introduced by #481. This also simplifies the `Dockerfile` as we don't need to switch between chain-specific abi generators.


pkg/ecdsa/tss/tss_celo_test.go
```diff
@@ -0,0 +1,11 @@
+//+build celo
+
+package tss
+
+import (
+	"github.com/celo-org/celo-blockchain/crypto"
+	"github.com/celo-org/celo-blockchain/crypto/secp256k1"
+)
+
+var S256 = secp256k1.S256
```
Shadowfiend, line 10: Could use a quick comment here to indicate why this file exists.

pkg/ecdsa/tss/tss_test.go
```diff
@@ -127,7 +127,7 @@ func TestGenerateKeyAndSign(t *testing.T) {
 
 	firstSigner := signers[groupMemberIDs[0].String()]
 	firstPublicKey := firstSigner.PublicKey()
-	curve := secp256k1.S256()
+	curve := S256()
```
Shadowfiend, line 130: Same here re: having a comment indicating where to find this function + why we're using it, I think?

Shadowfiend (commented):
Left a couple of comments. Are we expecting CI updates to be in this PR or a separate one?

> By the way, this change also gets rid of the go-ethereum's abigen installation.

Glorious 🙌

lukasz-zimnoch:
> Left a couple of comments. Are we expecting CI updates to be in this PR or a separate one?
> 
> > By the way, this change also gets rid of the go-ethereum's abigen installation.
> 
> Glorious 🙌

CI updates will be in the next PR.

pkg/ecdsa/tss/tss_celo_test.go
lukasz-zimnoch: Please see https://github.com/keep-network/keep-ecdsa/commit/382ba37582d155d0735476a65518b831f17dca52

pkg/ecdsa/tss/tss_test.go
lukasz-zimnoch: Please see https://github.com/keep-network/keep-ecdsa/commit/382ba37582d155d0735476a65518b831f17dca52

Dockerfile
```diff
@@ -1,12 +1,25 @@
 FROM golang:1.13.8-alpine3.10 AS gobuild
 
+# CHAIN argument defines the chain implementation which should be used
+# during image build process.
+ARG CHAIN=ethereum
```
pdyraga, line 5: Given that the client will soon talk to Bitcoin as well, I'd consider renaming `CHAIN` to `HOST_CHAIN` to be a little more specific. Especially that we mention the "host chain" later in this file in the comments.

pkg/ecdsa/tss/tss_ethereum_test.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/ethereum/go-ethereum/crypto/secp256k1"
+)
+
+// Components defined in this file act as a facade in front of components
+// provided by the `go-ethereum/crypto` package. This way, the client code
+// does not have to import `go-ethereum/crypto` directly. As result, the
+// backing package can be swapped to an other `go-ethereum/crypto`-like package
+// at build time, via according build tags. That solution aims to address
+// problems with package conflicts as package `go-ethereum/crypto` uses some
+// native C code underneath. If an other `go-ethereum/crypto`-like package
```
pdyraga, line 16: s/an other/another?

pkg/chain/gen/celo/Makefile
```diff
@@ -16,6 +16,10 @@ contract_files := $(addprefix contract/,$(addsuffix .go,$(subst ImplV1,,$(clean_
 abi_files := $(addprefix abi/,$(addsuffix .abi,$(clean_contract_stems)))
 abigen_files := $(addprefix abi/,$(addsuffix .go,$(clean_contract_stems)))
 
+# Additional build tags which should be passed while running `abigen` command.
+ABIGEN_BUILD_TAGS ?= default
```
pdyraga, line 20: What is `default`, where does it come from?

pdyraga (commented):
I was able to build locally ethereum and celo images. Three notes and I am good to go.

pkg/chain/gen/celo/Makefile
lukasz-zimnoch: It's just an arbitrary value indicating the default value of `ABIGEN_BUILD_TAGS` variable is in use. I felt this was more verbose than just empty quotes.

pdyraga: And if we drop this line entirely and `ABIGEN_BUILD_TAGS` are not set, what is going to happen?

lukasz-zimnoch: `go run` will fail as it will run with an empty build tag list.

pdyraga: I see. Can we add a comment explaining what `default` is and why is it needed?

Dockerfile
lukasz-zimnoch: Please see https://github.com/keep-network/keep-ecdsa/commit/c1fec15c6b018f1049e970b6937a6f92387d1edc

pkg/chain/gen/celo/Makefile
lukasz-zimnoch: Sure, please see https://github.com/keep-network/keep-ecdsa/commit/0f23d1f7f7937505a064915e37bef215a583f6cf

pdyraga (approved):
Looks good to me. Do you want to take a look @Shadowfiend?

pdyraga (approved):

PR #711, opened 2021-03-01
Liquidation Recovery Helpers

beaushinkle:
This PR implements the functions detailed in https://github.com/keep-network/keep-ecdsa/blob/master/docs/rfc/rfc-2.adoc with the only changes being to appease the compiler / linter / vetter.

Tagging @Shadowfiend for review!


go.mod
```diff
@@ -14,6 +14,7 @@ require (
 	github.com/BurntSushi/toml v0.3.1
 	github.com/binance-chain/tss-lib v1.3.1
 	github.com/btcsuite/btcd v0.20.1-beta
+	github.com/btcsuite/btcutil v0.0.0-20190425235716-9e5f4b9a998d
```
beaushinkle, line 17: I see that we have:
```
replace (
	github.com/BurntSushi/toml => github.com/keep-network/toml v0.3.0
	github.com/blockcypher/gobcy => github.com/keep-network/gobcy v1.3.1
	github.com/btcsuite/btcd => github.com/keep-network/btcd v0.0.0-20190427004231-96897255fd17
	github.com/btcsuite/btcutil => github.com/keep-network/btcutil v0.0.0-20190425235716-9e5f4b9a998d
```

Does that mean I don't need the above line? It looks like my environment generated this change

Shadowfiend:
Are these lifted directly from the RFC? If so, would be useful to mention that in the commit message. Also, a signed recovery transaction of what? The commit description could use a touch more detail.

Did there end up being any changes from RFC to here? Those are probably a good starting point, but want to flag that we shouldn't be afraid to recut or rename the functions once we put the glue code in to fit into the code more clearly.

beaushinkle:
Okay - reworked the commit history to show all the tweaks from the RFC!

Shadowfiend:
Fwiw I was comfortable with the single commit, just wanted it to call out that it was pulling from the RFC + making adjustments for compilation purposes, clarity, etc. This is also fine though :)

Will do a pass on this in current form today.

go.mod
Shadowfiend: I… Don't know, but it feels right to have this in there.

Shadowfiend (approved):
Since this is a direct lift I do want to ask @nkuba to have a second look now that it's in full-on Go-land.

pkg/extensions/tbtc/liquidation_recovery.go
```diff
@@ -0,0 +1,282 @@
...
+		0, 0, 0,
+		0, // one more dummy byte for the SigHashType
+	}
+	// The compressed public key requires 33 bytes.
+	dummyCompressedPublicKeyForWitness := []byte{
+		0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+		0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+		0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+		0, 0, 0,
+	}
```
nkuba, line 198: Would using `bytes.Repeat([]byte{0}, 74)` be cleaner?

```suggestion
	dummySignatureForWitness := bytes.Repeat([]byte{0}, 74) // one more dummy byte for the SigHashType
	
	// The compressed public key requires 33 bytes.
	dummyCompressedPublicKeyForWitness := bytes.Repeat([]byte{0}, 33)
```

nkuba (commented):
The code looks good to me. I left some minor comments inline. The only major problem I see is that we're missing unit tests. Do we plan to add tests later in a follow-up PR?

pkg/extensions/tbtc/liquidation_recovery.go
beaushinkle: I like it way more :100:

pkg/extensions/tbtc/liquidation_recovery.go
```diff
@@ -0,0 +1,282 @@
...
+	pkhAddress := pubKeyAddress.AddressPubKeyHash()
+
+	script, err := txscript.PayToAddrScript(pkhAddress)
+	if err != nil {
+		return nil, fmt.Errorf(
+			"error deriving p2wpkh scriptCode from public key: [%s]",
+			err,
+		)
+	}
+	if len(script) > 255 {
```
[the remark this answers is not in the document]
Shadowfiend: This is a great point. This check was to avoid an overflow of the `byte` cast below, but since we're specifically aiming for 0x19 as the first byte, checking that `len(script) == 25` and adjusting the error to `unexpected scriptCode length` makes sense to me.

pkg/extensions/tbtc/liquidation_recovery.go
```diff
@@ -0,0 +1,269 @@
...
+		return nil, fmt.Errorf(
+			"error extracting outpoint transaction hash: [%s]",
+			previousOutputTransactionHash,
+		)
+	}
+
+	// The witness signature field is the DER signature followed by the hash type.
+	// We write a dummy signature with 73 0 bytes. DER signatures vary in encoding
+	// between 71, 72, and 73 bytes, so we choose the longest for fee purposes.
+	dummySignatureForWitness := bytes.Repeat([]byte{0}, 74) // one more dummy byte for the SigHashType
```
Shadowfiend, line 182: With this change, I think we can move the “one more dummy byte” bit into the previous-line description, e.g. as an additional “We add one more dummy byte for the SigHashType, which is undecided for this unsigned transaction.” or similar.

Shadowfiend: Also RIP my beautiful grid of 0s 🪦

nkuba: I really appreciated you formed it into 10 columns. Made counting soooo much easier!

beaushinkle: My favorite detail was that there was 7 rows of 10, a row of 3, and then the extra lil guy all by themself

beaushinkle:
In the long term - I don't think we merge this integration branch until there's unit tests, and I'm working on unit tests until they're done, so merging this early doesn't unblock me or anything. In that light, putting them in this PR or making a new PR just comes down to github PR bookkeeping imo. Which do ya'll prefer?

Shadowfiend:
I feel like the right grouping is to include tests here, since it'll put the changes and their tests in the same diff for easy cross-referencing.

beaushinkle:
Okay - first three unit tests are ready for review. Figured I'd get them out there before working on the other specs to get some early feedback

beaushinkle:
Okay - got a test up for `BuildSignedTransactionHexString`, which means we now have happy-path tests for all of the helper functions. Any specific major failure modes ya'll want to include in this PR?

beaushinkle:
Gentle reminder that this is ready for another round of review!

pkg/extensions/tbtc/liquidation_recovery_test.go
```diff
@@ -0,0 +1,158 @@
...
+	"math/big"
+	"testing"
+
+	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/keep-network/keep-ecdsa/pkg/ecdsa"
+)
+
+func Test_PublicKeyToP2WPKHScriptCode_Works(t *testing.T) {
```
nkuba, line 18: We've got just one test  for `PublicKeyToP2WPKHScriptCode` so we can drop `_Works`.
```suggestion
func TestPublicKeyToP2WPKHScriptCode(t *testing.T) {
```

nkuba, line 18: Maybe we could also test the result of the conversion with some fixed input and outputs, e.g. based on the ones from [BIP-143](https://github.com/bitcoin/bips/blob/master/bip-0143.mediawiki)?

pkg/extensions/tbtc/derive_address_test.go
```diff
@@ -0,0 +1,98 @@
+package tbtc
```
nkuba, line 1: As the function we're testing is located in `liquidation_recovery.go` what do you think about naming this test file `liquidation_recovery_derive_address_test.go`? Or slicing `liquidation_recovery.go` to separate files? Maybe even a separate package for all liquidation recovery stuff?

pkg/extensions/tbtc/liquidation_recovery_test.go
```diff
@@ -0,0 +1,158 @@
...
+		t.Errorf("%v", err)
+	}
+
+	if len(scriptCodeBytes) != 26 {
+		t.Errorf("The script code must be exactly 26 bytes long. Instead, it was %v", len(scriptCodeBytes))
+	}
+}
+
+func Test_ConstructUnsignedTransaction(t *testing.T) {
+	recoveryAddresses := []string{
```
nkuba, line 33: The parameter of the function is named `recipientAddresses`. Can we replace `recoveryAddresses` to the same name?

pkg/extensions/tbtc/liquidation_recovery_test.go
```diff
@@ -0,0 +1,158 @@
...
+		"bcrt1q5sz7jly79m76a5e8py6kv402q07p725vm4s0zl",
+		"bcrt1qlxt5a04pefwkl90mna2sn79nu7asq3excx60h0",
+		"bcrt1qjhpgmmhaxfwj6t7zf3dvs2fhdhx02g8qn3xwsf",
+	}
+	transactionValue := int64(100000000)
```
nkuba, line 38: `transactionValue` -> `previousOutputValue`

pkg/extensions/tbtc/liquidation_recovery_test.go
```diff
@@ -0,0 +1,158 @@
...
+	"github.com/keep-network/keep-ecdsa/pkg/ecdsa"
+)
+
+func Test_PublicKeyToP2WPKHScriptCode_Works(t *testing.T) {
+	curve := elliptic.P224()
+	privateKey, _ := cecdsa.GenerateKey(curve, rand.Reader)
+	scriptCodeBytes, err := PublicKeyToP2WPKHScriptCode(&privateKey.PublicKey, &chaincfg.TestNet3Params)
+
+	if err != nil {
+		t.Errorf("%v", err)
```
nkuba, line 24: Would `t.Error(err)` work?

pkg/extensions/tbtc/liquidation_recovery_test.go
```diff
@@ -0,0 +1,158 @@
...
+	)
+	if err != nil {
+		t.Errorf("%v", err)
+	}
+	if len(msgTx.TxIn) != 1 {
+		t.Errorf("There should be 1 input transaction. Got %v instead.", len(msgTx.TxIn))
+	}
+	if len(msgTx.TxOut) != len(recoveryAddresses) {
+		t.Errorf(
+			"There should be an output transaction for each recovery address. Got %v but expected %v",
```
nkuba, line 55: Can we stick with the format with expected/actual `"test failure error message\nexpected: %v\nactual:   %v"`?

pkg/extensions/tbtc/liquidation_recovery_test.go
```diff
@@ -0,0 +1,158 @@
...
+			len(msgTx.TxOut),
+			len(recoveryAddresses),
+		)
+	}
+	share := msgTx.TxOut[0].Value
+	for _, txOut := range msgTx.TxOut {
+		feelessShare := transactionValue / int64(len(recoveryAddresses))
```
nkuba, line 62: Can we define the expected share value at the very begining of the test just below `transactionValue`?

pkg/extensions/tbtc/liquidation_recovery_test.go
```diff
@@ -0,0 +1,158 @@
...
+		if txOut.Value >= feelessShare {
```
nkuba, line 63: The value for each output will be constant across tests run, right? The transaction size won't change, so the value calculation won't change. Can we hardcode the value at the begining and do exact compare here?

pkg/extensions/tbtc/liquidation_recovery_test.go
```diff
@@ -0,0 +1,158 @@
...
+			)
+		}
+		if txOut.Value != share {
+			t.Errorf(
+				"Each output transaction should represent an equal share. Got %v but expected %v",
+				txOut.Value,
+				share,
+			)
+		}
+	}
```
nkuba, line 77: Maybe we could also serialize `msgTx` and compare with hardcoded expected string?

pkg/extensions/tbtc/liquidation_recovery_test.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+func Test_BuildSignedTransactionHexString(t *testing.T) {
+	tx := wire.NewMsgTx(wire.TxVersion)
```
nkuba, line 81: `tx` -> `unsignedTx`

pkg/extensions/tbtc/liquidation_recovery_test.go
```diff
@@ -0,0 +1,158 @@
...
+			R:          big.NewInt(int64(3)),
+			S:          big.NewInt(int64(7)),
+			RecoveryID: 1,
+		},
+		&privateKey.PublicKey,
+	)
+	if err != nil {
+		t.Errorf("Something went wrong building the signed transaction string: %v", err)
+	}
+
```
nkuba, line 114: Maybe we could use a fixed public key and also compare a result of `BuildSignedTransactionHexString` with a hardcoded expected string?

pkg/extensions/tbtc/liquidation_recovery_test.go
```diff
@@ -0,0 +1,158 @@
...
+	for _, txValue := range []int{100, 100, 100} {
+		tx.AddTxOut(wire.NewTxOut(
+			int64(txValue),
+			nil,
+		))
+	}
+	curve := elliptic.P224()
+	privateKey, _ := cecdsa.GenerateKey(curve, rand.Reader)
+	signedHexString, err := BuildSignedTransactionHexString(
```
nkuba, line 102: `signedHexString` -> `signedTxHex`
`decodedTx` -> `signedTx`

nkuba (changes requested):
I like the idea of where we're going with tests. I left some nitpicks and suggestions.

pkg/extensions/tbtc/liquidation_recovery_test.go
Shadowfiend: Tx -> Transaction :wink:

pkg/extensions/tbtc/derive_address_test.go
```diff
@@ -0,0 +1,98 @@
...
+	},
+}
+
+func Test_DeriveAddress(t *testing.T) {
+	for _, testData := range deriveAddressTestData {
+		address, err := DeriveAddress(testData.extendedAddress, testData.addressIndex)
+
+		if err != nil {
+			t.Errorf(
+				"Got %s while trying to derive %s at index %v",
```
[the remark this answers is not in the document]
beaushinkle: I wasn't getting any warnings, but I went ahead and switched to %d for all of the numbers anyway

beaushinkle:
Okay - knocked out the style/naming stuff!

beaushinkle:
Ready for another round of review!

beaushinkle:
Okay - rebased to make everything jive with master, and I think I addressed all of the review except the bit about extra testing from BIP-143, which Kuba and I talked about in flowdock

pkg/extensions/tbtc/derive_address.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	// Now to decide how we want to serialize the address...
+	var chainParams *chaincfg.Params
+	switch publicKeyDescriptor {
+	case "xpub", "ypub", "zpub":
+		chainParams = &chaincfg.MainNetParams
+	case "tpub", "upub", "vpub":
+		chainParams = &chaincfg.TestNet3Params
+	}
```
nkuba, line 75: Should we throw an error on a not supported `publicKeyDescriptor` value, e.g. `yprv`, `Ypub`, etc.?

pkg/extensions/tbtc/derive_address_test.go
```diff
@@ -0,0 +1,98 @@
...
+		"38Nfe3mepBtWicrDghCAhZitgeQotggyfF",
+	},
+}
+
+func TestDeriveAddress(t *testing.T) {
+	for _, testData := range deriveAddressTestData {
+		address, err := DeriveAddress(testData.extendedAddress, testData.addressIndex)
+
+		if err != nil {
+			t.Errorf(
```
nkuba, line 80: As we're not expecting any error in this test we could use `Fatal` and not execute further address validation.
```suggestion
			t.Fatalf(
```

pkg/extensions/tbtc/derive_address_test.go
```diff
@@ -0,0 +1,98 @@
...
+		7,
+		"3JbDbN7rYRFcBsU1BHwqrfua8TMAeMerqb",
+	},
+	// BIP141 at m/0'/0'
+	{
+		"ypub6Uy1fJC1qHtHWY6Gkavofq465y6Wi23cLGfoaQBLbTA5DjKLv7R2Qzz3ZJi6ZU8EopNGZEmHuCrmY7Dey5Fu9Jxa3XAAxZQPHhD3WLaGsin",
+		2,
+		"38Nfe3mepBtWicrDghCAhZitgeQotggyfF",
+	},
+}
```
nkuba, line 73: Let's add a test case for `BIP141: zpub`

We could also verify a negative case where a provided public key is too deep`BIP141 at m/0/0/0/0/0`.

pkg/extensions/tbtc/derive_address_test.go
```diff
@@ -0,0 +1,98 @@
...
+		"ypub6ZrwUDiiKM2b6CmJy6TSntc3b7FTE7dcUcu36jQLsXDu1uKt8jt19ZZc43vzfGs2f2r1hxWpbz8tCzJibmzwp2piYjzkhUCRzdrWU5qUoVZ",
+		0,
+		"3Aobe26f7QzKN73mvYQVbt1KLrCU1CgQpD",
+	},
+	{
+		"ypub6ZrwUDiiKM2b6CmJy6TSntc3b7FTE7dcUcu36jQLsXDu1uKt8jt19ZZc43vzfGs2f2r1hxWpbz8tCzJibmzwp2piYjzkhUCRzdrWU5qUoVZ",
+		16,
+		"3JuNnoMh8eWhtY5YLk3SMXfw7vm8y6zPLg",
+	},
+	// BIP141 at m/0
```
nkuba, line 61: Could we also add some cases to test descending?
```go
	// BIP141 at m/0/0
	{
		"ypub6UrXyh4TowUmicNsdic4E4AFtEkhLSTUnsPCYss2baT8BZit87ie8Duq1vRY64dnLupaReCGTWBpCYgvxFtTDJMdJBfytaPRG8QxoYA5ppp",
		7,
		"3JbDbN7rYRFcBsU1BHwqrfua8TMAeMerqb",
	},
	// BIP141 at m/0/0/0/0
	{
		"ypub6YyLK2V5voy6MLx3PviR4TS618Y56AuGp6gWXvRTLwQQX3oNYu8PB33DsKQcPomPEmYTUFHo1ikNrKNnsQBYehHZouzQ71b3ezywvaP3xWC",
		7,
		"3JbDbN7rYRFcBsU1BHwqrfua8TMAeMerqb",
	},
```

pkg/extensions/tbtc/derive_address_test.go
```diff
@@ -0,0 +1,98 @@
...
+	},
+}
+
+func TestDeriveAddress(t *testing.T) {
+	for _, testData := range deriveAddressTestData {
+		address, err := DeriveAddress(testData.extendedAddress, testData.addressIndex)
+
+		if err != nil {
+			t.Errorf(
+				"got %s while trying to derive %s at index %d",
```
nkuba, line 81: If we decide to use test cases map pattern we could probably simplify it to `unexpected error: %s`.

pkg/extensions/tbtc/derive_address_test.go
```diff
@@ -0,0 +1,98 @@
...
+				err,
+				testData.extendedAddress,
+				testData.addressIndex,
+			)
+		}
+
+		if address != testData.expectedAddress {
+			t.Errorf(
+				"unexpected address while trying to derive %s at index %v\nExpected: %s\nActual:   %s",
```
nkuba, line 90:
```suggestion
				"unexpected derived address\nexpected: %s\nactual:   %s",
```

nkuba (changes requested):
I got a little bit deeper into the code. Leaving the first part of the comments. I'll continue reviewing the remaining files tomorrow.

pkg/extensions/tbtc/derive_address.go
```diff
@@ -0,0 +1,105 @@
+package tbtc
```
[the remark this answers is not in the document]
beaushinkle: I think `pkg/extenstions/tbtc/recovery` will contain another higher-level helper function glues a lot of the existing helper functions together, as well as the code that manages the the latest derivation index.

I had left derive_address.go in the `tbtc` directory because it seems more specific to general btc operations than to assisting in recovery directly. I'm very indifferent, so if you have even a mild preference I'll change it right away.

pkg/extensions/tbtc/derive_address_test.go
```diff
@@ -0,0 +1,98 @@
...
+	},
+
+	{
+		"zpub6rePDVHfRP14VpYiejwepBhzu45UbvqvzE3ZMdDnNykG47mZYyGTjsuq6uzQYRakSrHyix1YTXKohag4GDZLcHcLvhSAs2MQNF8VDaZuQT9",
+		8,
+		"bc1quq0vrufxy05ypk45xmu3hpk6qsmlhr5vr3n8kz",
+	},
+	// BIP141: ypub again
+	{
+		"ypub6ZrwUDiiKM2b6CmJy6TSntc3b7FTE7dcUcu36jQLsXDu1uKt8jt19ZZc43vzfGs2f2r1hxWpbz8tCzJibmzwp2piYjzkhUCRzdrWU5qUoVZ",
```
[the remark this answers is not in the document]
beaushinkle: There's a comment on line 7 with how these were all generated:

```
// These tests use https://iancoleman.io/bip39/ with the bip39 mnemonic: loyal
// chuckle trade magnet tobacco jungle craft cram reduce climb size flip tongue
// tornado height
```

pkg/extensions/tbtc/derive_address_test.go
```diff
@@ -0,0 +1,98 @@
...
+		"xpub6Cg41S21VrxkW1WBTZJn95KNpHozP2Xc6AhG27ZcvZvH8XyNzunEqLdk9dxyXQUoy7ALWQFNn5K1me74aEMtS6pUgNDuCYTTMsJzCAk9sk1",
+		0,
+		"1MjCqoLqMZ6Ru64TTtP16XnpSdiE8Kpgcx",
+	},
+	{
+
+		"xpub6Cg41S21VrxkW1WBTZJn95KNpHozP2Xc6AhG27ZcvZvH8XyNzunEqLdk9dxyXQUoy7ALWQFNn5K1me74aEMtS6pUgNDuCYTTMsJzCAk9sk1",
+		4,
+		"1EEX8qZnTw1thadyxsueV748v3Y6tTMccc",
+	},
```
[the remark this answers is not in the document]
beaushinkle: Do you know how to get those accounts for addresses other than the BIP141 ones? the https://iancoleman.io/bip39/ tool doesn't let me change the derivation path for any of the other BIPs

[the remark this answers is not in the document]
beaushinkle: I was able to get an xpub from bip32, and can get y and zpubs from bip141. Good enough?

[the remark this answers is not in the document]
nkuba: I was thinking about this field:
image
It's available on other bips tabs too.

pkg/extensions/tbtc/derive_address.go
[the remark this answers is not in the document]
nkuba: As we plan to use it only for recovery and not as a util function across other packages, I'd lean towards placing it inside the `recovery` package.

pkg/extensions/tbtc/recovery/recovery.go
```diff
@@ -0,0 +1,174 @@
...
+			int64(0), // value is filled in after fee is computed below
+			outputScript,
+		))
+	}
+
+	// Compute weight and vsize per [BIP141], except vsize is truncated
+	// instead of rounded up, then compute the final fee and set the
+	// per-recipient value. Could result in a fractionally low fee.
+	weight := tx.SerializeSizeStripped()*3 + tx.SerializeSize()
+	vsize := weight / 4
```
nkuba, line 132: I'm wondering if it would be worth using a [`GetTxVirtualSize`function from btcd](https://github.com/btcsuite/btcd/blob/f86ae60936d7134fad9bf0ab1b3a1c261abe79b7/mempool/policy.go#L375) to calculate the virtual size? The only difference I see is that it rounds up the size, which is consistent with what was mentioned in BIP141:
> Virtual transaction size is defined as Transaction weight / 4 (rounded up to the next integer).

Is there any reason we don't want to round it up?

CC: @Shadowfiend 

```suggestion
	vsize := mempool.GetTxVirtualSize(btcutil.NewTx(tx))
```

nkuba (commented):
Next round of comments.

pkg/extensions/tbtc/recovery/recovery_test.go
```diff
@@ -0,0 +1,152 @@
...
+	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/keep-network/keep-ecdsa/pkg/ecdsa"
+)
+
+func TestPublicKeyToP2WPKHScriptCode(t *testing.T) {
+	curve := elliptic.P224()
+
+	publicKey := cecdsa.PublicKey{
```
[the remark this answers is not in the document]
beaushinkle: great idea! thanks for finding that test data :heart:

pkg/extensions/tbtc/recovery/recovery.go
```diff
@@ -0,0 +1,174 @@
...
+
+	// End goal here is a scriptCode that looks like
+	// 0x1976a914{20-byte-pubkey-hash}88ac . 0x19 should be the length of the
+	// script.
+	return append([]byte{byte(len(script))}, script...), nil
+}
+
+// ConstructUnsignedTransaction produces an unsigned transaction
+func ConstructUnsignedTransaction(
+	previousOutputTransactionHashHex string,
```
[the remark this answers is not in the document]
beaushinkle: makes total sense to me :100:

go.mod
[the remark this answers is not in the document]
beaushinkle: Okay - I tried giving this a shot by creating a new branch and updating that with master, but I don't have write access!

```
beaushinkle@Beaus-MBP ~/p/btcutil (update-upstream)> git push
ERROR: Permission to keep-network/btcutil.git denied to beaushinkle.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

Maybe @pdyraga can help here?

pkg/extensions/tbtc/recovery/recovery.go
beaushinkle: I'm going to go ahead and push the commit that uses this, since I think it's the standard route, and then we can back it out if that isn't the case

pkg/extensions/tbtc/recovery/derive_address_test.go
```diff
@@ -0,0 +1,183 @@
...
+		"3M9z86KGrNER18mKKhwxeeNxzmeoD9iUpt",
+	},
+
+	"BIP84: ypub at m/84'/0'/0'/0/0": {
+		"zpub6rePDVHfRP14VpYiejwepBhzu45UbvqvzE3ZMdDnNykG47mZYyGTjsuq6uzQYRakSrHyix1YTXKohag4GDZLcHcLvhSAs2MQNF8VDaZuQT9",
+		0,
+		"bc1q46uejlhm9vkswfcqs9plvujzzmqjvtfda3mra6",
+	},
+	"BIP84: ypub at m/84'/0'/0'/0/8": {
+		"zpub6rePDVHfRP14VpYiejwepBhzu45UbvqvzE3ZMdDnNykG47mZYyGTjsuq6uzQYRakSrHyix1YTXKohag4GDZLcHcLvhSAs2MQNF8VDaZuQT9",
```
Shadowfiend, line 49: This and the previous one both say “ypub” in their description but use a zpub.

pkg/extensions/tbtc/recovery/derive_address_test.go
```diff
@@ -0,0 +1,183 @@
...
+		"zpub6p6mUAk2dpLVVsguhHA27Qgd8e4q394Csha9jfAJCABrFRcnSv6AYPsgmAzgR8feBR4Spu2piv4xMcneocZajEvKtoHp111pizpe6aAEqfp",
+		0,
+		"bc1q9wwwcgcl2lw74quetxan4j6vhluyvyhy3dwt5l",
+	},
+	"BIP141: P2WPKH zpub from m/6'/4'": {
+		"zpub6p6mUAkAyUsTr1mTe4s3WsRWpxbWUD4grfoE4wf4EscKt2gJ7LYaHQnqRAMk6jtiNAc9qTrATeo4DCXUtjpV3gHN8AWtfcntZhsP5UeStdx",
+		1,
+		"bc1qwsuszzk93puxlgs6l6f54r66dev882pm2760dv",
+	},
+	"BIP141: P2WPKH zpub from m/6'/4'/9'": {
```
Shadowfiend, line 110: I'm unclear on how to read this description vs the early ones. The early ones seem to describe the full derivation path including the address index used for the test, these ones seem to only describe the derivation path excluding the address index. Is that right? If so, could we align them by any chance?

pkg/extensions/tbtc/recovery/derive_address_test.go
```diff
@@ -0,0 +1,183 @@
...
+	"BIP141 ypub is too deep at m/0/0/0/0/0/0": {
+		"ypub6bp11ZqNVMqm3C3eXAFGpEvKqNfEZ6Vhznd4Uo3S73RYTSFgmF7q9sWPoCFhLGVMSLqKZZpcpHoKgHNwStDuqQPnDfF13goQwS8qSFA6vnz",
+		0,
+		"extended public key is deeper than 4, depth: 5",
+	},
+}
+
+func ErrorContains(err error, expected string) bool {
+	return strings.Contains(err.Error(), expected)
+}
```
Shadowfiend, line 168: Incidentally I'm seeing just now that errors got some boosts in Go 1.13 we haven't really been making use of anywhere: https://blog.golang.org/go1.13-errors cc @keep-network/go

pkg/extensions/tbtc/recovery/recovery_test.go
```diff
@@ -0,0 +1,147 @@
...
+	if expectedSerializedMessageTransaction != serializedMessageTransaction {
+		t.Errorf(
+			"expectedSerializedMessageTransaction did not match the expectation\nexpected: %s\nactual:   %s",
+			expectedSerializedMessageTransaction,
+			serializedMessageTransaction,
+		)
+	}
+}
+
+func Test_BuildSignedTransactionHexString(t *testing.T) {
```
Shadowfiend, line 74: This _ seems out of place.

pkg/extensions/tbtc/recovery/recovery_test.go
```diff
@@ -0,0 +1,147 @@
...
+		nil,
+		nil,
+	))
+
+	for _, txValue := range []int{100, 100, 100} {
+		unsignedTransaction.AddTxOut(wire.NewTxOut(
+			int64(txValue),
+			nil,
+		))
+	}
```
Shadowfiend, line 93: Can we just use `ConstructUnsignedTransaction` here instead of manually assembling a `msg.TX`?

pkg/extensions/tbtc/recovery/recovery_test.go
```diff
@@ -0,0 +1,147 @@
...
+	if err != nil {
+		t.Errorf("Something went wrong building the signed transaction string: %v", err)
+	}
+
+	signedTransactionBytes, _ := hex.DecodeString(signedTransactionHex)
+	signedTransaction := wire.NewMsgTx(wire.TxVersion)
+	signedTransaction.BtcDecode(bytes.NewReader(signedTransactionBytes), 1, wire.WitnessEncoding)
+	serializedSignedTransactionBytes, _ := json.Marshal(signedTransaction)
+	serializedSignedTransaction := string(serializedSignedTransactionBytes)
+	expectedSerializedSignedTransaction := "{\"Version\":1,\"TxIn\":[{\"PreviousOutPoint\":{\"Hash\":[241,145,148,186,160,209,33,65,161,119,244,30,162,24,217,61,16,226,207,150,146,30,0,145,153,33,95,101,169,222,153,11],\"Index\":0},\"SignatureScript\":\"\",\"Witness\":[\"MAYCAQMCAQcB\",\"AgAAAAAH3j67ZA0rAhWQwJ1ec5WX0C2TkiTSJ6F0A2B1\"],\"Sequence\":4294967295}],\"TxOut\":[{\"Value\":100,\"PkScript\":\"\"},{\"Value\":100,\"PkScript\":\"\"},{\"Value\":100,\"PkScript\":\"\"}],\"LockTime\":0}"
```
Shadowfiend, line 121: Oh my. This is… Shorter than checking the fields one by one, but I wonder how painful this error message would be to decode if something went wrong and the status failed. I dig it as a ‘make sure nothing too unexpected happened’, but I feel like we should check TxIn (as below), then check TxOut (just the value and script, but not the witness), and finally check the JSON serialization. That means if there are issues in specific places we see them separately, and then if there's something real funky going on that affects the JSON output we can handle it separately.

Shadowfiend: Of note: I don't think that should block this PR or any further work---this test is Good Enough™ as-is. If we want to deal with it in the immediate term, we can, but we can also look at it again whenever we roll around to this again.

nkuba: Thoughts around hardcoding the unsigned input and expected signed transactions serialized as hex strings? We could then simply compare results as strings and in case of a mismatch, we can recover a struct and print a pretty diff. Example below.

```go
import "github.com/google/go-cmp/cmp"

func TestBuildSignedTransactionHexString(t *testing.T) {
	unsignedTxHex := "01000000000101f19194baa0d12141a177f41ea218d93d10e2cf96921e009199215f65a9de990b000000000000000000039003fc0100000000160014a405e97c9e2efdaed32709356655ea03fc1f2a8c9003fc0100000000160014f9974ebea1ca5d6f95fb9f5509f8b3e7bb0047269003fc010000000016001495c28deefd325d2d2fc24c5ac829376dccf520e0024a00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000002100000000000000000000000000000000000000000000000000000000000000000000000000"
	expectedSignedTx := "01000000000101f19194baa0d12141a177f41ea218d93d10e2cf96921e009199215f65a9de990b000000000000000000039003fc0100000000160014a405e97c9e2efdaed32709356655ea03fc1f2a8c9003fc0100000000160014f9974ebea1ca5d6f95fb9f5509f8b3e7bb0047269003fc010000000016001495c28deefd325d2d2fc24c5ac829376dccf520e0020930060201030201070121020000000007de3ebb640d2b021590c09d5e739597d02d939224d227a17403607500000000"

	publicKey := &cecdsa.PublicKey{
		Curve: elliptic.P224(),
		X:     bigIntFromString(t, "828612351041249926199933036276541218289243364325366441967565889653"),
		Y:     bigIntFromString(t, "985040320797760939221216987624001720525496952574017416820319442840"),
	}

	signature := &ecdsa.Signature{
		R:          big.NewInt(int64(3)),
		S:          big.NewInt(int64(7)),
		RecoveryID: 1,
	}

	unsignedTx := decodeTransaction(t, unsignedTxHex)

	signedTxHex, err := BuildSignedTransactionHexString(
		unsignedTx,
		signature,
		publicKey,
	)
	if err != nil {
		t.Fatalf("failed to build signed transaction: %v", err)
	}

	if signedTxHex != expectedSignedTx {
		t.Errorf(
			"invalid signed transaction\n- actual\n+ expected\n%s",
			cmp.Diff(decodeTransaction(t, signedTxHex), decodeTransaction(t, expectedSignedTx)))
	}
}

func decodeTransaction(t *testing.T, txHex string) *wire.MsgTx {
	txBytes, err := hex.DecodeString(txHex)
	if err != nil {
		t.Fatalf("failed to decode transaction [%s]: [%v]", txHex, err)
	}
	tx := wire.NewMsgTx(0)
	tx.BtcDecode(bytes.NewReader(txBytes), wire.ProtocolVersion, wire.WitnessEncoding)

	return tx
}
```

Failure would look similiar to:
```
=== RUN   TestBuildSignedTransactionHexString
    recovery_test.go:142: invalid signed transaction
        - actual
        + expected
          &wire.MsgTx{
                Version: 1,
                TxIn: []*wire.TxIn{
                        &{
                                PreviousOutPoint: {Hash: {0xf1, 0x91, 0x94, 0xba, ...}},
                                SignatureScript:  {},
                                Witness: wire.TxWitness{
                                        {
                                                0x30,
        -                                       0x06,
        +                                       0x07,
                                                0x02,
                                                0x01,
                                                ... // 5 identical elements
                                        },
                                        {0x02, 0x00, 0x00, 0x00, ...},
                                },
                                Sequence: 0,
                        },
                },
                TxOut:    {&{Value: 33293200, PkScript: {0x00, 0x14, 0xa4, 0x05, ...}}, &{Value: 33293200, PkScript: {0x00, 0x14, 0xf9, 0x97, ...}}, &{Value: 33293200, PkScript: {0x00, 0x14, 0x95, 0xc2, ...}}},
                LockTime: 0,
          }
--- FAIL: TestBuildSignedTransactionHexString (0.00s)
FAIL
FAIL    github.com/keep-network/keep-ecdsa/pkg/extensions/tbtc/recovery 0.241s
FAIL
```

pkg/extensions/tbtc/recovery/recovery_test.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+	curve := elliptic.P224()
+	privateKey := cecdsa.PrivateKey{
+		PublicKey: cecdsa.PublicKey{
+			Curve: curve,
+			X:     bigIntFromString(t, "828612351041249926199933036276541218289243364325366441967565889653"),
+			Y:     bigIntFromString(t, "985040320797760939221216987624001720525496952574017416820319442840"),
+		},
+		D: bigIntFromString(t, "13944630350777130481687329643969358734978178969757056469087366016470"),
+	}
```
nkuba, line 102: We don't need the whole private key, public key should be enough.
```suggestion
	publicKey := &cecdsa.PublicKey{
		Curve: elliptic.P224(),
		X:     bigIntFromString(t, "828612351041249926199933036276541218289243364325366441967565889653"),
		Y:     bigIntFromString(t, "985040320797760939221216987624001720525496952574017416820319442840"),
	}

```

pkg/extensions/tbtc/recovery/derive_address_test.go
beaushinkle: All of the tests that say `... from ....` are cases where we infer the path: we're given `m/6'/4'/9'`, and then expand it out to `m/6'/4'/9/0`, and then take index-11, so the final path would be `m/6'/4'/9/0/11`.

Maybe something like `BIP141: P2WPKH zpub at m/6'/4'/9/0/11 inferred from m/6'/4'/9'` would be the most descriptive?

pkg/extensions/tbtc/recovery/recovery_test.go
beaushinkle: I thought about that, but decided against it because it would add a dependency on `ConstructUnsignedTransaction` in the unit test for `BuildSignedTransactionHexString`. What do you think?

beaushinkle:
Alrighty - I think I addressed everything. Good for another look!

beaushinkle:
Okay! Got this rebased and up-to-date with master and the other branches that were being merged in. Anything else ya'll want to see here?

nkuba (approved):
Let's roll!

beaushinkle:
:tada: :tada: :100: :tada:

PR #712, opened 2021-03-02
CI job for Celo client

lukasz-zimnoch:
Depends on: https://github.com/keep-network/keep-ecdsa/pull/710

Here we introduce the `build-test-publish` job for Celo. Its structure is almost identical than the existing `build-test-publish` job for Ethereum. Both jobs are now placed in separate `client-celo.yml` and `client-ethereum.yml` files. This is done in the sake of readability and allows easily grouping jobs under an appropriate namespace (`Go / Celo` and `Go / Ethereum`). 

Regarding differences, the Celo job introduces additional build args (`HOST_CHAIN=celo`) while building Docker image and runs
client tests with `celo` build tag. Also, cache names in both jobs are intentionally differentiated to completely separate them as `actions/cache` action sets the cache scope to the key and branch. Something similar refers to the intermediary `go-build-env` images. They now have chain-specific suffixes to avoid problems while pushing them to local repo or cache in the future.

There is also a problem with reporting test results. @michalinacienciala noted a known bug (https://github.com/EnricoMi/publish-unit-test-result-action/issues/12) in `EnricoMi/publish-unit-test-result-action@v1` action we use to report test results. Because of that bug, it might happen that the test result summary will be under the wrong workflow while being presented in the PR summary. We're dropping the test results publishing step until the aforementioned bug gets resolved by the action maintainer.


.github/workflows/client-celo.yml
```diff
@@ -0,0 +1,79 @@
+name: Go / Celo
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
```
nkuba, line 9: We can remove the filter and run iyt on all PRs.
```suggestion
  pull_request:
```

nkuba (changes requested):

.github/workflows/client-celo.yml
michalinacienciala, line 9: Consider triggering the workflow on pull requests on all branches. Thanks to that the workflow would trigger also on PRs from `feature_branch_A` to `feature_branch_B`.
```suggestion
  pull_request:
```

.github/workflows/client-celo.yml
```diff
@@ -0,0 +1,79 @@
...
+            gotestsum --junitfile /mnt/test-results-celo/unit-tests-celo.xml -- -tags=celo,musl ./...
+
+      - name: Publish unit test results
+        uses: EnricoMi/publish-unit-test-result-action@v1
+        if: always()
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          files:  ./test-results-celo/unit-tests-celo.xml
+          check_name: Celo Client Test Results # name under which test results will be presented in GitHub (optional)
+          comment_on_pr: false # turns off commenting on Pull Requests
```
michalinacienciala, line 67: I don't think there is a need to add `-celo` suffix to `test-results` folder and `unit-tests.xml` file. We don't populate the files anywhere outside the workflow, they 'live' within the runner, only for the duration of the job execution.

.github/workflows/client-celo.yml
```diff
@@ -0,0 +1,79 @@
...
+      - name: Create test results directory
+        run: mkdir test-results-celo
+
+      - name: Run Go tests
+        run: |
+          docker run \
+            --volume $GITHUB_WORKSPACE/test-results-celo:/mnt/test-results-celo \
+            --workdir /go/src/github.com/keep-network/keep-ecdsa \
+            go-build-env-celo \
+            gotestsum --junitfile /mnt/test-results-celo/unit-tests-celo.xml -- -tags=celo,musl ./...
```
michalinacienciala, line 58: I'm not 100% sure, but `./...` looks to me as unnecessary. We already provide the path specifying `/mnt/test-results-celo/unit-tests-celo.xml`.

.github/workflows/client-ethereum.yml
```diff
@@ -41,40 +41,40 @@ jobs:
...
       - name: Publish unit test results
         uses: EnricoMi/publish-unit-test-result-action@v1
         if: always()
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
-          files:  ./test-results/unit-tests.xml
-          check_name: Go Test Results # name under which test results will be presented in GitHub (optional)
+          files:  ./test-results-ethereum/unit-tests-ethereum.xml
+          check_name: Ethereum Client Test Results # name under which test results will be presented in GitHub (optional)
           comment_on_pr: false # turns off commenting on Pull Requests
```
michalinacienciala, line 70: I don't think there is a need to add `-ethereum` suffix to `test-results` folder and `unit-tests.xml` file. We don't populate the files anywhere outside the workflow, they 'live' within the runner, only for the duration of the job execution.

michalinacienciala (changes requested):

.github/workflows/client-celo.yml
lukasz-zimnoch: It's not set the path of test results. That tells the `gotestsum` where to find test files. This is needed because we use double dash (`--`) to set build tags so positional arguments are taken into account by the `gotestsum` command and the default test command is not used in that case. Without that no tests are invoked.

.github/workflows/client-celo.yml
lukasz-zimnoch: This a leftover from my trials to fix problems with the test reporter and I wasn't aware of https://github.com/EnricoMi/publish-unit-test-result-action/issues/12 yet. So, I tried to separate everything I can to check whether the situation can be improved. Now, I agree we can drop those chain-specific suffixes.

.github/workflows/client-ethereum.yml
lukasz-zimnoch: See https://github.com/keep-network/keep-ecdsa/pull/712#discussion_r587400670

.github/workflows/client-celo.yml
michalinacienciala: Oh, I see, thanks for the explanation

.github/workflows/client-celo.yml
lukasz-zimnoch: Good catch! Please see: https://github.com/keep-network/keep-ecdsa/commit/03aafa389975e71405a6a3441ab777ea8c9ce46d

lukasz-zimnoch: Please see: https://github.com/keep-network/keep-ecdsa/commit/03aafa389975e71405a6a3441ab777ea8c9ce46d

.github/workflows/client-celo.yml
lukasz-zimnoch: Please see https://github.com/keep-network/keep-ecdsa/commit/6266c7a884d8ab3cdd1d68c65c03b4a7df94356f

.github/workflows/client-ethereum.yml
lukasz-zimnoch: Please see https://github.com/keep-network/keep-ecdsa/commit/6266c7a884d8ab3cdd1d68c65c03b4a7df94356f

lukasz-zimnoch:
@michalinacienciala all comments addressed!

.github/workflows/client-celo.yml
michalinacienciala: Actually, I think we could clean it up even a little more. We no longer need the step that creates the test results directory. Also `--volume $GITHUB_WORKSPACE/test-results:/mnt/test-results \` and `--junitfile /mnt/test-results/unit-tests.xml` fragments are redundant.
Those lines of code were introduced to pass the test-results from docker image to step publishing the tests results

lukasz-zimnoch: Yeah, I know but I thought we want to keep them as we will return to publishing step results once the bug in the publisher is resolved or we'll switch to another publisher. Anyway, got rid of that for now: https://github.com/keep-network/keep-ecdsa/commit/4287e7f38e5d6d939bf758c2bad56e0268d8a34b

nkuba (dismissed):

.github/workflows/client-celo.yml
```diff
@@ -51,14 +50,29 @@ jobs:
...
+          # HOSTNAME/PROJECT-ID/IMAGE:TAG
+          # We don't use TAG yet, will be added at later stages of work on RFC-18.
+          tags: ${{ secrets.GCR_REGISTRY_URL }}/${{ env.GOOGLE_PROJECT_ID }}/${{ env.IMAGE_NAME }}
+          labels: revision=${{ github.sha }}
+          build-args: |
+            HOST_CHAIN=celo
+            REVISION=${{ github.sha }}
+          # VERSION= ? TODO: Configure version, sample: 1.7.6
+          push: |
+            ${{ startsWith(github.ref, 'refs/heads/releases/alfajores') }}
```
nkuba, line 78: We could simplify branch name check with: 
```
${{ startsWith(github.head_ref, 'releases/alfajores') }}
```

But it's something that I'd not block this PR with.

cc: @michalinacienciala

nkuba (approved):

nkuba:
LGTM leaving final approval to @michalinacienciala

michalinacienciala (approved):

PR #713, opened 2021-03-02
Read MaxFeePerVByte from Configuration

beaushinkle:
This PR surfaces a `MaxFeePerVByte` from the configuration file that will eventually be used in the construction of the unsigned refund transaction.

Tagging @Shadowfiend for review


beaushinkle:
I really dislike passing in these configuration details all the way in from `start.go`. Is there a cleaner way?

Shadowfiend:
Yeah, so during the cross-chain refactor my conclusion (which I believe @pdyraga ended up agreeing with) was that we should move the `TBTC` configuration under the specific chain, using the top-level configuration as a fallback.

I think what we should be passing around is the full TBTC config. We should also consider how we can move the monitoring + signing we're currently doing into the tbtc extension code at pkg/extensions/tbtc, or more tightly couple that extension and the underlying code if needed. The latter thing is a bigger move that probably needs some more thought, though.

For now, I would suggest seeing about passing the full TBTC config down. Alternatively, consider using the already-passed `clientConfig` object, which I believe should include the `.Extensions.TBTC...` data.

beaushinkle:
I actually tried that originally, but when I was tracing the client config code I found:
```
// Config contains configuration for tss protocol execution.
type Config struct {
	// Defines the look-back period to check if existing, active keeps are awaiting
	// signer generation on the client start. The client does not check keeps older
	// than the look-back value.
	AwaitingKeyGenerationLookback configtime.Duration

	// Timeout for key generation and signature calculation.
	KeyGenerationTimeout configtime.Duration
	SigningTimeout       configtime.Duration
}
```
I'm totally down to include the Extensions in there though!

Shadowfiend:
Ah yeah that makes sense, I missed that this was `client.Config` instead of the top-level one.

Shadowfiend:
@beaushinkle were you planning on switching how config is passed down in this PR, or do you want to do it in a future one?

beaushinkle:
I think it makes sense to do that refactor in a separate PR

PR #714, opened 2021-03-03
Unchained Applications: Make chain.TBTCHandle stand alone, move tBTC extension setup into client.go

Shadowfiend:
The core goal here was to make the tBTC handle a standalone interface that
does not pull in all of `chain.Handle`. In the process, we reify the concept of
a `BondedECDSAKeepApplicationHandle`, which subsumes methods related
to application status and registration. We also provide explicit access to the
tBTC application handle as `chain.BondedECDSAKeepFactory`'s
`TBTCApplicationHandle()` method. Finally, we rework the config setup to
accommodate both the fact that applications are now explicitly enumerated in
the code, and to reflect the fact that address information for them is chain-
specific and therefore should live under the chain-specific contract config.

---------

This is a pretty big set of changes, and I strongly recommend going in three
groups of commits (links go to the specific subdiffs):

- Everything [before the `chain.TBTCHandle` decoupling](https://github.com/keep-network/keep-ecdsa/pull/714/files/13335aa9c2bc33a21e61afabd1000eae53430c66..d049da7b6695b92c22bcb3577dc1dbc0c6fddfff).
- [The `chain.TBTCHandle` decoupling](https://github.com/keep-network/keep-ecdsa/pull/714/commits/de16a9f890b82c01ae086b33ff9e93438ed95485).
- Everything [after the `chain.TBTCHandle` decoupling](https://github.com/keep-network/keep-ecdsa/pull/714/files/de16a9f890b82c01ae086b33ff9e93438ed95485..6c2b4ffc89ec71db000a0b2ce1ff995297d4ad80).

This PR will remain in draft until:

- [x] #709 lands (and this PR is rebased against master).
- [x] `keep-network/tbtc` gains generated contracts for Celo. (https://github.com/keep-network/tbtc/pull/758)
- [x] ~~The `chain.TBTCHandle` decoupling commit is updated to provide Celo's
  `chain.TBTCHandle` implementation (which will largely mirror Ethereum's).~~ ->
  implemented separately in #716, merged here.


pkg/client/registration.go
```diff
@@ -117,29 +117,31 @@ func registerAsMemberCandidate(
...
 	// the registration.
-	registerAsMemberCandidateWhenEligible(parentCtx, ethereumChain, application)
+	registerAsMemberCandidateWhenEligible(parentCtx, blockCounter, application)
 }
 
 // registerAsMemberCandidateWhenEligible for each new block checks the operator's
 // eligibility to be registered as a keep member candidate for the application.
 // As soon as the operator becomes eligible, function triggers the registration.
+//
+// TODO Move this to application.RegisterAsMemberCandidateWhenEligible() <-chan error?
```
pdyraga, line 127: I'd be fine either way TBH. This file contains the logic about monitoring and updating the status, the registration itself, so `registerAsMemberCandidateWhenEligible` fits well here as well in my opinion.

pdyraga (commented):
Looks very good, waiting for more! 🙌

lukasz-zimnoch:
Celo contract bindings for tBTC are here: https://github.com/keep-network/tbtc/pull/758

pdyraga:
Also, #709 landed. 🛩️

pkg/chain/ethereum/connect.go
```diff
@@ -106,17 +108,28 @@ func Connect(
...
 	if err != nil {
-		return nil, nil, fmt.Errorf(
+		return nil, fmt.Errorf(
 			"failed to create Ethereum blockcounter: [%v]",
 			err,
 		)
 	}
 
+	tbtcSystemAddress, err := config.ContractAddress(
+		BondedECDSAKeepFactoryContractName,
```
lukasz-zimnoch, line 118: Shouldn't it be: 
```suggestion
		TBTCSystemContractName,
```
??

lukasz-zimnoch:
PR with Celo's `chain.TBTCHandle` implementation: https://github.com/keep-network/keep-ecdsa/pull/716

pkg/chain/ethereum/connect.go
Shadowfiend: Yepp, good catch 😅 I think I flubbed this when I was splitting the work into good commits.

pkg/client/registration.go
Shadowfiend: One of the reasons I was thinking about this is that `registerAsMemberCandidateWhenEligible` is what requires the block counter, which as we've discussed I'd like to think about making chain-internal concept, to be threaded through every function in this file. It also doesn't seem like a crazy thing to have the application handle be conceptually responsible for. But I'm not ready to commit to the approach :grimacing:

pdyraga: I see, good note about the block counter, I did not take it into account.

configs/config.toml.SAMPLE
```diff
@@ -51,6 +51,7 @@
 # Addresses of contracts deployed on ethereum blockchain.
 [ethereum.ContractAddresses]
   BondedECDSAKeepFactory = "0xCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC"
+  TBTCSystem = "0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF"
```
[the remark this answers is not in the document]
Shadowfiend: … Yes…

The main reason I didn't do this originally was to avoid dropping the whole tBTC-specific section in the config, but since we're talking about moving all extension config under chains, I think doing it now makes sense.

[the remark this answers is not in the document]
Shadowfiend: Done.

pkg/chain/ethereum/connect.go
Shadowfiend: Fix't.

pkg/client/client.go
```diff
@@ -233,18 +253,66 @@ func Initialize(
...
+			tbtcApplicationHandle
+	}
+	for _, applicationAddress := range sanctionedApplications {
+		handle, exists := supportedApplicationsByStringID[applicationAddress.String()]
+		if exists {
+			go checkStatusAndRegisterForApplication(ctx, blockCounter, handle)
+		} else {
+			logger.Warnf(
+				"unknown sanctioned application address [%s]",
+				applicationAddress,
```
lukasz-zimnoch, line 276: That prints a byte mess in logs. Can we call `.String()` ? 

```suggestion
				applicationAddress.String(),
```

pkg/extensions/tbtc/tbtc.go
```diff
@@ -934,31 +938,31 @@ func (t *tbtc) waitDepositStateChangeConfirmation(
...
+	keep chain.BondedECDSAKeepHandle,
 ) bool {
-	currentBlock, err := t.chain.BlockCounter().CurrentBlock()
+	currentBlock, err := t.blockCounter.CurrentBlock()
 	if err != nil {
 		logger.Errorf(
 			"could not get current block while confirming "+
 				"keep [%v] is not active: [%v]",
-			keepAddress,
+			keep.ID(),
```
lukasz-zimnoch, line 948: Same problem in logs.

```suggestion
			keep.ID().String(),
```

pkg/extensions/tbtc/tbtc.go
```diff
@@ -934,31 +938,31 @@ func (t *tbtc) waitDepositStateChangeConfirmation(
...
 		func() (bool, error) {
-			return t.chain.IsActive(common.HexToAddress(keepAddress))
+			return keep.IsActive()
 		},
 	)
 	if err != nil {
 		logger.Errorf(
 			"could not confirm if keep [%v] is not active: [%v]",
-			keepAddress,
+			keep.ID(),
```
lukasz-zimnoch, line 965: And here

```suggestion
			keep.ID().String(),
```

pkg/client/client.go
```diff
@@ -233,18 +253,66 @@ func Initialize(
...
+	tbtcApplicationHandle, err := hostChain.TBTCApplicationHandle()
+	if err != nil {
+		logger.Errorf(
+			"failed to look up on-chain tBTC application information for chain " +
+				"[%s]; tBTC application functionality will be disabled",
+		)
+	} else {
+		supportedApplicationsByStringID[tbtcApplicationHandle.ID().String()] =
+			tbtcApplicationHandle
+	}
```
lukasz-zimnoch, line 268: In case the `TBTCSystem` address is not set, `hostChain.TBTCApplicationHandle()` returns an error. If so, the `supportedApplicationsByStringID` map doesn't contain the handle and  `checkStatusAndRegisterForApplication` behavior is not triggered at all. Is this intentional behavior? My understanding is that if `TBTCSystem` is not set, only tBTC extensions don't work, not the entire client registration process.

pkg/extensions/tbtc/tbtc.go
pdyraga: This is a known problem, we discussed it in the previous PR: https://github.com/keep-network/keep-ecdsa/pull/709

We decided to go ahead and merge because we'll refactor the ID later this week anyway.

pkg/chain/ethereum/tbtc.go
```diff
@@ -8,52 +8,138 @@ import (
...
-// TBTCEthereumChain represents an Ethereum chain handle with
+// tbtcApplication represents an Ethereum chain handle with
 // TBTC-specific capabilities.
-type TBTCEthereumChain struct {
-	*ethereumChain
+type tbtcApplication struct {
+	chainHandle *ethereumChain
 
-	tbtcSystemContract *contract.TBTCSystem
+	bondedECDSAKeepFactoryContract *contract.BondedECDSAKeepFactory
```
lukasz-zimnoch, line 24: `bondedECDSAKeepFactoryContract` is not set during `tbtcApplication` initialization in `TBTCApplicationHandle` method. That causes some panics in the working app.

pkg/extensions/tbtc/tbtc.go
lukasz-zimnoch: Got it, thanks! I mark this as resolved then.

pkg/extensions/tbtc/tbtc.go
lukasz-zimnoch: Resolving as per https://github.com/keep-network/keep-ecdsa/pull/714#discussion_r586510627

pkg/client/client.go
lukasz-zimnoch: Resolving as per https://github.com/keep-network/keep-ecdsa/pull/714#discussion_r586510627

Shadowfiend: See [this conversation](https://github.com/keep-network/keep-ecdsa/pull/709#discussion_r585467417). Tldr: the next PR should be the move to chain.*ID, and should fix this without requiring explicit `.String` calls everywhere.

pkg/client/client.go
Shadowfiend: Hmmm, yes, that was my understanding as well, but you're right. We were previously leaning entirely on `SanctionedApplications` to manage the registration piece, and `Extensions.TBTC` to manage the extension. I think this approach is _strictly better_, and should deprecate `SanctionedApplications`---but it also means we need to deal with backwards-compatibility with `SanctionedApplications` being set without `Extensions.TBTC`.

Since currently-running operators should only have one entry in `SanctionedApplications` and it should match `TBTCSystem`, I'm going to suggest we adjust a couple of things:

- Check `SanctionedApplications` in addition to `Extensions.TBTC` for a fallback address.
- panic if both exist but don't match.
- Deprecate `SanctionedApplications`.

Thoughts?

lukasz-zimnoch: Yep sounds good!

Shadowfiend: Ok, made these changes.

pkg/chain/ethereum/tbtc.go
Shadowfiend: This is fixed.

Shadowfiend:
Since this was still in draft, I went ahead and rebased the two branches against master. Pulling out of draft now.

Major change that was added here is that `SanctionedApplications` is also deprecated, and only `ContractAddresses.TBTCSystem` is respected now. If `ContractAddresses.TBTCSystem` is unset, `Extensions.TBTC.TBTCSystem` is used. If `Extensions.TBTC.TBTCSystem` is also unset, the first entry is used in `SanctionedApplications` if it has only one entry. If `ContractAddresses.TBTCSystem` or `Extensions.TBTC.TBTCSystem` is set and the resulting value is not the same as the first entry in `SanctionedApplications`, the client panics with instructions to set `ContractAddresses.TBTCSystem` and drop `SanctionedApplications`.

The reasoning here is that we can only safely fall back on `SanctionedApplications` if we know that the entry in there is `TBTCSystem`. Since `SanctionedApplications` is a list, we don't know the identity of the underlying contract in a straightforward way. So, if there's any doubt that it's correct, we try to panic early if we can.

pdyraga (approved):
One question and one note and it looks good to me. Leaving it in your hands @lukasz-zimnoch.

cmd/connector_ethereum.go
```diff
@@ -29,17 +28,65 @@ func connectChain(
 		)
 	}
 
-	// FIXME tBTC stuff needs some luuuuv.
-	ethereumChain, tbtcHandle, err := ethereum.Connect(ctx, ethereumKey, &config.Ethereum, config.Extensions.TBTC.TBTCSystem)
+	// DEPRECATED: config.Ethereum.ContractAddresses is the correct container
```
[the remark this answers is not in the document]
lukasz-zimnoch: We discussed this already in 716: https://github.com/keep-network/keep-ecdsa/pull/716#discussion_r586438943

cmd/connector_ethereum.go
```diff
@@ -29,17 +28,65 @@ func connectChain(
...
+	}
+
+	// DEPRECATED: config.Ethereum.ContractAddresses is the correct container
+	// for the TBTCSystem address from now on; read SanctionedApplications and
+	// assume it has a single entry that is TBTCSystem, warn if
+	// SanctionedApplications needs to be used.
+	applicationAddresses := config.SanctionedApplications.AddressesStrings
+	if !exists && len(applicationAddresses) == 1 {
+		logger.Warn(
+			"TBTCSystem address configuration in Extensions.TBTC.TBTCSystem " +
```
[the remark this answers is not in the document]
Shadowfiend: Whomp. Thanks.

lukasz-zimnoch (approved):
No more comments on my side!

PR #715, opened 2021-03-03
Move publishing of Go client to GH Actions

michalinacienciala:
and tBTC release processes in order to reduce human involvment and make
the processes faster and less error prone. A part of this task is
migration from CircleCI jobs to GitHub Actions. This commit enhances the
GitHub Action `client.yml` workflow with step that publishes built
docker image to `keep-test` Google Container Registry.


.github/workflows/client.yml
```diff
@@ -43,7 +44,7 @@ jobs:
           target: gobuild
           tags: go-build-env
           build-args: |
-          # REVISION=${{ github.sha }} TODO: set revision only when we want to publish the image
+            REVISION=${{ github.sha }}
           # VERSION= ? TODO: Configure version, sample: 1.7.6
```
nkuba, line 48: Actually, we don't need to configure `REVISION` and `VERSION` args for `go-build-env` image. We need to move it to the `Build and publish Docker Runtime Image` step.

nkuba (changes requested):

.github/workflows/client.yml
michalinacienciala: See 9fe88148564f31778bfe592e05c9916de4347f34

michalinacienciala:
Apart from doing changes related to publishing of client, I also got rid of the step that publishes the test results (in commit fc96d56419c2a1af2ad7ca85924986dd14ec80c1). The step was sometimes causing wrong association between Test Results check and the workflows (due to a known bug: https://github.com/EnricoMi/publish-unit-test-result-action/issues/12).
After the change failing tests still will cause the job to be reported as failed.
The workflow that was run on the code without the publish action can be seen [here](https://github.com/keep-network/keep-ecdsa/actions/runs/621289880).

nkuba (approved):

PR #716, opened 2021-03-03
tBTC extensions for Celo

lukasz-zimnoch:
Depends on: #714
Depends on: https://github.com/keep-network/tbtc/pull/758

This pull request introduces the Celo's `chain.TBTCHandle` implementations thus enables the tBTC extensions for the Celo chain. The whole implementation mirrors the already existing `ethereum.tbtcApplication` type.


cmd/connector_celo.go
```diff
@@ -28,6 +28,21 @@ func connectChain(
...
+	if !exists && len(config.Extensions.TBTC.TBTCSystem) != 0 {
+		logger.Warn(
+			"TBTCSystem address configuration in Extensions.TBTC.TBTCSystem " +
+				"is DEPRECATED and will be removed. Please configure the " +
+				"TBTCSystem address alongside BondedECDSAKeep under " +
+				"Celo.ContractAddresses.",
+		)
+		config.Celo.ContractAddresses[celo.TBTCSystemContractName] =
+			config.Extensions.TBTC.TBTCSystem
+	}
```
Shadowfiend, line 44: Since we don't have any live users of the Celo integration, I don't know if we need to be concerned with backwards compatibility there. Thoughts on dropping this bit entirely?

cmd/connector_ethereum.go
```diff
@@ -35,7 +35,7 @@ func connectChain(
 	if !exists && len(config.Extensions.TBTC.TBTCSystem) != 0 {
 		logger.Warn(
 			"TBTCSystem address configuration in Extensions.TBTC.TBTCSystem " +
-				"is DEPRECATED and will be removed. PLease configure the " +
+				"is DEPRECATED and will be removed. Please configure the " +
```
Shadowfiend, line 38: PlEaSe CoNfIgUrE lolol.

cmd/connector_celo.go
lukasz-zimnoch: Sure, done in https://github.com/keep-network/keep-ecdsa/commit/27e91c2599745263ffca647ed1a25ff7782a24e9

go.mod
```diff
@@ -19,9 +19,9 @@ require (
 	github.com/gogo/protobuf v1.3.1
 	github.com/google/gofuzz v1.1.0
 	github.com/ipfs/go-log v1.0.4
-	github.com/keep-network/keep-common v1.3.1-0.20210225144425-98d03fe6e9bd
+	github.com/keep-network/keep-common v1.3.1-0.20210226151147-28633644e7e5
 	github.com/keep-network/keep-core v1.3.2-0.20210225151037-3f9f104b3d8b
-	github.com/keep-network/tbtc v1.1.1-0.20210224153824-cd94a765339f
+	github.com/keep-network/tbtc v1.1.1-0.20210303113031-adf361fd085b
```
Shadowfiend, line 24: Obviously not a blocker, but should we be tagging a discrete version that includes the Celo artifacts (e.g. v1.1.2)?

pkg/chain/celo/tbtc.go
```diff
@@ -0,0 +1,462 @@
...
+
+	"github.com/celo-org/celo-blockchain/common"
+	"github.com/keep-network/keep-common/pkg/chain/celo/celoutil"
+	"github.com/keep-network/keep-common/pkg/subscription"
+	"github.com/keep-network/keep-ecdsa/pkg/chain"
+	"github.com/keep-network/keep-ecdsa/pkg/chain/gen/celo/contract"
+	tbtcchain "github.com/keep-network/tbtc/pkg/chain/celo/gen/contract"
+)
+
+// tbtcApplication represents an Celo chain handle with
```
Shadowfiend, line 18: I forgot to update this on Ethereum as well, will adjust both to clarify a little in 714 post-merge.

Shadowfiend (approved):
A'ight let's do this thing.

go.mod
pdyraga: It is changing quite often recently :grimacing: I think we should tag against a discrete version once we are done with the largest chunks of work.

PR #722, opened 2021-03-08
Celo testnet nodes configuration

lukasz-zimnoch:
Here we introduce `keep-test` configuration for the Celo Alfajores network. A quick recap of changes:
- Moved existing k8s nodes configs from `infrastructure/kube/keep-test` to `infrastructure/kube/keep-test/ethereum` to explicitly indicate they use Ethereum as host chain.
- For the aforementioned resources, two labels `chain = ethereum` and `network = ropsten` have been added. Those labels should help with the management of resources which are now separated per chain and per chain network.
- Created `10` nodes configs inteded to work with Celo. They are placed in `infrastructure/kube/keep-test/celo` and have labels `chain = celo` and `network = alfajores`.
- Adjusted the init container script to work with Celo blockchain. This required to introduce the Celo contract kit package as plain web3 doesn't work well with Celo due to additional transaction fields Celo has.

Regarding deployment, described changes have been applied to current ethereum testnet nodes. For Celo, `5` clients has been deployed and registered for a custom application. Confirmed they work well using several smoke test runs. Then, all Celo nodes have been switched to register the freshly deployed `TBTCSystem` (`0x87ABAD43f9e31067712dd5E2460C70b666AFfcE0`) contract as their application.


nkuba:
Looks fine on early review.

nkuba (approved):

PR #723, opened 2021-03-09
Pass Around TBTC Extensions

beaushinkle:
This PR is a small refactor that saves us some headache with passing around configured values directly in favor of passing around the configuration object.

Points of contention:

* The client's `Initialize` method gets the `TBTC` object (which is the the only child) rather than the `Config` object directly. I figured that since TBTC is currently the only extension, it makes more sense to work with it directly, and refactor in the future if/when we add more (yagni).

* The new home for the `Extensions` object is in `pkg/chain/config.go`, which has a package name of `eth`. I *think* this is because there's some in-flight work around making the codebase operate on abstract chains rather than ethereum specifically. I'm not in love with that location, and am happy to put this code wherever ya'll want.

tagging @Shadowfiend or @pdyraga for review


pdyraga:
👀

pkg/chain/config.go
```diff
@@ -0,0 +1,19 @@
+package eth
```
pdyraga, line 1: It is fixed on `master`:
https://github.com/keep-network/keep-ecdsa/blob/d6bedeaac806bb03a8af050063570a9631c5a444/pkg/chain/chain.go#L4

Historically, `eth` was the only host chain and we had `btc` chain implementation for some functions that are no longer the responsibility of the client.

pkg/chain/config.go
```diff
@@ -0,0 +1,19 @@
...
+type TBTC struct {
+	TBTCSystem string
+	BTCRefunds BTCRefunds
+}
+
+// BTCRefunds stores configuration related to recovering BTC from a closed keep.
+type BTCRefunds struct {
+	BeneficiaryAddress string
+	MaxFeePerVByte     int32
+}
```
pdyraga, line 19: Why don't we leave this structure in `config/config.go`? Feels like a good place, especially that part of it is BTC specific and we [discussed](https://github.com/keep-network/keep-ecdsa/pull/693#discussion_r576261733) having a `bitcoin` namespace in the config.

beaushinkle: I tried that originally, but you end up creating circular dependencies!

`pkg/chain/config.go` needs to be able to reference a type, and if you reference `config.Extensions.TBTC`, then you have to import `config` from `chain`. That creates circles since `config` imports `chain` :(

pdyraga: Oh my. I see it now. 

I'd suggest to introduce `pkg/chain/bitcoin/config.go` with
```
type Config struct {
	BeneficiaryAddress string
	MaxFeePerVByte     int32
}
```

This will be aligned with `config.Config` where we reference `ethereum.Config`, `libp2p.Config`, and `client.Config`. This will also be aligned with the previously referenced discussion about `bitcoin` namespace in the configuration file.

I'd consider leaving 
```
type Extensions struct {
	TBTC TBTC
}
``` 
inside `config/config.go`. `TBTCSystem` address is [now stored in ethereum config](https://github.com/keep-network/keep-ecdsa/blob/d6bedeaac806bb03a8af050063570a9631c5a444/cmd/connector_ethereum.go#L34) so you may save yourself headache when merging `master` changes into the integration branch later although for now, it will not look that elegant.

beaushinkle: i like it!

beaushinkle:
good for another look!

pdyraga:
Just to figure out the right ordering, would you like to have this PR or Mergarino #727 merged first?

beaushinkle:
Let's do the Merge first and then I'll resolve conflicts everywhere else

config/config.go
```diff
@@ -82,13 +83,7 @@ type Extensions struct {
...
 type TBTC struct {
 	TBTCSystem string
-	BTCRefunds BTCRefunds
-}
-
-// BTCRefunds stores configuration related to recovering BTC from a closed keep.
-type BTCRefunds struct {
-	BeneficiaryAddress string
-	MaxFeePerVByte     int32
+	BTCRefunds bitcoin.Config
```
pdyraga, line 86: Just flagging that the entire `TBTC` struct with `BTCRefunds` inside should be 🔥 after the `master` reverse-merge, and `bitcoin.Config` should be a standalone part of the config with its own `[bitcoin]` namespace in the TOML file.

pkg/chain/config.go
```diff
@@ -0,0 +1 @@
+package eth
```
pdyraga, line 1: Can we now remove this empty file?

pdyraga (approved):

PR #724, opened 2021-03-10
Surface Funding Info

beaushinkle:
This PR surfaces the ability to go from keep address to the funding info, which includes the utxo value and utxo outpoint. This information is necessary to construct the unsigned transaction to recover a liquidated deposit.

tagging @nkuba for review!


pkg/chain/ethereum/tbtc.go
```diff
@@ -158,6 +158,46 @@ func (tec *TBTCEthereumChain) OnDepositRedeemed(
 	).OnEvent(onEvent)
 }
 
+// PastCreatedEvents returns all created events for the given deposit which
+// occurred after the provided start block. Returned events are sorted by the
+// block number in the ascending order.
+func (tec *TBTCEthereumChain) PastCreatedEvents(
```
nkuba, line 164: I think we should include `Deposit` in the name. This will also be consistent with another function we've got in this file `PastDepositRedemptionRequestedEvents`.
```suggestion
func (tec *TBTCEthereumChain) PastDepositCreatedEvents(
```

pkg/chain/ethereum/tbtc.go
```diff
@@ -158,6 +158,46 @@ func (tec *TBTCEthereumChain) OnDepositRedeemed(
 	).OnEvent(onEvent)
 }
 
+// PastCreatedEvents returns all created events for the given deposit which
+// occurred after the provided start block. Returned events are sorted by the
+// block number in the ascending order.
```
nkuba, line 163:
```suggestion
// PastDepositCreatedEvents returns all deposit created events with the given keep
// address which occurred after the provided start block. Returned events are sorted by the
// block number in the ascending order.
```

pkg/chain/ethereum/tbtc.go
```diff
@@ -374,3 +414,37 @@ func (tec *TBTCEthereumChain) getDepositContract(
 
 	return depositContract, nil
 }
+
+// DepositAddressForKeepAddress retrieves the deposit address for a particular keep address
+func (tec *TBTCEthereumChain) DepositAddressForKeepAddress(
+	keepAddress string,
+) (string, error) {
+	createdEvents, err := tec.PastCreatedEvents(0, keepAddress)
```
nkuba, line 422: An alternative way of obtaining a deposit address for a keep could be calling the keep's `getOwner()` function. When creating a keep the deposit should set itself as the owner of the keep.

Maybe this would be more efficient?

beaushinkle: That sounds *much* more efficient. How confident are we that that's always true?

beaushinkle:
whipping up some test cases for these functions as well

pkg/chain/ethereum/tbtc.go
nkuba: I'm 99,99999% sure. @Shadowfiend @pdyraga do you guys see any use case where this won't be true?

pdyraga: Depends on Ethereum deployment. If it's a self-hosted Ethereum client, a single call for targeted information is more efficient than pulling events from logs. If it's Infura, (a selfish :wink:) operator does not care because both are just one request. If it is Alchemy, they charge for compute units so a single call will be better.

beaushinkle: It sounds like using `getOwner()` is never worse and sometimes better, which means we should use it!

pkg/chain/tbtc.go
```diff
@@ -93,14 +93,46 @@ type TBTCSystem interface {
...
 	// block number in the ascending order.
 	PastDepositRedemptionRequestedEvents(
 		startBlock uint64,
 		depositAddress string,
 	) ([]*DepositRedemptionRequestedEvent, error)
+
+	// DepositAddressForKeepAddress retrieves the deposit address for a particular keep address
+	DepositAddressForKeepAddress(
+		keepAddress string,
+	) (string, error)
```
Shadowfiend, line 116: I would drop this and `PastDepositCreatedEvents` until you need them externally. That is, if `FundingInfo(depositAddress string) (*FundingInfo, error)` is the end goal here, that should be the only public method you add to the interface. You can keep the current implementations of `DepositAddressForKeepAddress` and `PastDepositCreatedEvents` in `ethereum`, `celo`, and `local` if it makes sense, but they can remain unexported.

Basically make sure the end goal you're aiming for is the only thing that introduces additional public API surface, as we're more or less committing to long-term support of that API surface.

Shadowfiend: You'll see this play out in `local`, for example, where those extra two methods bring nothing to the table---and I don't think (but may be remembering wrong!) that the Bitcoin liquidation stuff requires more than just `FundingInfo`.

beaushinkle: Yeah - I started transitioning over to exposing `GetOwner`, and the first commit in that line is to just delete `PastDepositCreatedEvents` and `DepositAddressForKeepAddress`

pkg/chain/local/local_test.go
```diff
@@ -63,7 +63,7 @@ func TestOnSignatureRequested(t *testing.T) {
 	keepAddress := common.Address([20]byte{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1})
 	digest := [32]byte{1}
 
-	keep := localChain.OpenKeep(keepAddress, []common.Address{})
+	keep := localChain.OpenKeep(keepAddress, keepAddress, []common.Address{})
```
beaushinkle, line 66: Adding in the concept that local keeps have owning addresses means that we have to supply owners in all of the tests that don't actually care. I defaulted to making them be owned by the keep ID, but I'm up for more elegant solutions

Shadowfiend: Dang that's a great solve to avoid the event hop. 

I would go with the empty address tbh.

pkg/chain/tbtc.go
```diff
@@ -95,12 +95,31 @@ type TBTCSystem interface {
...
+	FundedAt       *big.Int
+	UtxoOutpoint   []uint8
+}
+
+// CreatedEvent is an event emitted when a deposit has been created.
+type CreatedEvent struct {
+	DepositAddress string
+	KeepAddress    string
+	BlockNumber    uint64
 }
```
nkuba, line 123: Do we still need it? Are we using it anywhere else besides local chain implementation?

pkg/extensions/tbtc/tbtc_test.go
```diff
@@ -2286,6 +2286,71 @@ func TestGetSignerActionDelay(t *testing.T) {
...
+		145, 90, 101, 17, 151, 96, 39, 6, 78, 252,
+		77, 81, 1, 0, 0, 0,
+	}
+	if !bytes.Equal(fundingInfo.UtxoOutpoint, expectedUtxoOutpoint) {
+		t.Errorf(
+			"utxo outpoint does not match\nexpected: %v\nactual:   %v",
+			expectedUtxoOutpoint,
+			fundingInfo.UtxoValueBytes,
+		)
+	}
```
nkuba, line 2325: We're missing validation for `FundedAt` property.

pkg/extensions/tbtc/tbtc_test.go
```diff
@@ -2286,6 +2286,71 @@ func TestGetSignerActionDelay(t *testing.T) {
 	}
 }
 
+func TestFundingInfo(t *testing.T) {
```
nkuba, line 2289: Maybe we could compare the results with a `DeepEqual`? This should simplify the code a little bit.

```go
func TestFundingInfo(t *testing.T) {
	expectedFundingInfo := &chain.FundingInfo{
		UtxoValueBytes: [8]uint8{0, 101, 205, 29, 0, 0, 0, 0},
		FundedAt:       big.NewInt(1615172517),
		UtxoOutpoint: []byte{
			194, 124, 59, 250, 130, 147, 172, 107, 48, 59,
			159, 116, 85, 174, 35, 183, 194, 75, 136, 20,
			145, 90, 101, 17, 151, 96, 39, 6, 78, 252,
			77, 81, 1, 0, 0, 0,
		},
	}

	ctx, cancelCtx := context.WithCancel(context.Background())
	defer cancelCtx()

	tbtcChain := local.NewTBTCLocalChain(ctx)

	tbtcChain.CreateDeposit(depositAddress, local.RandomSigningGroup(3))

	fundingInfo, err := tbtcChain.FundingInfo(depositAddress)
	if err != nil {
		t.Error(err)
	}

	if !reflect.DeepEqual(expectedFundingInfo, fundingInfo) {
		t.Errorf(
			"funding info does not match\nexpected: %v\nactual:   %v",
			expectedFundingInfo,
			fundingInfo,
		)
	}
}
```

nkuba (approved):

PR #725, opened 2021-03-10
Use EIP-155 signing for Ethereum contracts

Shadowfiend:
go-ethereum 1.10.0 has switched to denying non-EIP155 signatures by default.
Although our code uses go-ethereum's own ABI binding generation, the version
currently in our dependencies never added support for EIP155 signers, and
instead always uses Homestead signers for the transactors it provides helpers
for.

This commit bumps the keep-common version to a hotfix that enables EIP-155
signing for contract bindings. It updates generated code from keep-common to
reflect these new contract bindings, and updates Ethereum client instantiation
to resolve the chain id and pass it into the contract binding constructor
functions.


go.sum
```diff
@@ -1,3 +1,4 @@
+<<<<<<< Updated upstream
```
pdyraga, line 1: Merge conflict?

pdyraga: Dropped in https://github.com/keep-network/keep-ecdsa/pull/725/commits/1baa1eb72194531326ab2660aab344061a472893

pdyraga:
Tested:
- client without a fix can not submit a TX to geth v1.10.1 with default start options
- client with a fix can submit a TX to geth v1.10.1 with default start options
- client with a fix can submit a TX to geth v1.10.1 with `--rpc.allow-unprotected-txs`
- client with a fix can submit a TX to (old) geth v1.9.18

pdyraga (approved):
LGTM but given that I was committing here, we need another approval.

nkuba (approved):

dimpar (approved):

PR #726, opened 2021-03-10
Backport: Use EIP-155 signing for Ethereum contracts

Shadowfiend:
go-ethereum 1.10.0 has switched to denying non-EIP155 signatures by default.
Although our code uses go-ethereum's own ABI binding generation, the version
currently in our dependencies never added support for EIP155 signers, and
instead always uses Homestead signers for the transactors it provides helpers
for.

This commit bumps the keep-common version to a fixed master that enables
EIP-155 signing for contract bindings. It updates generated code from keep-common
to reflect these new contract bindings, and updates Ethereum and Celo client
instantiation to resolve the chain id and pass it into the contract binding
constructor functions.


Shadowfiend:
Notably this backport touches a lot more of the generated files than the changes that are relevant to the backport itself; however, that should be the last major rearrangement of those files as it's the first regeneration of contract files since keep-common started generating methods in stable order in keep-network/keep-common#68.

pdyraga:
We need to update tbtc bindings and run an experiment with Celo. I can do it tomorrow.

pdyraga (dismissed):
Tested with Celo, works as expected. tBTC contract bindings updated.

I am good to merge but since I was also committing here, we need another approval.

pkg/chain/gen/celo/contract/BondedECDSAKeepFactory.go
```diff
@@ -60,9 +62,25 @@ func NewBondedECDSAKeepFactory(
 		From: accountKey.Address,
 	}
 
-	transactorOptions := bind.NewKeyedTransactor(
-		accountKey.PrivateKey,
-	)
+	// FIXME Switch to bind.NewKeyedTransactorWithChainID when go-ethereum dep
+	// FIXME bumps beyond 1.9.25.
+	key := accountKey.PrivateKey
+	keyAddress := crypto.PubkeyToAddress(key.PublicKey)
```
nkuba, line 68: Are we missing `chainID != nil`  verification in the generated contracts code?

pdyraga: Yes, my fault. I did not regenerate contracts after locally updating `keep-common` dependency one commit further. It is fixed now.

nkuba (approved):

PR #727, opened 2021-03-10
Liquidation Recovery Master Mergarino

beaushinkle:
This PR brings the `liquidation-recovery-integration` branch up to date, since the frequency at which the divergences were relevant was starting to worry me.

I'm not sure what the best way to review this is, but there's only two files had any actual conflict resolution: `pkg/client/client.go` and `cmd/start.go`. As long as those look reasonable on a line-by-line basis, I think we're good


pkg/client/client.go
```diff
@@ -164,24 +174,24 @@ func Initialize(
...
 				networkProvider,
 				btcRecoveryAddress,
 				maxFeePerVByte,
 				clientConfig,
 				tssNode,
 				operatorPublicKey,
+				keep,
 				keepsRegistry,
 				eventDeduplicator,
-				keepAddress,
```
beaushinkle, line 184: Here's one of the main conflict sources: I was originally passing in a `keepAddress`, but master supplies a `keep`, so we don't need the address directly anymore

pkg/client/client.go
```diff
@@ -190,7 +200,7 @@ func Initialize(
 
 	go checkAwaitingKeyGeneration(
 		ctx,
-		ethereumChain,
+		hostChain,
```
beaushinkle, line 203: Here's another source of conflicts: master did a partial rename of `ethereumChain` to `hostChain` and a complete rename of `eth.Handle` to `chain.Handle`. I completed the rename to `hostChain` in this merge, since a bunch of the `ethereumChain` names were causing conflicts.

beaushinkle:
In order to get this to actually compile, I had to run `go gen ./...`. Do we want to commit the generated files?

cmd/start.go
```diff
@@ -161,32 +133,24 @@ func Start(c *cli.Context) error {
...
-
 	clientHandle := client.Initialize(
 		ctx,
-		operatorPublicKey,
-		ethereumChain,
+		operatorKeys.public,
+		chainHandle,
 		networkProvider,
 		persistence,
-		sanctionedApplications,
```
beaushinkle, line 178: `sanctionApplications` is the main source of conflicts in this file. I saw that master doesn't use it anymore, so I removed it here, and in `client.go`

Shadowfiend:
> In order to get this to actually compile, I had to run `go gen ./...`. Do we want to commit the generated files?

I… Ran a full `go gen` on contracts in #726 🙈  It should be stable over executions now though.

nkuba (approved):

PR #738, opened 2021-03-17
Cleanup afer moving to GH Actions

michalinacienciala:
and tBTC release processes in order to reduce human involvement and
make the processes faster and less error prone. A part of this task is
migration from CircleCI jobs to GitHub Actions.

Now that we have all CircleCI jobs moved to GH Actions (and verified),
we can implement some final changes removing obsolete CircleCI configs
and cleaning up pieces of code used for workflows testing. Here is a
list of changes applied:
- removed unnecessary comments
- cleaned `config.yml` (CircleCI jobs are no longer needed)
- removed memory monitoring step in client workflows
- modified conditions for workflow triggering and artifacts publication
  (according to the decision that we should migrate contracts and
  publish artifacts to registries after accepting PRs i.e. merging to
  master)
- removed `-wip` suffix from published images
- removed `--dry-run` flag when publishing to npm registry
- removed `circleci-migrate-contracts.sh` which is no longer used
  (`ci-provision-external-contracts.sh` is used instead)


michalinacienciala:
I wonder whether we can completely remove `.circleci\config.yml`? I'm not sure if it will result in removal of history of executed jobs for the `keep-ecdsa` in CircleCI or not (I think it would be good to not loose that). I did some googling and haven't found an answer yet.

nkuba:
> I wonder whether we can completely remove `.circleci\config.yml`? I'm not sure if it will result in removal of history of executed jobs for the `keep-ecdsa` in CircleCI or not (I think it would be good to not loose that). I did some googling and haven't found an answer yet.

Let's remove the file.

nkuba (approved):

PR #739, opened 2021-03-17
Backport: Cache more on-chain data between firewall checks

pdyraga:
See https://github.com/keep-network/keep-ecdsa/pull/737

There are ~8 nodes in the network right now which undelegated their stake. Given that they do not have a minimum stake now, we execute an additional check to see if the node is a member of at least one active keep.

All those nodes try to connect to bootstrap at the same time and bootstrap is executing the additional check for all of them at the same time.

This check is expensive and it may happen that it can not complete in less than 30 seconds (concurrency limits, compute unit per second limit on Ethereum client) and libp2p will retry. Libp2p retries firewall check, in this case, every 30 seconds and we may end up stacking firewall checks for the same address - the previous check can still be in progress while a new check for the same address will be executing.

Increasing or completely removing concurrency limit and/or switching to Ethereum client not limiting compute units per second helps but we need to make sure the firewall is as least expensive as possible.

With this change, we add caching to [keep index, keep on-chain ID] mapping that never changes on the chain. This way, even if two or more firewall checks are executed at the same time, most requests will be served from the cache. It will help serve those requests faster and with fewer calls to Ethereum client.

For the same reason, we add caching to positive isActive check results for a keep. So far, we were only caching negative isActive checks because they never change and information in the cache can last forever. Positive isActive check result can only be time-cached for a certain duration and there is a risk that information in the cache is out of date. Still, given that we already cache the final result of active/non-active keep member check (activeKeepMembersCache, noActiveKeepMembersCache), caching positive isActive check results for keeps for the same or shorter time will not change the behavior of the firewall but will help with reducing the number of requests to Ethereum client and making the policy check faster.

Last but not least, unit tests for firewall caching have been improved to actually validate how many chain calls are executed and to see if the given piece of information is fetched from the cache.

### DIFF
--- a/pkg/firewall/firewall.go
+++ b/pkg/firewall/firewall.go
@@ -70,16 +70,18 @@ type stakeOrActiveKeepPolicy struct {
 func (soakp *stakeOrActiveKeepPolicy) Validate(
 	remotePeerPublicKey *ecdsa.PublicKey,
 ) error {
+	remotePeerNetworkPublicKey := coreKey.NetworkPublic(*remotePeerPublicKey)
+	remotePeerAddress := coreKey.NetworkPubKeyToChainAddress(
+		&remotePeerNetworkPublicKey,
+	)
+
+	logger.Debugf("validating firewall rules for [%v]", remotePeerAddress)
+
 	// Validate minimum stake policy. If the remote peer has the minimum stake,
 	// we are fine and we should let to connect.
 	if err := soakp.minimumStakePolicy.Validate(remotePeerPublicKey); err == nil {
 		return nil
 	}
-
-	remotePeerNetworkPublicKey := coreKey.NetworkPublic(*remotePeerPublicKey)
-	remotePeerAddress := coreKey.NetworkPubKeyToChainAddress(
-		&remotePeerNetworkPublicKey,
-	)
 
 	// Check if the remote peer has authorization on the factory.
 	// The authorization cannot be revoked.
@@ -100,6 +102,11 @@ func (soakp *stakeOrActiveKeepPolicy) Validate(
 func (soakp *stakeOrActiveKeepPolicy) validateAuthorization(
 	remotePeerAddress string,
 ) error {
+	logger.Debugf(
+		"validating authorization for [%v]",
+		remotePeerAddress,
+	)
+
 	// Before hitting ETH client, consult the in-memory time cache.
 	// If the caching time for the given entry elapsed or if that entry is
 	// not in the cache, we'll have to consult the chain and execute a call
@@ -122,7 +129,7 @@ func (soakp *stakeOrActiveKeepPolicy) validateAuthorization(
 	)
 	if err != nil {
 		return fmt.Errorf(
-			"could not check authorization for address [%v]: [%v]",
+			"could not validate authorization for address [%v]: [%v]",
 			remotePeerAddress,
 			err,
 		)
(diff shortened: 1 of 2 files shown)


nkuba (approved):

pkg/firewall/firewall.go
```diff
@@ -28,10 +28,10 @@ const (
 	// We use the cache to minimize calls to Ethereum client.
 	authorizationCachePeriod = 24 * time.Hour
 
-	// activeKeepCachePeriod is the time the cache maintains information
+	// activeKeepMemberCachePeriod is the time the cache maintains information
 	// about active and no active keep members. We use the cache to minimize
 	// calls to Ethereum client.
-	activeKeepCachePeriod = 168 * time.Hour // one week
+	activeKeepMemberCachePeriod = 168 * time.Hour // one week
```
Shadowfiend, line 34: Doesn't the use of this for `keepInfoCache` mean the original name was correct?

pkg/firewall/firewall.go
```diff
@@ -51,9 +51,9 @@ func NewStakeOrActiveKeepPolicy(
 		minimumStakePolicy:          coreFirewall.MinimumStakePolicy(stakeMonitor),
 		authorizedOperatorsCache:    cache.NewTimeCache(authorizationCachePeriod),
 		nonAuthorizedOperatorsCache: cache.NewTimeCache(authorizationCachePeriod),
-		activeKeepMembersCache:      cache.NewTimeCache(activeKeepCachePeriod),
-		noActiveKeepMembersCache:    cache.NewTimeCache(activeKeepCachePeriod),
-		keepInfoCache:               newKeepInfoCache(),
+		activeKeepMembersCache:      cache.NewTimeCache(activeKeepMemberCachePeriod),
+		noActiveKeepMembersCache:    cache.NewTimeCache(activeKeepMemberCachePeriod),
+		keepInfoCache:               newKeepInfoCache(activeKeepMemberCachePeriod),
```
Shadowfiend, line 56: I just had a small epiphany: I don't think this caching should be happening in the firewall package. I think it should be happening in the chain implementation.

pkg/firewall/firewall.go
Shadowfiend: It's also not clear to me that we should be using the same lifetime for active and inactive caching.

pkg/firewall/firewall.go
pdyraga: I had the same thought at some point but I landed with caching here, given that not all the information we cache is invariant and it may depend on the particular implementation whether they are fine with some information being out of date or not.

For example, we can cache in the chain implementation information that a keep is inactive. We can also cache keep members - this information never changes. On the other hand, we should not be caching information whether keep is active as this may change at any moment. Firewall will be fine with having this information out of date for some time, but tBTC chain extensions, for instance, may want to have it always up-to-date.

pkg/firewall/firewall.go
pdyraga: Yes... After looking at it again I think they both are correct but the original one was shorter so probably better, heh.

pdyraga: > It's also not clear to me that we should be using the same lifetime for active and inactive caching.

They can have different values. I think we historically decided for one because it fits well in both cases. 

It is quite confusing but "no active keep member" and "inactive keep" caching are two different things.
"inactive keep" is cached forever in `keepInfoCache` and means the keep with the given ID is inactive.
"no active keep member" is cached for one week and it means that the given operator is not a member in any active keep.

It could happen that someone kept trying to connect all the time to our node without a minimum stake (with a modified client) and one week cache lifetime for that address would let us block those attempts early. With a shorter caching lifetime, we'd have to execute the loop of checks more often.

In the case of a node who is a member in at least one active keep, one week cache lifetime felt like a good balance between too frequent checks and keeping this information relatively up to date.

pkg/firewall/firewall.go
Shadowfiend: Hmmm… I see some of this yeah.  I guess a better way to put it might be, “we might be conflating things that we want the firewall to cache and things that we want the chain to cache”. For example, keep active status can probably be cached at the chain level since it's immutable once settled.

But… Not something to dig into too much right now I think.

PR #740, opened 2021-03-17
Guard execution of on-chain firewall calls with a mutex

pdyraga:
This PR is a continuation of firewall fixes started in https://github.com/keep-network/keep-ecdsa/pull/737.

Keeping as a draft until properly tested.

This change is effectively reducing firewall on-chain concurrency to one
call but this is exactly what we need. The loop going through keeps is
always executing checks in a sequence. When two on-chain calls were
executed at the same time, they were always a duplicate of the same
call. Mutex should not slow anything down and should in fact speed up
the loop. Instead of executing the same check from 8 concurrent threads
and being rate-limited, we will execute 8 different checks at more or
less the same time.

### DIFF
--- a/pkg/firewall/firewall.go
+++ b/pkg/firewall/firewall.go
@@ -70,14 +70,16 @@ type stakeOrActiveKeepPolicy struct {
 func (soakp *stakeOrActiveKeepPolicy) Validate(
 	remotePeerPublicKey *ecdsa.PublicKey,
 ) error {
+	remotePeerNetworkPublicKey := coreKey.NetworkPublic(*remotePeerPublicKey)
+	remotePeerAddress := coreKey.NetworkPubKeyToEthAddress(&remotePeerNetworkPublicKey)
+
+	logger.Debugf("validating firewall rules for [%v]", remotePeerAddress)
+
 	// Validate minimum stake policy. If the remote peer has the minimum stake,
 	// we are fine and we should let to connect.
 	if err := soakp.minimumStakePolicy.Validate(remotePeerPublicKey); err == nil {
 		return nil
 	}
-
-	remotePeerNetworkPublicKey := coreKey.NetworkPublic(*remotePeerPublicKey)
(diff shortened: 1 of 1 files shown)


pkg/firewall/firewall.go
```diff
@@ -245,14 +256,21 @@ func (soakp *stakeOrActiveKeepPolicy) getKeepAtIndex(
 		return common.HexToAddress(cachedAddress), nil
 	}
 
+	cache.mutex.Lock()
+	defer cache.mutex.Unlock()
+
+	cachedAddress, isCached = cache.address[index.String()]
+	if isCached {
+		return common.HexToAddress(cachedAddress), nil
+	}
```
nkuba, line 265: Should we have separate locks for reading?
```suggestion
	cache.mutex.RLock()
	cachedAddress, isCached = cache.address[index.String()]
	cache.mutex.RUnlock()
	if isCached {
		return common.HexToAddress(cachedAddress), nil
	}
	
	cache.mutex.Lock()
	defer cache.mutex.Unlock()
```

pkg/firewall/firewall.go
```diff
@@ -280,6 +298,21 @@ func (soakp *stakeOrActiveKeepPolicy) isKeepActive(
 		return true, nil
 	}
 
+	cache.mutex.Lock()
```
nkuba, line 301: Shouldn't we have a separate mutex for each property cached in `keepInfoCache`?

pkg/firewall/firewall.go
pdyraga: No, we shouldn't. We need to make sure the second check happens in the critical section with the fetch from the chain. This is just a variation of a double-checked locking pattern.

pkg/firewall/firewall.go
pdyraga: I don't think it will help. All those checks need to be executed sequentially so we'll not gain anything from separate mutexes IMO.

PR #741, opened 2021-03-18
Set ERC20 details in `TestToken` contract

r-czajkowski:
We use `TestToken` contract as a mock of LP token in the `LPRewards`
contract, in local/Ropsten deployment. Due to changes from
https://github.com/MetaMask/metamask-extension/pull/10606 if the user
calls `approve` on the `TestToken` the MetaMask shows the loading screen
which never disappears because `TestToken` doesn't implement
`ERC20Detailed`. This commit will prevent MetaMask from freezing.

### DIFF
--- a/solidity/contracts/test/TestToken.sol
+++ b/solidity/contracts/test/TestToken.sol
@@ -1,8 +1,12 @@
 pragma solidity 0.5.17;
 
 import "openzeppelin-solidity/contracts/token/ERC20/ERC20.sol";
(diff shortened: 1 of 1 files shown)


michalsmiarowski (dismissed):
Looking good 👍

nkuba (dismissed):

nkuba:
LGTM, @r-czajkowski you just need to sort out the linting.

michalsmiarowski (approved):

PR #742, opened 2021-03-18
Backport: Guard execution of on-chain firewall calls with a mutex

pdyraga:
Backport of https://github.com/keep-network/keep-ecdsa/pull/740

This change is effectively reducing firewall on-chain concurrency to one
call but this is exactly what we need. The loop going through keeps is
always executing checks in a sequence. When two on-chain calls were
executed at the same time, they were always a duplicate of the same
call. Mutex should not slow anything down and should in fact speed up
the loop. Instead of executing the same check from 8 concurrent threads
and being rate-limited, we will execute 8 different checks in more or
less the same time.


nkuba:
@pdyraga aren't we missing debug loggers from the commit https://github.com/keep-network/keep-ecdsa/pull/740/commits/2e8910d41e0465fd49a71c9b125b97b247be506f?

pdyraga:
> @pdyraga aren't we missing debug loggers from the commit [2e8910d](https://github.com/keep-network/keep-ecdsa/commit/2e8910d41e0465fd49a71c9b125b97b247be506f)?

Please see https://github.com/keep-network/keep-ecdsa/pull/739/commits/9236894e3d47b215b2271c5c37a24c2eeb69f2d3 from https://github.com/keep-network/keep-ecdsa/pull/739.

They are here:
https://github.com/keep-network/keep-ecdsa/blob/24555b81812cc3638157207bc0c8c59a923e0ce6/pkg/firewall/firewall.go#L78

https://github.com/keep-network/keep-ecdsa/blob/24555b81812cc3638157207bc0c8c59a923e0ce6/pkg/firewall/firewall.go#L105-L108

https://github.com/keep-network/keep-ecdsa/blob/24555b81812cc3638157207bc0c8c59a923e0ce6/pkg/firewall/firewall.go#L150-L153

https://github.com/keep-network/keep-ecdsa/blob/24555b81812cc3638157207bc0c8c59a923e0ce6/pkg/firewall/firewall.go#L314-L317

https://github.com/keep-network/keep-ecdsa/blob/24555b81812cc3638157207bc0c8c59a923e0ce6/pkg/firewall/firewall.go#L355-L358

nkuba (approved):

PR #744, opened 2021-03-18
Unchained IDs: Implement chain.ID

Shadowfiend:
This is a very large commit, as it touches a tremendous amount of the
codebase. Here is the set of additional changes that had to be made to
make everything fit together as expected:

- chain.ID is a new interface for generic IDs on a given chain. IDs from
  different chain handles are not cross-compatible unless the underlying
  chain is the same.
- chain.OfflineHandle is a new type of chain handle that exposes access
  to certain information about a partially-initialized chain
  representation that is not connected to the underlying chain. This is
  provided in order to expose the UnmarshalID, Name, and OperatorID
  methods in a chain-independent way for reasons explained subsequently.
- All chain.* interfaces that were returning or taking common.Address as
  parameters are now taking or returning chain.ID. This includes event
  fields that had common.Address or slices thereof as their type.
- chain.BondedECDSAKeepCreatedEvent now carries a field that directly
  reports whether the chain's operator is a member of the created keep.
  This was done to avoid excessively relying on inter-chain.ID
  comparison where avoidable.
- Each chain implementation has an ids.go file that implements the shims
  for the chain's internal ID type to be treated as a chain.ID. These
  shims generally also include conversion functions in the opposite
  direction.
- Where relevant, exported functions that are not externally used are
  no longer exported.
- The firewall, event deduplicator, etc all interface with chain.ID now.
- Several cases where strings or []bytes were being used to pass around
  keep and member addresses are now using chain.ID, such as in
  AnnounceProtocol. To this end, all chains provide a
  `PublicKeyToOperatorID` method that can be used to go from a
  crypto/ecdsa.PublicKey to a chain.ID.
- Several cases where string representations of addresses were being
  used now immediately go to/from chain.ID using a chain-provided
  `UnmarshalID` function. Most notably, this includes the keep registry.
- celo.Offline and ethereum.Offline provide direct access to a
  chain.OfflineHandle, and this is exposed in the cmd package via a
  conditionally compiled `offlineChain` call that lives alongside
  `connectChain`.
- A new `buildPersistenceHandle` helper in the `cmd` package takes care
  of using the chain name to resolve keep registry information under a
  subdirectory for the particular chain. The Ethereum chain is
  special-cased as living at the top level for backwards compatibility.
  This function also validates the chain handle and fails early if the
  handle does not conform to the right requirements, in particular some
  strict requirements around allowed characters ([a-z][A-Za-z0-9-_]*).
- `cmd/signing.go` now works cross-chain thanks to the `offlineChain`
  and `buildPersistenceHandle` helpers.

## Testing

One note for testing: anyplace where `chain.ID` is used as a map key is
a potential place where things won't work as expected. Half the map keys
were strings and half were `common.Address`, and I have not had a chance
to verify that all the `common.Address`->`chain.ID` conversions worked
as expected.


cmd/start.go
```diff
@@ -127,7 +128,29 @@ func Start(c *cli.Context) error {
...
+			// TODO allowlist vs sanitization? [A-Za-z0-9-_.] and replace all
+			//      else with -?
+			// TODO max length
+			strings.NewReplacer(
+				"/", "", // no path delimiters please
+				" ", "-", // spaces become dashes
+				"0x", "", // no 0x to avoid address clashes
+			).Replace(
+				strings.ToLower(chainHandle.Name()),
+			)
```
pdyraga, line 151: Should we extract this part to some utility function in... ... Well, I do not know where would that fit into, heh. We are clearly reformatting some chain-specific string into something fitting our expected directory layout.

pkg/chain/celo/ids.go
```diff
@@ -0,0 +1,64 @@
...
+func (cc *celoChain) PublicKeyToOperatorID(publicKey *cecdsa.PublicKey) chain.ID {
+	return celoChainID(crypto.PubkeyToAddress(*publicKey))
+}
+
+// celoChainID is the local chain.ID type; it is an alias for
+// go-ethereum/common.Address.
+type celoChainID common.Address
+
+func (ci celoChainID) ChainName() string {
+	return "local"
```
pdyraga, line 35: I am confused, why `"local"` here?

Shadowfiend: … Listen, just 'cause I'm copying and pasting doesn't mean you get to call me out on it :laughing:

I was hoping to find a good way to introduce some sort of common `ids` package but sadly I had to resort to copy/paste. Will fix.

cmd/start.go
Shadowfiend: I'm going to rework `chainHandle.Name()` to require this behavior in the first place in the docs, and then going to impose a failing check for it here I think.

Shadowfiend: This is redone now, once I fix the commits it'll make a little more sense hehe.

Shadowfiend:
… Ok so I wasn't able to break this commit down. It's just a giant pile of changes, but the commit has a detailed message on what went down.

Caches and similar bits should get particular attention in the testing, as per the note in the PR description.

pkg/chain/local/local.go
```diff
@@ -26,6 +26,8 @@ import (
 type Chain interface {
 	chain.Handle
 
+	OperatorAddress() common.Address
+
```
pdyraga, line 30: Should we refactor functions below to accept `chain.ID` instead of `common.Address?

pkg/firewall/firewall.go
```diff
@@ -70,41 +68,38 @@ type stakeOrActiveKeepPolicy struct {
 func (soakp *stakeOrActiveKeepPolicy) Validate(
 	remotePeerPublicKey *ecdsa.PublicKey,
 ) error {
-	remotePeerNetworkPublicKey := coreKey.NetworkPublic(*remotePeerPublicKey)
-	remotePeerAddress := coreKey.NetworkPubKeyToChainAddress(
-		&remotePeerNetworkPublicKey,
-	)
+	remotePeerOperatorID := soakp.chain.PublicKeyToOperatorID(remotePeerPublicKey)
```
pdyraga, line 71: Yesssss! Removing this insane `coreKey.NetworkPublic` -> `coreKey.NetworkPubKeyToChainAddress` conversion warms my heart. :fire: ❤️

pkg/firewall/firewall.go
```diff
@@ -70,41 +68,38 @@ type stakeOrActiveKeepPolicy struct {
...
 	// we are fine and we should let to connect.
 	if err := soakp.minimumStakePolicy.Validate(remotePeerPublicKey); err == nil {
 		return nil
 	}
 
-	// Check if the remote peer has authorization on the factory.
-	// The authorization cannot be revoked.
 	// If peer has no authorization on the factory it means it has never
-	// participated in any group selection so there is no chance it can be
 	// a member of any keep.
```
pdyraga, line 85: Did we remove those comment lines on purpose?

pkg/registry/keeps_test.go
```diff
@@ -332,8 +350,10 @@ func (phm *persistenceHandleMock) ReadAll() (<-chan persistence.DataDescriptor,
 	outputData := make(chan persistence.DataDescriptor, 3)
 	outputErrors := make(chan error)
 
-	outputData <- &testDataDescriptor{"/membership_0", keepAddress1.String(), signerBytes1}
-	outputData <- &testDataDescriptor{"/membership_0", keepAddress2.String(), signerBytes2}
+	fmt.Println(keepID1)
+	fmt.Println(signer1)
```
pdyraga, line 354: :fire:?

pdyraga:
Found nothing blocking, time for tests.

pkg/chain/local/local.go
Shadowfiend: (Appearing in my capacity as open source contributor rather than employee..... it didn't seem worth the effort since the functions are just for local testing, rather than for external interface exposure.)

pkg/firewall/firewall.go
Shadowfiend: (Same caveat as above... this seems likely to have been a merge fumble.)

cmd/start.go
```diff
@@ -94,7 +92,7 @@ func Start(c *cli.Context) error {
 		return fmt.Errorf("error obtaining stake monitor handle: [%v]", err)
 	}
 	hasMinimumStake, err := stakeMonitor.HasMinimumStake(
-		chainHandle.Address().Hex(),
+		chainHandle.OperatorID().String(),
```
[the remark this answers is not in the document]
Shadowfiend: (And same: without poking deep, I believe the reasoning here was that stake monitor was a keep-core interface, and all of the cross chain changes here were scoped to keep-ECDSA for this pass. Pulling the appropriate bits to keep-core and keep-common would be amazing.

pkg/chain/local/ids.go
```diff
@@ -0,0 +1,61 @@
+package local
+
+import (
+	cecdsa "crypto/ecdsa"
+	"fmt"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/crypto"
```
lukasz-zimnoch, line 8: This import causes the fail of the Celo CI job. `go-ethereum/crypto` conflicts with `celo-blockchain/crypto` package used by the Celo code (when `celo` build tag is used). We need to extract `crypto.PubkeyToAddress` to separate files and use a common alias. Those files can be, for example, `crypto_ethereum.go` which will import `go-ethereum/crypto`  and `crypto_celo.go` which will take `celo-blockchain/crypto`. Both files should also use correct build tags.

lukasz-zimnoch:
We can also get rid of the `pkg/chain/celo/ducktape.go` file now.

lukasz-zimnoch (commented):
We can also get rid of the `pkg/chain/celo/ducktape.go` file now.

pkg/node/node.go
```diff
@@ -70,55 +70,45 @@ func NewNode(
...
 func createAddressFilter(
-	addresses []common.Address,
+	keepMemberIDs []chain.ID,
 ) net.BroadcastChannelFilter {
-	authorizations := make(map[string]bool, len(addresses))
-	for _, address := range addresses {
-		authorizations[hex.EncodeToString(address.Bytes())] = true
+	operatorAuthorizations := make(map[string]bool, len(keepMemberIDs))
+	for _, keepMemberID := range keepMemberIDs {
+		operatorAuthorizations[keepMemberID.String()] = true
```
lukasz-zimnoch, line 104: This change breaks the key generation process. I observed a lot of logs like that:
```
keep-ecdsa rejecting message from [...]; author is not authorized
```
And the key is not generated as expected. I think the problem is because `keepMemberID.String()` represents an address hex with the checksum (according to `common.Address`'s `Hex` method). Down the road, in line `111`, we look at the authorization using an address constructed from encoded address bytes, without the checksum. This cause the mismatch and the autorization is not granted.

Lines 108-110 should be probably replaced by:
```
authorAddress := operator.PubkeyToAddress(*authorPublicKey).Hex()
```
But this cause a leak of `chain.ID` details as we need to know what `chain.ID`'s `String` method does under the hood.

lukasz-zimnoch:
Just a note: I'll try to fix the problems I mentioned above and @pdyraga will review those fixes and do a final testing.

pkg/chain/local/ids.go
lukasz-zimnoch: Fixed in https://github.com/keep-network/keep-ecdsa/pull/744/commits/b780e8f2ea2b31184bff7e2331e3d0f1ecf71028

pkg/chain/celo/celo.go
```diff
@@ -5,22 +5,50 @@ package celo
...
 	corechain "github.com/keep-network/keep-core/pkg/chain"
 	"github.com/keep-network/keep-ecdsa/pkg/chain"
 )
 
 var logger = log.Logger("keep-chain-celo")
 
-// Address returns client's Celo address.
-func (cc *celoChain) Address() ExternalAddress {
-	return toExternalAddress(cc.accountKey.Address)
+// Offline returns a chain.Handle for an offline Ethereum client. Use Connect to
```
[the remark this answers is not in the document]
lukasz-zimnoch: Addressed in https://github.com/keep-network/keep-ecdsa/pull/744/commits/53eb92875f151f5ad4c0255959969aef3661ed48

lukasz-zimnoch:
> We can also get rid of the `pkg/chain/celo/ducktape.go` file now.

Addressed in https://github.com/keep-network/keep-ecdsa/pull/744/commits/53eb92875f151f5ad4c0255959969aef3661ed48

pkg/node/node.go
lukasz-zimnoch: Fixed in https://github.com/keep-network/keep-ecdsa/pull/744/commits/9f6a0a6654ce61209c5b5c6967832ed7e3e25828. Some minor improvements has been also made in https://github.com/keep-network/keep-ecdsa/pull/744/commits/d24d3254e239d5772803e2e30bfe6050cca5bae8

lukasz-zimnoch (approved):
Testing completed! Seems everything works well! @pdyraga please double-check my changes and merge at will.

pkg/chain/local/ids.go
pdyraga: I don't love `keep-core/pkg/operator` and the fact we need to call it hurts me but it is not blocking this PR. We will need to do something about `keep-core/pkg/operator` in the future though.

pdyraga (approved):

Shadowfiend:
excited scream gif

PR #749, opened 2021-03-24
Use env property to get TBTCTSystem address in init script

nkuba:
Adding a dependency to TBTCSystem contract to obtain its address in the init container (https://github.com/keep-network/keep-ecdsa/commit/23db2b85e47c11c9b5b5deaa97614011366bda20) gave an additional cyclic dependency to `tbtc` which already depends on `keep-ecdsa`. We used to handle this case by defining the TBTCSystem address as an environment variable on the pod (SANCTIONED_APPLICATIONS). With this PR, we go back to this solution.

Right now keep-ecdsa init container build is coupled with keep-ecdsa contracts migration. We could separate them and add tbtc migration in between but this isn't that straightforward as it would require calls to external projects: `keep-ecdsa migration -> tbtc migration -> keep-ecdsa initcontainer build`. We are planning to handle such situations soon with RFC-18. We will revisit pulling the address from a contract artifact when working on the integration of the modules.

### DIFF
--- a/infrastructure/kube/templates/keep-ecdsa/initcontainer/provision-keep-ecdsa/provision-keep-ecdsa.js
+++ b/infrastructure/kube/templates/keep-ecdsa/initcontainer/provision-keep-ecdsa/provision-keep-ecdsa.js
@@ -131,7 +131,7 @@ async function provisionKeepTecdsa() {
       !sortitionPoolContractAddress ||
       sortitionPoolContractAddress === "0x0000000000000000000000000000000000000000"
     ) {
-      throw new Error(`missing sortition pool for application: [${tbtcSystemAddress}]`)
+      throw new Error(`missing sortition pool for TBTCSystem contract: [${tbtcSystemAddress}]`)
     }
 
     console.log(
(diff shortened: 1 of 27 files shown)


infrastructure/kube/templates/keep-ecdsa/initcontainer/provision-keep-ecdsa/provision-keep-ecdsa.js
```diff
@@ -117,37 +120,29 @@ async function provisionKeepTecdsa() {
...
-      await authorizeSortitionPoolContract(
-        operatorAddress,
-        sortitionPoolContractAddress,
-        authorizer
-      )
+    if (
+      !sortitionPoolContractAddress ||
+      sortitionPoolContractAddress === "0x0000000000000000000000000000000000000000"
+    ) {
+      throw new Error(`missing sortition pool for application: [${tbtcSystemAddress}]`)
```
lukasz-zimnoch, line 134:
```suggestion
      throw new Error(`missing sortition pool for TBTCSystem contract: [${tbtcSystemAddress}]`)
```

lukasz-zimnoch (approved):

PR #751, opened 2021-03-25
Move Go scan job (gosec.yml) to Go workflow (client-ethereum.yml)

michalinacienciala:
There is no need to perform Go scan and Go building/testing/publishing
in separate Github Actions workflows. Moving `scan` job from `gosec.yml`
to `client-ethereum.yml`.


nkuba:
As we've got separate builds for ethereum and celo clients we should leave go sec workflow as is. Are you fine with closing this PR @michalinacienciala ?

michalinacienciala:
Closing PR. Celo and Etherum workflows will be merged into one in https://github.com/keep-network/keep-ecdsa/pull/823 and this is where we'll change how we handle gosec workflow.

PR #753, opened 2021-03-26
Use path-filter action in workflows to support required checks in PRs

michalinacienciala:
and tBTC release processes in order to reduce human involvement and to
make the processes faster and less error prone. We've migrated most of
CircleCI jobs to GitHub Actions. Now we want to introduce branch
protection rules on master, requiring the passing of particular jobs
before merging PR to master. Those jobs in `keep-ecdsa` are:
* `client-build-test-publish`
* `client-scan`
* `contracts-build-and-test`
* `contracts-lint`
* `staker-rewards-test`
* `staker-rewards-lint`

What's more, we want to execute workflows containing those jobs only if
there were changes in particular paths. GitHub Actions allows for
specifying those paths by using `on.<event_name>.paths` syntax.
Unfortunately, such configuration for the `pull-request` event clashes
with the functionality of protected branches.
Example: If we set branch protection rule requiring the passing of
`job_a` and `job_a` will be part of `workflow_a` using
`on.pull_request.paths = folder_a` setting, the `workflow_a` won't be
started if we create PR that changes some files outside of `folder_a`.
Hence `job_a` won't be executed. But the PR will still require `job_a`
to be passing (or skipped) and will not allow for the merge of the PR
to the protected branch.
As skipping of the required jobs doesn't block the PRs from merging, we
decided to move the checking of paths for `pull_request` events from
workflow level to job level. It's not something natively supported by
GH Actions, but there is a `dorny/paths-filter` action that allows for
doing exactly that. We've set up the action to execute only if a
workflow was started by the `pull_request` event. Action checks the
list of files changed by the PR and when any of the files matches
provided filter, it sets the output to `true`. We then use that output
as a condition for execution of rest of the jobs in the workflow. This
way, if a particular job is required for the PR, but the path
conditions are not met, the job will be skipped (allowing for PR
merge).

Here is a summary what has been configured for particular workflows:

Client (Ethereum / Celo):

* workflow dispatch:
    Everything executed
* push on `master`:
    Everything executed if there are any changes OUTSIDE of these paths:
    - `docs/**`
    - `infrastructure/**`
    - `scripts/**`
    - `staker-rewards/**`
* pull_request:
    Everything (except publishing) executed if there are any changes
    OUTSIDE of these paths:
    - `docs/**`
    - `infrastructure/**`
    - `scripts/**`
    - `staker-rewards/**`

Solidity:
* workflow dispatch:
    Everything executed
* push on `master`
    Everything executed if there are changes IN:
    - `solidity/**`
* pull_request:
    Everything (except migrating/publishing contracts and 
    building/publishing initcontainer) executed if there are changes IN:
    - `solidity/**`

Staker Rewards:
* workflow dispatch:
    Everything executed
* push on `master`:
    Everything executed if there are changes IN:
    - `staker-rewards/**`
* pull_request:
    Everything executed if there are changes IN:
    - `staker-rewards/**`


michalinacienciala:
@nkuba, before approving PR, following jobs would need to be configured as required in GitHub's settings:
* `client-build-test-publish`
* `client-scan`
* `contracts-build-and-test`
* `contracts-lint`
* `staker-rewards-test`
* `staker-rewards-lint`
I decided to add prefixes `client-`, `contracts-` and `staker-rewards-` to avoid confusion in the future as GH in protected branches settings aggregates all jobs with the same name from different workflows (which means it is not possible to make `job_a` in `workflow_1` required, but to leave `job_a` in `workflow_2` as not required).

.github/workflows/client-celo.yml
```diff
@@ -70,5 +93,16 @@ jobs:
           # VERSION= ? TODO: Configure version, sample: 1.7.6
           push: |
             ${{ github.ref == 'refs/heads/master'
-              && (github.event_name == 'push'
-              || github.event_name == 'workflow_dispatch') }}
\ No newline at end of file
+              && github.event_name != 'pull_request' }}
+
+  client-scan:
```
nkuba, line 98: Can we revert this change and leave the scan function in `gosec.yml`? I'm not sure if `client-celo.yml` is a good place for the job.

.github/workflows/gosec.yml
```diff
@@ -1,22 +0,0 @@
-name: Gosec
```
nkuba, line 1: Let's leave it as was before.

nkuba (changes requested):

.github/workflows/client-celo.yml
michalinacienciala: Sorry, it was a mixup on my side when I was manually copying/pasting code from different branch. I intended to have `gosec` in `client-ethereum` (as covered by https://github.com/keep-network/keep-ecdsa/pull/751/files), but by mistake I pasted it to the wrong file (the code I wanted to use to start a new branch was depending on two PRs and I ended up doing checkout from one and manually applying some changes from another - I should have use cherry-pick instead, or don't include the change at all and later deal with it via resolving merge conflicts).
Anyway, I restored the original config - see 46b71aedb51496efa534022a9482d060ce8f6b2a.

nkuba (approved):

nkuba (approved):

PR #756, opened 2021-03-31
Derivation Index Persistence Framework

beaushinkle:
This PR implements a file-storage based persistence layer that stores already-used indexes for hierarchical deterministic (HD) wallets. This allows us to keep generating new btc addresses to send liquidations to so that we don't leak information about users.

This implementation creates a file structure in the configured `[Storage.DataDir]` that looks like:
```
beaushinkle@Beaus-MBP ~/example-storage> tree
.
└── bitcoin
    └── derivation_indexes
        ├── xpub_1sameidx
        │   └── 777
        ├── xpub_zCAk9sk1
        │   └── 5090
        ├── ypub_6Z2Yy4ZH
        │   └── 8982
        └── zpub_VDaZuQT9
            └── 8559
```
where the `bitcoin` folder is a sibling of `current`, and `archive` from the existing storage handler.

The folder names in the `derivation_indexes` folder are truncated *pub addresses (since a user may change their configured *pub), and the 0-byte files within those folders are the highest-used index. We clean up old files after successfully writing a higher index to the folder to prevent unlimited folder growth.


pkg/extensions/tbtc/recovery/storage.go
```diff
@@ -0,0 +1,135 @@
...
+		return 0, err
+	}
+	index, ok := addressIndexes[extendedPublicKey]
+	if ok {
+		return index, nil
+	}
+	return 0, nil
+}
+
+func (dis DerivationIndexStorage) readMap() (map[string]int, error) {
```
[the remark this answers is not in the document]
beaushinkle: I really like the idea of each persistence having it's own parent folder name - that way the user configures a single root storage directory like `~/local/storage/keep-ecdsa/1`, and then you wind up with:
```
~/local/storage/keep-ecdsa/1/beacon-group-data
~/local/storage/keep-ecdsa/1/keep-group-data
~/local/storage/keep-ecdsa/1/tbtc-data 
```
and then none of them step on each other's toes

pkg/extensions/tbtc/recovery/storage.go
```diff
@@ -0,0 +1,135 @@
...
+import (
+	"crypto/sha256"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"os"
+
+	"github.com/keep-network/keep-common/pkg/encryption"
+)
+
```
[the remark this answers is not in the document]
beaushinkle: > and would abstract a new type of persistence to keep-common at some point

Yup, I think this is the eventual end-state! I think the time to do that is when we have a second use-case that wants to share code.

pkg/extensions/tbtc/recovery/storage.go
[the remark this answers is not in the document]
beaushinkle: I like it! Do you want the files containing the btc addresses to be encrypted?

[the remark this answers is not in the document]
pdyraga: The only problem with this approach is backward compatibility. We'd have all our operators update their directory structure or update them automatically in the client but then, a downgrade is complicated. Long-term I think it makes sense, short-term... I'd consider nesting a new persistence next to `current/archive/snapshot` or add a configuration option to `[bitcoin]` namespace in the config to configure a path to a directory where we'd store derivation indices.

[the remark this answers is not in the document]
pdyraga: Actually, I've just checked the RFC and it says that:

> To avoid strictly requiring a connection to a Bitcoin node, this RFC proposes tracking the latest address generated, and allowing an additional config key, TargetDerivationStartIndex, that sets a starting point for such derivations. The client should write the updated index to the current/ directory that tracks active key shares, to an encrypted file named current-btc-refund-derivation-index. The operator should then seek to obtain from the staker an extended public key that is only used for this purpose. In cases where in-memory/on-disk derivation index and config file index both exist, the highest index should be used.

I see two problems with it though. `ReadAll` in the storage will try to load `current-btc-refund-derivation-index` file as a threshold signer and will explode. The second problem is what happens if the operator changes their BTC public key in the config? Do we use the existing derivation index or do we start from 0? I'd say starting from 0 makes more sense.

If we ignore the starting index issue, we could update the existing `persistentStorage` to do not read anything that [is not a valid address](https://github.com/keep-network/keep-ecdsa/blob/693068b463ca9268fda8732dab73bb750ff60dd0/pkg/registry/storage.go#L120-L126) and add a separate function next to `readRaw` that would read specific file bytes without trying to deserialize it to `tss.ThresholdSigner`. We'd also need `saveRaw`. 

I think the RFC aimed at least-effort implementation and honestly, I think that what you are doing here is even less effort than trying to adapt `persistentStorage` the way we want it to work.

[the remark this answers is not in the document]
pdyraga: Good question on encryption. I... don't think it needs to be encrypted. The same information can be evaluated just by looking at the config file 🤔 I think the encryption piece comes from the fact we aimed at least-effort implementation. Did I miss anything?

pkg/extensions/tbtc/recovery/storage.go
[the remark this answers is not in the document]
pdyraga: Yeah, I am comfortable with having this code in here. Let's just make sure it is kept separated and we can easily extract it any time we want.

pkg/extensions/tbtc/recovery/storage.go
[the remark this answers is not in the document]
beaushinkle: That's what I was wondering - if the extended public key is in plain text in the config, and the way the derivation path stuff works is open source, I don't see us *currently* protecting folks from figuring out who owns which btc addresses if someone manages to get filesystem access to the box

beaushinkle:
I updated the implementation to be the append-based design where each extended pubkey gets its own folder and we write the wallet address to a file named by the index

```
beaushinkle@Beaus-MBP ~/example-storage> tree
.
└── derivation_indexes
    └── extendedPublicKey
        └── 6
beaushinkle@Beaus-MBP ~/example-storage> cat derivation_indexes/extendedPublicKey/6
btcAddress⏎
```

Is what gets written to my local after running the included tests

beaushinkle:
@pdyraga how are we lookin'?

pdyraga:
👀

pkg/extensions/tbtc/recovery/storage_test.go
```diff
@@ -0,0 +1,57 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	expectedIndex := 0
+	if index != expectedIndex {
+		t.Errorf("incorrect extendedPublicKey index\nexpected: %d\nactual:   %d", expectedIndex, index)
+	}
+}
+
+func TestDerivationIndexStorage_WriteThenRead(t *testing.T) {
```
pdyraga, line 29: Shouldn't it be `SaveThenRead` given that we `.Save(...)`?

pkg/extensions/tbtc/recovery/storage.go
```diff
@@ -0,0 +1,106 @@
...
+	for _, file := range files {
+		fileIndex, err := strconv.Atoi(file.Name())
+		if err != nil {
+			return 0, err
+		}
+		if fileIndex > index {
+			index = fileIndex
+		}
+	}
+	return index + 1, nil
```
pdyraga, line 76: I think `+1` in this place is confusing. I'd expect `Read` in the storage to return the last index written to disk and if I want to get the next index that should be used, it should happen in another function, e.g. `GetNextIndex` that will `Read` and `+1`. WDYT?

pkg/extensions/tbtc/recovery/storage_test.go
```diff
@@ -0,0 +1,57 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	index, err := dis.Read(extendedPublicKey)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if index != expectedIndex {
+		t.Errorf("incorrect extendedPublicKey index\nexpected: %d\nactual:   %d", expectedIndex, index)
+	}
```
pdyraga, line 56: Can we add another test that will write more than one key?

pkg/extensions/tbtc/recovery/storage.go
```diff
@@ -0,0 +1,106 @@
...
+		return nil, err
+	}
+	return &DerivationIndexStorage{
+		path: path,
+	}, nil
+}
+
+func (dis DerivationIndexStorage) getStoragePath(extendedPublicKey string) string {
+	publicKeyDescriptor := extendedPublicKey[:4]
+	suffix := extendedPublicKey[len(extendedPublicKey)-8:]
```
pdyraga, line 30: Would be good to document how the key is transformed into directory name on some example.

pkg/extensions/tbtc/recovery/storage.go
```diff
@@ -0,0 +1,106 @@
...
+
+const (
+	fileName = "derivation_indexes"
+)
+
+type DerivationIndexStorage struct {
+	path string
+}
+
+func NewDerivationIndexStorage(path string) (*DerivationIndexStorage, error) {
```
pdyraga, line 18: I am thinking one more time about the layout and where `path` should be pointing to and the following structure came to my mind:
```
|
| - current
| - archive
| - snapshot
| - bitcoin <- that's where we are here
```
What do you think?

pkg/extensions/tbtc/recovery/storage.go
```diff
@@ -0,0 +1,106 @@
...
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"strconv"
+)
+
+const (
+	fileName = "derivation_indexes"
```
pdyraga, line 11: `fileName` or `directoryName`?

pkg/extensions/tbtc/recovery/storage_test.go
```diff
@@ -0,0 +1,57 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	index, err := dis.Read("extendedPublicKey")
+	if err != nil {
+		t.Fatal(err)
+	}
+	expectedIndex := 0
+	if index != expectedIndex {
+		t.Errorf("incorrect extendedPublicKey index\nexpected: %d\nactual:   %d", expectedIndex, index)
```
pdyraga, line 25: Should we be able to check if there is something in the registry for the particular public key? This is in case it is used for the first time or if it was altered in the config.

pkg/extensions/tbtc/recovery/storage_test.go
beaushinkle: yup!

pkg/extensions/tbtc/recovery/storage.go
beaushinkle: I like the additional clarity

beaushinkle:
Alrighty - seems like the core approach is looking good. I'll rebase these commits, add more tests, clean up the api, etc and then ping you when it's ready to be a real PR

pkg/extensions/tbtc/recovery/storage.go
beaushinkle: Right now, we have an implementation that looks like:
```
beaushinkle@Beaus-MBP ~/example-storage> tree
.
├── archive
│   ├── 0x0120Db73c97869943Ba8eE87400978051860B6C0
│   │   └── membership_046dd0d18ac092d4cdaf65198bfa1e3535ae57de
│   ├── 0x03c81f70618702b0fceA6F1656cE07cd4c5132d6
│   │   └── membership_046dd0d18ac092d4cdaf65198bfa1e3535ae57de
├── current
│   ├── 0x582Db6B4417a7bEB6a2Ac2964824e01d8a5B9548
│   │   └── membership_046dd0d18ac092d4cdaf65198bfa1e3535ae57de
│   ├── 0x6c672F95DCDA92Dd0aB5F26dBD16eC29deF2312B
│   │   └── membership_046dd0d18ac092d4cdaf65198bfa1e3535ae57de
└── snapshot
    ├── 0x0120Db73c97869943Ba8eE87400978051860B6C0
    │   └── membership_046dd0d18ac092d4cdaf65198bfa1e3535ae57de.1613059456465
    ├── 0x03c81f70618702b0fceA6F1656cE07cd4c5132d6
    │   └── membership_046dd0d18ac092d4cdaf65198bfa1e3535ae57de.1614208084167
    ├── 0x0e0EcbCD48Cb498c5eF47C25fE35A4f86aA72271
    │   └── membership_046dd0d18ac092d4cdaf65198bfa1e3535ae57de.1615579334556
    ├── 0x14A81Fb36A4cd13EaDB1FE271DC522F5CD7B00Fd
└── derivation_indexes
    ├── exte_ublicKey
    └── xpub_zCAk9sk1
        └── 6
```
where `derivation_indexes` lives along side `archive`, `current`, and `snapshot`. This should be enough to make sure that the derivation index stuff never interferes with the existing keep signer serialization stuff and it's backwards compatible.

I was also thinking that if we needed to add more types of persistence, they could have their own unique folder names just like `derivation_indexes`. Do you want to see an additional layer of hierarchy like `bitcoin/derivation_indexes`?

pkg/extensions/tbtc/recovery/storage_test.go
beaushinkle: can you elaborate here?

beaushinkle:
Ready for another look!

pkg/client/client.go
```diff
@@ -961,6 +971,7 @@ func monitorKeepTerminatedEvent(
 	operatorPublicKey *operator.PublicKey,
 	keep chain.BondedECDSAKeepHandle,
 	keepsRegistry *registry.Keeps,
+	derivationIndexStorage *recovery.DerivationIndexStorage,
```
pdyraga, line 974: We'll have to refactor `client.go` once the integration branch changes are merged to `master`. We have been postponing it for too long and now we have 13 parameters in a function and pass these parameters through the entire stack of function calls. 😓 We should do this refactor a couple of months ago.

pkg/extensions/tbtc/recovery/derive_address.go
```diff
@@ -129,17 +129,27 @@ func deriveAddress(extendedPublicKey string, addressIndex uint32) (string, error
...
 	_, err := btcutil.DecodeAddress(beneficiaryAddress, chainParams)
 	if err != nil {
-		return deriveAddress(beneficiaryAddress, addressIndex)
-	} else {
-		return beneficiaryAddress, nil
+		addressIndex, err := storage.GetNextIndex(beneficiaryAddress)
+		if err != nil {
+			return "", err
+		}
+		derivedAddress, err := deriveAddress(beneficiaryAddress, uint32(addressIndex))
```
pdyraga, line 144: Why don't we return `uint32` from `GetNextIndex`?

pkg/extensions/tbtc/recovery/derive_address.go
```diff
@@ -129,17 +129,27 @@ func deriveAddress(extendedPublicKey string, addressIndex uint32) (string, error
...
+		if err != nil {
+			return "", err
+		}
+		err = storage.Save(beneficiaryAddress, addressIndex, derivedAddress)
+		if err != nil {
+			return "", err
+		}
```
pdyraga, line 151: Should we `Save` the new address once we know the transaction was signed and broadcasted (or at least written to logs)? Right now, in case of errors from signing, we'll keep writing new addresses to the storage and those addresses will never be used.

pkg/extensions/tbtc/recovery/derive_address_test.go
```diff
@@ -223,8 +225,18 @@ var resolveAddressData = map[string]struct {
...
+				t.Fatal(err)
+			}
+			defer os.RemoveAll(dir)
+			dis, err := NewDerivationIndexStorage(dir)
+			if err != nil {
+				t.Fatal(err)
+			}
 			resolvedAddress, err := ResolveAddress(
 				testData.beneficiaryAddress,
+				dis,
```
pdyraga, line 239: I don't have enough context but shouldn't this test fail now? There are no files in temp dir created so `ResolveAddress` will always use the first index while `deriveAddressTestData` uses a bunch of other address indexes.

Another option to do not use on-disk persistence here would be to define an interface or just accept a `GetIndex` function as a parameter to `ResolveAddress`, especially if we take into account when `Save` function should be called (see the previous comment).

pkg/extensions/tbtc/recovery/storage.go
pdyraga: > Do you want to see an additional layer of hierarchy like `bitcoin/derivation_indexes`?

Yes. The approach for storing data presented in #744 will let us organize the group data per chain:
```
$data_dir/celo/<current,snapshot,archive>
$data_dir/solana/<current,snapshot,archive>
$data_dir/sovryn/<current,snapshot,archive>
$data_dir/<current,snapshot,archive>     // Ethereum, for backward-compatibility 
                                         // but we can migrate it one day.
```

Having a separate directory for `bitcoin` data would play nicely with this structure, especially if we decide to store more information for `bitcoin` in the future than just derivation indexes.

pkg/extensions/tbtc/recovery/storage_test.go
pdyraga: I was not sure how the storage code will be used and I was thinking whether there should be a function allowing to query the storage to see if the public key with the given index was used or not. It seems that such functions are not necessary though.

pkg/extensions/tbtc/recovery/derive_address.go
pdyraga: Ah, it is here because we want to make sure two keeps cannot use the same index for BTC recovery? I wouldn't care about it for now, to be honest. Liquidations are not happening that often and even with the current version there is still a risk in case two liquidations happen at the same time. Especially that in the current version we derive address (and store new index to disk, no matter if it's been used or not) in `DoWithDefaultRetry` loop that is potentially retried after each problem, including announce protocol not completed successfully.

pdyraga (commented):
Left some comments, leaving this in @nkuba's hands now for a detailed review.

pkg/extensions/tbtc/recovery/storage.go
beaushinkle: Alrighty: I added a `chainName` constant so that now we get:
```
beaushinkle@Beaus-MBP ~> cd example-storage/
beaushinkle@Beaus-MBP ~/example-storage> tree
.
└── bitcoin
    └── derivation_indexes
        ├── xpub_zCAk9sk1
        │   ├── 172
        │   ├── 3544
        │   ├── 39
        │   ├── 5
        │   ├── 5090
        │   └── 513
        ├── ypub_6Z2Yy4ZH
        │   ├── 1692
        │   ├── 223
        │   ├── 48
        │   └── 8982
        └── zpub_VDaZuQT9
            ├── 112
            ├── 6311
            ├── 6999
            └── 8559
```

pkg/extensions/tbtc/recovery/derive_address_test.go
beaushinkle: > I don't have enough context but shouldn't this test fail now?

The test data we were using was a single extended public key at `/0` (the default), and the rest were different forms of btc addresses that don't touch the persistence codepath.

I edited the test to be less implementation specific by providing a list of used indexes in each test case - should be good now!

pkg/extensions/tbtc/recovery/derive_address.go
beaushinkle: I moved the `Save` piece to after the broadcast call, so now we have a lower chance of persisting unused indexes

pkg/client/client.go
beaushinkle: image

nkuba (changes requested):

pkg/extensions/tbtc/recovery/storage_test.go
```diff
@@ -0,0 +1,120 @@
...
+			dis, err := NewDerivationIndexStorage(dir)
+
+			if err != nil {
+				t.Fatal(err)
+			}
+			for _, input := range testData.inputs {
+				err = dis.Save(
+					input.publicKey,
+					uint32(input.index),
+					"<btc-address>",
```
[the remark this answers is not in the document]
beaushinkle: That's a little tricky to do without giving the test an explicit understanding of the implementation of the function.

I could write a `ReadBtcAddress` (or something similar) method in `storage.go`, but I don't have any use for that outside of testing. Thoughts?

In order to retrieve the saved btc address, I'd need to tell the testing code to retrieve the file

pkg/extensions/tbtc/recovery/storage.go
```diff
@@ -0,0 +1,141 @@
...
+func (dis DerivationIndexStorage) getStoragePath(extendedPublicKey string) (string, error) {
+	if len(extendedPublicKey) < 12 {
+		return "", fmt.Errorf("insufficient length for public key %s", extendedPublicKey)
+	}
+	publicKeyDescriptor := extendedPublicKey[:4]
+	suffix := extendedPublicKey[len(extendedPublicKey)-8:]
+	return fmt.Sprintf("%s/%s/%s/%s_%s", dis.path, chainName, directoryName, publicKeyDescriptor, suffix), nil
+}
+
+func ensureDirectoryExists(path string) error {
```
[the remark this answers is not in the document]
beaushinkle: Yeah - I haven't opened up a cross-project dependency before - how does that work?

pkg/extensions/tbtc/recovery/storage.go
```diff
@@ -0,0 +1,141 @@
...
+		return err
+	}
+
+	err = ensureDirectoryExists(dirPath)
+	if err != nil {
+		return err
+	}
+	filePath := fmt.Sprintf("%s/%d", dirPath, index)
+
+	return write(filePath, []byte(btcAddress))
```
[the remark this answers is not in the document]
beaushinkle: It gets overwritten, which should be okay given that extended public key + index always yields the same btc address:

Here's an example of a local run where I write to the same address + index multiple times:
```
beaushinkle@Beaus-MBP ~/e/b/derivation_indexes> cd xpub_1sameidx/
beaushinkle@Beaus-MBP ~/e/b/d/xpub_1sameidx> ls
777
beaushinkle@Beaus-MBP ~/e/b/d/xpub_1sameidx> cat 777
<btc-address>⏎
```

beaushinkle:
Alrighty, ready for another round

beaushinkle:
Any clues on what could be going on with the permission failures? 
```

=== Failed
=== FAIL: pkg/extensions/tbtc/recovery TestDerviationIndexStorage_BadPermissions/execute_only (0.00s)
    --- FAIL: TestDerviationIndexStorage_BadPermissions/execute_only (0.00s)
        storage_test.go:244: unexpected error
            expected: cannot read from the storage directory
            actual:   <nil>

=== FAIL: pkg/extensions/tbtc/recovery TestDerviationIndexStorage_BadPermissions/write_only (0.00s)
    --- FAIL: TestDerviationIndexStorage_BadPermissions/write_only (0.00s)
        storage_test.go:244: unexpected error
            expected: cannot read from the storage directory
            actual:   <nil>

=== FAIL: pkg/extensions/tbtc/recovery TestDerviationIndexStorage_BadPermissions/write_and_execute (0.00s)
    --- FAIL: TestDerviationIndexStorage_BadPermissions/write_and_execute (0.00s)
        storage_test.go:244: unexpected error
            expected: cannot read from the storage directory
            actual:   <nil>

=== FAIL: pkg/extensions/tbtc/recovery TestDerviationIndexStorage_BadPermissions/read_only (0.00s)
    --- FAIL: TestDerviationIndexStorage_BadPermissions/read_only (0.00s)
        storage_test.go:244: unexpected error
            expected: cannot write to the storage directory
            actual:   <nil>

=== FAIL: pkg/extensions/tbtc/recovery TestDerviationIndexStorage_BadPermissions/read_and_execute (0.00s)
    --- FAIL: TestDerviationIndexStorage_BadPermissions/read_and_execute (0.00s)
        storage_test.go:244: unexpected error
            expected: cannot write to the storage directory
            actual:   <nil>

=== FAIL: pkg/extensions/tbtc/recovery TestDerviationIndexStorage_BadPermissions/read_and_write (0.00s)
    --- FAIL: TestDerviationIndexStorage_BadPermissions/read_and_write (0.00s)
        storage_test.go:244: unexpected error
            expected: cannot write to the storage directory
            actual:   <nil>

=== FAIL: pkg/extensions/tbtc/recovery TestDerviationIndexStorage_BadPermissions (0.00s)
```
They all work locally @_@. Judging from the error, it would seem like the workflow build is perfectly okay with me reading/writing from directories that have been chmod'd to prevent that.

I would also think that if the chmod didn't go through, then this would fail:
```
			err = os.Chmod(dir, os.FileMode(testData.mode))
			if err != nil {
				t.Fatal(err)
			}
```

pkg/extensions/tbtc/recovery/storage.go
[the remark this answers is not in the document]
nkuba: We need to declare the mentioned functions as exported in the `keep-common` repo and once we push the commit we can pin it here in `keep-ecdsa` by executing `go get github.com/keep-network/keep-common@COMMIT_HASH`.

pkg/extensions/tbtc/recovery/storage.go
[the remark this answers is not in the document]
nkuba: Cool, cool. I would advise covering it with a unit test.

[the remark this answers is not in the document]
beaushinkle: covered!

beaushinkle:
Update: the permission tests were failing because the docker that runs the workflow test builds/runs `go` as `root`, and `root` ignores any set permissions.

I added a commit that builds the docker as a non-root user, which hopefully fixes this!

go.sum
```diff
@@ -449,8 +449,15 @@ github.com/keep-network/go-libp2p-bootstrap v0.0.0-20200423153828-ed815bc50aec/g
 github.com/keep-network/keep-common v1.3.1-0.20210304083053-baf172c68552/go.mod h1:eM8DwRcQYzq5TYQSn9iOFXfvnAh2w1xWQmpIoG1cdpo=
 github.com/keep-network/keep-common v1.4.1-0.20210319095805-ebf46d0b62db h1:CrAKFfc0ynlQv0M/IwTVALSh6nDF3Kki71Y6GGybJQk=
 github.com/keep-network/keep-common v1.4.1-0.20210319095805-ebf46d0b62db/go.mod h1:E515VGCNeth7/yJafIWcZyZrHslQHBUKwIjpdipqMM4=
+github.com/keep-network/keep-common v1.4.1-0.20210503154715-f16f7bd7efa8 h1:EHXQS+2nN8fJNaPj32wf4XrZBxgEnyJaSbER/vTSeuk=
```
nkuba, line 452: Can we try to run `go mod tidy`?

pkg/client/client.go
```diff
@@ -1122,6 +1134,18 @@ func monitorKeepTerminatedEvent(
...
 
+						if bitcoinConfig.BeneficiaryAddress != beneficiaryAddress {
+							err = derivationIndexStorage.Save(bitcoinConfig.BeneficiaryAddress, addressIndex)
+							if err != nil {
+								logger.Errorf(
+									"failed to persist the latest address derivation index for keep [%s]: [%v]",
+									keep.ID(),
+									err,
+								)
+								return err
```
nkuba, line 1145: As we return the error this means we will retry the execution of the whole function, right? Maybe we should just log an error (without return) and let the process complete?

pkg/extensions/tbtc/recovery/storage.go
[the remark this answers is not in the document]
nkuba: I believe these are already exposed in keep-common, can we update the code we have here to utilize them?

beaushinkle:
good for another look!

nkuba (approved):

beaushinkle:
Okay - got approval and it's green. @pdyraga or @Shadowfiend do you want to do the honors?

nkuba:
@beaushinkle we need a conflict resolution here. Unless @Shadowfiend and @pdyraga drop any comments I'll merge this PR as soon as the conflict is resolved.

beaushinkle:
green!

nkuba (approved):

PR #759, opened 2021-04-08
Updates to run client guide after ropsten migrations

nkuba:
We update guides with addresses of the contracts that were recently migrated on ropsten and are backed by keep-test nodes. Jobs that migrated contracts: 
- https://github.com/keep-network/keep-ecdsa/actions/runs/690005293
- https://github.com/keep-network/tbtc/actions/runs/698185957

We no longer define Sanctioned Applications property in the `config.toml` file. Instead, we have TBTCSystem property.

### DIFF
--- a/docs/run-keep-ecdsa.adoc
+++ b/docs/run-keep-ecdsa.adoc
@@ -394,8 +394,8 @@ Contract addresses needed to boot a Keep ECDSA client:
 |BondedECDSAKeepFactory
 |`0xA7d9E842EFB252389d613dA88EDa3731512e40bD`
 
-|Sanctioned Applications
-|`0xe20A5C79b39bC8C363f0f49ADcFa82C2a01ab64a` (tBTC's system contract)
+|TBTCSystem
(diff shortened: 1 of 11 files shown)


docs/run-keep-ecdsa.adoc
```diff
@@ -394,8 +394,8 @@ Contract addresses needed to boot a Keep ECDSA client:
 |BondedECDSAKeepFactory
 |`0xA7d9E842EFB252389d613dA88EDa3731512e40bD`
 
-|Sanctioned Applications
-|`0xe20A5C79b39bC8C363f0f49ADcFa82C2a01ab64a` (tBTC's system contract)
+|TBTCSystem
+|`0xe20A5C79b39bC8C363f0f49ADcFa82C2a01ab64a`
```
michalinacienciala, line 398: Shouldn't it be `0x3b9072d3F1E3a7af139A9eF2A4f035cBFcf27BaF`?

michalinacienciala (approved):

PR #762, opened 2021-04-12
Read chaincfg.Params from Config

beaushinkle:
This PR allows the user to supply a target BTC network at `Extensions.TBTC.BTCRefunds.ChainName` in the configuration. If none is supplied, we default to the main network.

tagging @nkuba for review!

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -155,7 +155,9 @@ func TestParseChainParams(t *testing.T) {
 	}
 	for testName, testData := range parseChainParamTests {
 		t.Run(testName, func(t *testing.T) {
-			// use a string builder and a single-value list to represent optionality
+			// Use a string builder and a single-value list to represent optionality.
+			// If the list has a chain name, we define it in the config, otherwise,
+			// we have an empty `[Extensions.TBTC.BTCRefunds]` section.
 			var b strings.Builder
 			fmt.Fprint(&b, "[Extensions.TBTC.BTCRefunds]")
 			for _, name := range testData.chainName {
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -189,15 +189,11 @@ func TestParseChainParams_ExpectedFailure(t *testing.T) {
 	if err == nil {
 		t.Fatalf("expecting an error but found none")
 	}
-	if !errorContains(err, expectedError) {
+	if expectedError != err.Error() {
 		t.Errorf(
 			"unexpected error\nexpected: %s\nactual:   %v",
 			expectedError,
 			err,
 		)
 	}
-}
-
-func errorContains(err error, expected string) bool {
-	return strings.Contains(err.Error(), expected)
 }
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -99,8 +99,8 @@ func TestReadConfig(t *testing.T) {
 			readValueFunc: func(c *Config) interface{} { return c.Extensions.TBTC.BTCRefunds.MaxFeePerVByte },
 			expectedValue: int32(73),
 		},
-		"Extensions.TBTC.BTCRefunds.ChainName": {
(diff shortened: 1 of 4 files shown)


config/config_test.go
```diff
@@ -105,5 +121,81 @@ func TestReadConfig(t *testing.T) {
...
+			chaincfg.MainNetParams,
+		},
+		"empty": {
+			[]string{""},
+			chaincfg.MainNetParams,
+		},
+	}
+	for testName, testData := range parseChainParamTests {
+		t.Run(testName, func(t *testing.T) {
+			// use a string builder and a single-value list to represent optionality
```
beaushinkle, line 158: I'm curious as to whether or not you like this potentially-empty-list approach as a way to handle optionality.

The other main option that I can see is splitting the empty option off into its own test case. Thoughts?

pkg/chain/bitcoin/config.go
```diff
@@ -1,7 +1,33 @@
...
+	"fmt"
+
+	"github.com/btcsuite/btcd/chaincfg"
+)
+
 // Config stores configuration related to recovering BTC from a closed keep.
 type Config struct {
 	BeneficiaryAddress string
 	MaxFeePerVByte     int32
+	ChainName          string
```
nkuba, line 13: Would it be worth adding `Bitcoin` to the property's name? `BitcoinChain` or `BitcoinChainName`?

config/config_test.go
nkuba: I like this solution, although I have to admit that I was confused at the beginning and had to figure out what's happening. It may be enough to enhance the comment clarifying more details of this approach. 👌🏻

config/config_test.go
```diff
@@ -105,5 +121,81 @@ func TestReadConfig(t *testing.T) {
...
+	config := &Config{}
+	if _, err := toml.Decode(configString, config); err != nil {
+		t.Fatal(err)
+	}
+	_, err := config.Extensions.TBTC.BTCRefunds.ChainParams()
+	expectedError := "unable to find chaincfg param for name: [bleeble blabble]"
+	if err == nil {
+		t.Fatalf("expecting an error but found none")
+	}
+	if !errorContains(err, expectedError) {
```
nkuba, line 190: Thoughts on comparing it here instead of defining a function?
```suggestion
	if expectedError != err.Error() {
```

pkg/chain/bitcoin/config.go
```diff
@@ -1,7 +1,33 @@
...
+	switch c.ChainName {
+	case "mainnet":
+		return &chaincfg.MainNetParams, nil
+	case "regtest":
+		return &chaincfg.RegressionNetParams, nil
+	case "simnet":
+		return &chaincfg.SimNetParams, nil
+	case "testnet3":
+		return &chaincfg.TestNet3Params, nil
+	case "":
```
nkuba, line 27: We could combine two cases.

```suggestion
	case "mainnet", "":
```

nkuba (commented):
It looks very good to me. I left some minor comments to consider.

pkg/chain/bitcoin/config.go
beaushinkle: that's a handy one!

beaushinkle:
alrighty - good for another pass

config/config_test.go
```diff
@@ -105,5 +121,79 @@ func TestReadConfig(t *testing.T) {
...
+		"empty": {
+			[]string{""},
+			chaincfg.MainNetParams,
+		},
+	}
+	for testName, testData := range parseChainParamTests {
+		t.Run(testName, func(t *testing.T) {
+			// Use a string builder and a single-value list to represent optionality.
+			// If the list has a chain name, we define it in the config, otherwise,
+			// we have an empty `[Extensions.TBTC.BTCRefunds]` section.
```
nkuba, line 160: 🥇

nkuba (approved):

PR #766, opened 2021-04-16
Store the TransactionHash and Index On FundingInfo

beaushinkle:
We also correctly calculate the transaction hash by reversing the first 32 bytes of the outpoint before encoding them in hex, since the outpoint stores them in little-endian.

This one is a little trickier to test, since not-reversing the bytes also passes our validators and we hard-code our funding info for unit tests. With that in mind, here's the whole flow:

First, pull down the branch for https://github.com/keep-network/local-setup/pull/71 (or merge that first), since we need the `getBalance` and `sendRawTransaction` commands it provides.

Generate 3 new bitcoin addresses via:
```
$ bitcoind-wallet getNewAddress
```

Edit `keep-ecdsa/configs/config.local.1.toml`, `keep-ecdsa/configs/config.local.2.toml`, and `keep-ecdsa/configs/config.local.3.toml` to have:

```
[Extensions.TBTC]
	TBTCSystem = "<address from [ethereum.ContractAddresses.TBTCSystem]"

  [Extensions.TBTC.BTCRefunds]
  	BeneficiaryAddress = "<address from bitcoind-wallet getNewAddress> "
  	MaxFeePerVByte = 75
	BitcoinChainName = "regtest"

[TSS]
	PreParamsTargetPoolSize = 2 # not strictly required, but makes the whole thing faster
```

Run keep-core and the 3 keep ecdsa nodes

Create a new deposit from `local-setup/tbtc.js`:
```
$ bin/tbtc.js --rpc ws://127.0.0.1:8546 deposit new 10000000
```
Watch for a message like:
```
Got deposit address: <funding address> ; fund with: 10000000 satoshis please.
```
Then, fund that address:
```
$ bitcoind-wallet sendToAddress <funding address> 0.1
```

Wait for it to confirm with a message like:
```
Deposit is active, minting...
[object Promise]	99950000000000000
```

Now, we have to undercollateralize the deposit from the `local-setup/tbtc/solidity/scripts` folder:
```
$ truffle exec multiply-local-eth-btc-price.js 0.7
```
Once that completes, you can check the active deposits from `local-setup/tbtc.js`:
```
$ bin/tbtc.js --rpc ws://127.0.0.1:8546 deposit list --vending-machine

(node:28834) ExperimentalWarning: Importing JSON modules is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
<deposit address>	                        ACTIVE	10000000	104
```

And then liquidate it:

```
[...]

PR #768, opened 2021-04-19
Connect an Optional Electrs Endpoint

beaushinkle:
dependencies: #766 

This PR adds a new configuration field: `Extensions.TBTC.ElectrsURL = "http://localhost:3002"`, and then uses that information to retrieve vbyte fee estimates (at `/fee-estimates`) and submit transactions (at `POST /tx`).

If no configuration is provided, we default that URL to `https://www.blockstream.info/api/` (see https://www.blockstream.info/api/fee-estimates), and if that configuration is specifically set to `""`, we fall back on hardcoded fee values and logging a message to broadcast the transaction five times.

tagging @nkuba for review

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -16,7 +16,6 @@ import (
 	"github.com/keep-network/keep-core/pkg/net"
 	"github.com/keep-network/keep-core/pkg/operator"
 	"github.com/keep-network/keep-ecdsa/pkg/chain"
-	"github.com/keep-network/keep-ecdsa/pkg/chain/bitcoin"
 	"github.com/keep-network/keep-ecdsa/pkg/client/event"
 	"github.com/keep-network/keep-ecdsa/pkg/ecdsa/tss"
 	"github.com/keep-network/keep-ecdsa/pkg/extensions/tbtc"
@@ -60,7 +59,7 @@ func Initialize(
 	networkProvider net.Provider,
 	persistence persistence.Handle,
 	clientConfig *Config,
-	bitcoinConfig *bitcoin.Config,
+	tbtcConfig *chain.Config,
 	tssConfig *tss.Config,
 ) *Handle {
 	keepsRegistry := registry.NewKeepsRegistry(
@@ -203,7 +202,7 @@ func Initialize(
 				tbtcApplicationHandle,
 				networkProvider,
 				clientConfig,
-				bitcoinConfig,
+				tbtcConfig,
 				tssNode,
 				operatorPublicKey,
 				keep,
@@ -221,7 +220,7 @@ func Initialize(
 		tbtcApplicationHandle,
 		networkProvider,
 		clientConfig,
-		bitcoinConfig,
+		tbtcConfig,
 		tssNode,
 		operatorPublicKey,
 		keepsRegistry,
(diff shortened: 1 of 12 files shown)


pkg/client/client.go
```diff
@@ -1113,11 +1129,14 @@ func monitorKeepTerminatedEvent(
...
 						}
 
-						logger.Warningf("Please broadcast Bitcoin transaction %s", recoveryTransactionHex)
-						logger.Warningf("Please broadcast Bitcoin transaction %s", recoveryTransactionHex)
-						logger.Warningf("Please broadcast Bitcoin transaction %s", recoveryTransactionHex)
-						logger.Warningf("Please broadcast Bitcoin transaction %s", recoveryTransactionHex)
-						logger.Warningf("Please broadcast Bitcoin transaction %s", recoveryTransactionHex)
+						err = electrsConnection.Broadcast(recoveryTransactionHex)
+						if err != nil {
+							logger.Errorf(
```
nkuba, line 1134: In case of a failure maybe we should still log 5x warning?

pkg/extensions/tbtc/recovery/electrs.go
```diff
@@ -0,0 +1,77 @@
+package recovery
```
nkuba, line 1: Maybe this code should go into `pkg/chain/bitcoin` as we already got bitcoin specific config there?

pkg/extensions/tbtc/recovery/electrs.go
```diff
@@ -0,0 +1,77 @@
...
+
+func (e *ElectrsConnection) setClient(client httpClient) {
+	e.client = client
+}
+
+// Broadcast broadcasts a transaction the configured bitcoin network.
+func (e ElectrsConnection) Broadcast(transaction string) error {
+	if e.apiURL == "" {
+		for i := 0; i < 5; i++ {
+			logger.Warningf("Please broadcast Bitcoin transaction %s", transaction)
```
nkuba, line 38: Oh, I see the warning log has been moved here. I think it would be better if we pull it out to the previous location.

pkg/extensions/tbtc/recovery/electrs.go
```diff
@@ -0,0 +1,77 @@
...
+		return nil
+	}
+	resp, err := e.client.Post(fmt.Sprintf("%s/tx", e.apiURL), "text/plain", strings.NewReader(transaction))
+	if err != nil {
+		return err
+	}
+	if resp.StatusCode != 200 {
+		return fmt.Errorf("something went wrong with broadcast: [%s], transaction: [%s]", resp.Status, transaction)
+	}
+	logger.Info("successfully broadcast the bitcoin transaction")
```
nkuba, line 49: Does the `resp` contain any transaction ID we could log here as well?

nkuba (commented):
We need to remember to update `config.toml.SAMPLE` when we finish up updating the config structure.

pkg/extensions/tbtc/recovery/electrs.go
beaushinkle: I ended up keeping them in both places: if we configure the electrs URL to be `""` specifically, we do the 5x log, and if something goes wrong with the broadcast we fall back to the 5x log

pkg/extensions/tbtc/recovery/electrs.go
beaushinkle: Yup! we're now reading the response body and including it in the logs:

```
2021-04-27T17:13:44.478-0400	INFO	bitcoin	successfully broadcast the bitcoin transaction: 4d8cd9e11af8a421a5519bbc5172dfb17c408557973c3998077e12bd0d74561c
```
image

image

nkuba (approved):

PR #771, opened 2021-04-29
Update the Dockerfile to Build the Client as a non-root User

beaushinkle:
Root has special permissions where it overrides any chmod settings and is allowed to do anything. This makes permission tests (like the permission tests proposed in https://github.com/keep-network/keep-ecdsa/pull/756/commits/030a31817a2939fa12a3891a4ec24967d7866ace) fail. To fix that, we build the app as a new user with more limited permissions and test with that user instead.

tagging @nkuba for review

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,8 +33,12 @@ RUN apk add --update --no-cache \
 	rm -rf /var/cache/apk/ && mkdir /var/cache/apk/ && \
 	rm -rf /usr/share/man
 
+RUN adduser -D -g '' keep
+RUN chown keep $BIN_PATH
+USER keep
+
 # Install Solidity compiler.
-COPY --from=ethereum/solc:0.5.17 /usr/bin/solc /usr/bin/solc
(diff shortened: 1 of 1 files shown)


Shadowfiend:
Since we never change the user back away from `keep`, I'm wondering if this has the same breaking caveats as #467. I'm comfortable if the answer is “yes”, except if that's the case we should ALSO change the user to `nobody` before `CMD []` so we can kill that bird as well and coalesce this into a single breaking change for operators to manage. We'll also need to make sure to document the breaking nature + include instructions on how to manage it in the release notes and announcement.

beaushinkle:
I don't think this is a breaking change for normal use since the user doesn't change for the execution environment, though I'm not super sure how to test that well

Shadowfiend:
Ah yep, right you are. Word, still want to get that runtime change in, but no need to muck with it here.

Shadowfiend (approved):
I'm comfortable here and all the builds are green, let's do it.

beaushinkle:
:tada:

PR #773, opened 2021-05-03
Load non-sensitive data from file instead of from secrets

michalinacienciala:
Non-sensitive data previously accessed by using `secrets` context is
now accessed by invoking `keep-network/load-env-variables` action and
using `env` context. 
BACKGROUND:
We've been storing id of the test environment as 
`KEEP_TEST_ETH_NETWORK_ID` organization secret in GitHub's settings to
allow for easy access to that value from multiple projects.
But it turned out to be problematic, as GHA in some situations redacts
workflow logs and actions outputs/inputs based on the values stored in
secrets. So if the secret holds small numeric value, it is quite
probable that at some point in time this value will be recognized as a
substring of `github.sha` or another variable and the `github.sha` will
be redacted with `***` in place of the recognized pattern. If such
redacted value is configured as input of some action, the action gets
invoked by the GHA without that input, which can cause run failure or
can mess up the workflow results.
As a solution for this problem we decided to no longer store
frequently-used, non-sensitive data such as `KEEP_TEST_ETH_NETWORK_ID`
in GitHub's secrets, but instead store them in a file kept in
`keep-network/ci` repository and read that data to `env` context in
order to be able to use it when running workflows. A
`keep-network/load-env-variables` action has been implemented to allow
for loading of the variables to `env` context in one step. The scope of
the loaded variables is limited to the job in which they were loaded.
To use the variables in other job, action `load-env-variables` needs to
be executed there as well.


michalinacienciala:
Test configuration (references to development branch) is now removed.
Here are the links to the successfully executed workflows which used that test configuration:
Go / Ethereum: https://github.com/keep-network/keep-ecdsa/actions/runs/809209933
Go / Celo: https://github.com/keep-network/keep-ecdsa/actions/runs/809138054
Solidity: https://github.com/keep-network/keep-ecdsa/actions/runs/809137548

.github/workflows/client-celo.yml
```diff
@@ -67,30 +75,30 @@ jobs:
...
             HOST_CHAIN=celo
             REVISION=${{ github.sha }}
           # VERSION= ? TODO: Configure version, sample: 1.7.6
           push: |
-            ${{ github.ref == 'refs/heads/master'
-              && github.event_name != 'pull_request' }}
\ No newline at end of file
+            ${{ (github.ref == 'refs/heads/master'
+              && github.event_name != 'pull_request')
+              || github.ref == 'refs/heads/rfc-18/load-env-var' }}
```
nkuba, line 104: Can we remove it already?

nkuba (changes requested):

.github/workflows/client-celo.yml
michalinacienciala: Done, see 7c45d4f3dea23cfcdfb84431e543788a41caf425.

nkuba (approved):

PR #775, opened 2021-05-05
Split upload of artifacts for initcontainer to two steps

michalinacienciala:
When we're building initcontainer from Docker image, we expect to have
four contracts (artifacts of previous job) in the step's working
directory.
When upload of artifacts was done in one step, artifacts (which come
from two different directories) were being saved together with their
path structure (least common ancestor is used as root directory).
Spliting the action to two allowed for upload of files without their
parent folders.

### DIFF
--- a/.github/workflows/contracts.yml
+++ b/.github/workflows/contracts.yml
@@ -224,13 +224,19 @@ jobs:
           echo //registry.npmjs.org/:_authToken=${{ secrets.NPM_TOKEN }} > .npmrc
           npm publish --access=public
 
-      - name: Upload contracts artifacts for initcontainer build
+      - name: Upload contracts artifacts for initcontainer build (from solidity/build)
         uses: actions/upload-artifact@v2
         with:
           name: Selected contracts (Node.js ${{ matrix.node-version }})
           path: |
             ./solidity/build/contracts/BondedECDSAKeepFactory.json
             ./solidity/build/contracts/KeepBonding.json
+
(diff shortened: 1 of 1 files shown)


.github/workflows/contracts.yml
```diff
@@ -224,13 +224,19 @@ jobs:
           echo //registry.npmjs.org/:_authToken=${{ secrets.NPM_TOKEN }} > .npmrc
           npm publish --access=public
 
-      - name: Upload contracts artifacts for initcontainer build
+      - name: Upload contracts artifacts for initcontainer build (from solidity/build)
```
nkuba, line 227:
```suggestion
      - name: Upload keep-ecdsa contracts artifacts for initcontainer build
```

.github/workflows/contracts.yml
```diff
@@ -224,13 +224,19 @@ jobs:
...
         uses: actions/upload-artifact@v2
         with:
           name: Selected contracts (Node.js ${{ matrix.node-version }})
           path: |
             ./solidity/build/contracts/BondedECDSAKeepFactory.json
             ./solidity/build/contracts/KeepBonding.json
+
+      - name: Upload contracts artifacts for initcontainer build (from solidity/node_modules)
```
nkuba, line 235:
```suggestion
      - name: Upload keep-core contracts artifacts for initcontainer build
```

nkuba (approved):
I left two non-blocking suggestions.

PR #776, opened 2021-05-05
Local installation scripts refactoring

nkuba:
This PR refactors local installation scripts to utilize changes introduced in https://github.com/keep-network/keep-ecdsa/pull/774.

Changes:
- use npm links between local modules to obtain migrated contracts instead of referring to the path where the artifacts sit,
- introduce contracts only flag to installation script to let skip go client installation (useful for dashboard installation).

Follow-up PRs in other repositories are expected to be created.

Refs: https://github.com/keep-network/keep-core/pull/2452


michalinacienciala (changes requested):
Please also look at linter suggestions and fix accordingly

michalinacienciala (approved):

PR #779, opened 2021-05-07
Clean up eslint Errors

beaushinkle:
This PR is the result of running `eslint . --fix` in the `solidity` and `staker-rewards` directories. Without doing this, JS files can't be committed without ignoring the pre-commit hook.


beaushinkle:
The workflow run of `npm run lint` wants there explicitly to be no trailing `,`s, but locally it wants the opposite. Anyone know what's up there?

Shadowfiend:
Try running `npm i` in `solidity` to make sure you're on the right version fo the keep eslint config.

Shadowfiend:
Oh… Another possibility is that if you run it at the repo root, it doesn't actually pick up the configs for each child directory. i.e., you may need to run `eslint --fix` in each subdirectory that has it configured, rather than running it at the repo root. Not sure how you went about running the `--fix`.

beaushinkle:
The changes in this PR are a result of the following commands:
```
cd solidity
npm i
eslint . --fix
cd ../staker-rewards
npm i
eslint . --fix
```

Shadowfiend:
Another possibility: `eslint` also may be running a global eslint install. `npx eslint` will run the one that's installed by the local package. `npx eslint . --fix` seems to revert the comma changes you're flagging here for me. Keep in mind also that to deal with this most if not all of our packages have a `lint:fix` script and a `lint:fix:js` script that run in the correct namespace. `npm run lint:fix:js` will do the trick.

beaushinkle:
image

beaushinkle:
Okay - figured out that this was related to my local ecdsa prettier inheriting configuration from the parent `local-setup` directory, but the github workflow has an isolated ecdsa directory.

I opened up https://github.com/keep-network/local-setup/pull/83 so now we're good

PR #785, opened 2021-05-11
Check Unused Bitcoin Adddresses

beaushinkle:
The current persistence solution has two main issues:

* Multiple concurrent calls to `GetNextIndex` return the same index. This can happen if a btc-eth price shift happens that triggers multiple liquidations in which an operator is a signer. Since this would lead to multiple liquidations being sent to the same btc address, this leaks user information and is counter to the purpose of using HD wallets.

* We do not leverage the electrs connection to check to see if a derived address is actually unused. This can lead to address reuse, which leaks user information in the same way described above.

To fix this, we make the following behavior changes:

* `GetNextIndex` is in charge of saving used addresses, and accepts a bitcoin handler that can check to see if addresses are used or not
* `GetNextIndex` keeps searching for unused addresses until it finds one, and blocks other concurrent calls to `GetNextIndex` until it does.
* Multiple calls to `GetNextIndex` have to wait until the first call finds an unused index, and then marks it as used, making sure that each call returns a different value.

open questions:
* How do we want the unused address search to work? Right now, it starts at the next index and checks the next 100 after that. That was a naive/easy implementation rather than a solid one to get this in a reviewable state.

tagging @Shadowfiend and @nkuba for review

### DIFF
--- a/pkg/chain/bitcoin/electrs.go
+++ b/pkg/chain/bitcoin/electrs.go
@@ -99,10 +99,16 @@ func (e ElectrsConnection) IsAddressUnused(btcAddress string) (bool, error) {
 		return false, err
 	}
 	if resp.StatusCode != 200 {
+		transactionIDBuffer := new(strings.Builder)
+		_, err = io.Copy(transactionIDBuffer, resp.Body)
+		if err != nil {
+			logger.Error("something went wrong trying to unmarshal the error response for address %s", btcAddress)
+		}
 		return false, fmt.Errorf(
-			"something went wrong trying to get information about address %s: [%s]",
+			"something went wrong trying to get information about address %s - status: [%s], payload: [%s]",
 			btcAddress,
 			resp.Status,
+			transactionIDBuffer.String(),
 		)
 	}
 
--- a/pkg/chain/bitcoin/electrs.go
+++ b/pkg/chain/bitcoin/electrs.go
@@ -61,8 +61,8 @@ func (e ElectrsConnection) Broadcast(transaction string) error {
 	return nil
 }
 
-// VbyteFee retrieves the 25-block estimate fee per vbyte on the bitcoin network.
-func (e ElectrsConnection) VbyteFee() (int32, error) {
+// VbyteFeeFor25Blocks retrieves the 25-block estimate fee per vbyte on the bitcoin network.
+func (e ElectrsConnection) VbyteFeeFor25Blocks() (int32, error) {
 	if e.apiURL == "" {
 		return 0, fmt.Errorf("attempted to call VbyteFee with no apiURL")
 	}
--- a/pkg/chain/bitcoin/electrs.go
+++ b/pkg/chain/bitcoin/electrs.go
@@ -89,3 +89,28 @@ func (e ElectrsConnection) VbyteFee() (int32, error) {
 	logger.Info("retrieved a vbyte fee of [%d]", fee)
 	return int32(fee), nil
 }
+
+func (e ElectrsConnection) IsAddressUnused(btcAddress string) (bool, error) {
+	if e.apiURL == "" {
+		return true, nil
+	}
+	resp, err := e.client.Get(fmt.Sprintf("%s/address/%s/txs", e.apiURL, btcAddress))
+	if err != nil {
+		return false, err
+	}
+	if resp.StatusCode != 200 {
+		return false, fmt.Errorf(
+			"something went wrong trying to get information about address %s: [%s]",
+			btcAddress,
+			resp.Status,
+		)
+	}
+
+	responses := []interface{}{}
+	err = json.NewDecoder(resp.Body).Decode(&responses)
+	if err != nil {
+		return false, err
+	}
+
+	return len(responses) == 0, nil
+}
--- a/pkg/chain/bitcoin/electrs.go
+++ b/pkg/chain/bitcoin/electrs.go
@@ -82,7 +82,7 @@ func (e ElectrsConnection) VbyteFeeFor25Blocks() (int32, error) {
 			return err
 		}
 		if resp.StatusCode != 200 {
-			return fmt.Errorf("something went wrong with broadcast: [%s]", resp.Status)
+			return fmt.Errorf("something went wrong retreiving the vbyte fee: [%s]", resp.Status)
 		}
 
 		var fees map[string]float32
@@ -104,6 +104,8 @@ func (e ElectrsConnection) VbyteFeeFor25Blocks() (int32, error) {
 	return vbyteFee, nil
 }
 
+// IsAddressUnused returns true if and only if the supplied bitcoin address has
+// no recorded transactions.
 func (e ElectrsConnection) IsAddressUnused(btcAddress string) (bool, error) {
 	if e.apiURL == "" {
 		return false, fmt.Errorf("attempted to call IsAddressUnused with no apiURL")
--- a/pkg/chain/bitcoin/electrs.go
+++ b/pkg/chain/bitcoin/electrs.go
@@ -26,15 +26,17 @@ type httpClient interface {
 
 // ElectrsConnection exposes a native API for interacting with an electrs http API.
 type ElectrsConnection struct {
-	apiURL string
-	client httpClient
+	apiURL  string
+	client  httpClient
+	timeout time.Duration
 }
 
 // NewElectrsConnection is a constructor for ElectrsConnection.
 func NewElectrsConnection(apiURL string) *ElectrsConnection {
 	return &ElectrsConnection{
-		apiURL: apiURL,
-		client: http.DefaultClient,
+		apiURL:  apiURL,
+		client:  http.DefaultClient,
+		timeout: defaultTimeout,
 	}
 }
 
@@ -47,7 +49,7 @@ func (e ElectrsConnection) Broadcast(transaction string) error {
 	if e.apiURL == "" {
 		return fmt.Errorf("attempted to call Broadcast with no apiURL")
 	}
-	return utils.DoWithDefaultRetry(defaultTimeout, func(ctx context.Context) error {
+	return utils.DoWithDefaultRetry(e.timeout, func(ctx context.Context) error {
 		resp, err := e.client.Post(fmt.Sprintf("%s/tx", e.apiURL), "text/plain", strings.NewReader(transaction))
 		if err != nil {
 			return err
@@ -74,7 +76,7 @@ func (e ElectrsConnection) VbyteFeeFor25Blocks() (int32, error) {
 		return 0, fmt.Errorf("attempted to call VbyteFee with no apiURL")
 	}
 	var vbyteFee int32
-	err := utils.DoWithDefaultRetry(defaultTimeout, func(ctx context.Context) error {
+	err := utils.DoWithDefaultRetry(e.timeout, func(ctx context.Context) error {
 		resp, err := e.client.Get(fmt.Sprintf("%s/fee-estimates", e.apiURL))
 		if err != nil {
 			return err
@@ -107,7 +109,7 @@ func (e ElectrsConnection) IsAddressUnused(btcAddress string) (bool, error) {
 		return false, fmt.Errorf("attempted to call IsAddressUnused with no apiURL")
 	}
 	isAddressUnused := false
-	err := utils.DoWithDefaultRetry(defaultTimeout, func(ctx context.Context) error {
+	err := utils.DoWithDefaultRetry(e.timeout, func(ctx context.Context) error {
 		resp, err := e.client.Get(fmt.Sprintf("%s/address/%s/txs", e.apiURL, btcAddress))
 		if err != nil {
 			return err
--- a/pkg/chain/bitcoin/electrs.go
+++ b/pkg/chain/bitcoin/electrs.go
@@ -24,28 +24,28 @@ type httpClient interface {
 	Get(url string) (*http.Response, error)
 }
 
-// ElectrsConnection exposes a native API for interacting with an electrs http API.
-type ElectrsConnection struct {
+// electrsConnection exposes a native API for interacting with an electrs http API.
+type electrsConnection struct {
 	apiURL  string
 	client  httpClient
 	timeout time.Duration
 }
 
-// NewElectrsConnection is a constructor for ElectrsConnection.
-func NewElectrsConnection(apiURL string) *ElectrsConnection {
-	return &ElectrsConnection{
+// newElectrsConnection is a constructor for ElectrsConnection.
+func newElectrsConnection(apiURL string) *electrsConnection {
+	return &electrsConnection{
 		apiURL:  apiURL,
 		client:  http.DefaultClient,
 		timeout: defaultTimeout,
 	}
 }
 
-func (e *ElectrsConnection) setClient(client httpClient) {
+func (e *electrsConnection) setClient(client httpClient) {
 	e.client = client
 }
 
 // Broadcast broadcasts a transaction the configured bitcoin network.
-func (e ElectrsConnection) Broadcast(transaction string) error {
+func (e electrsConnection) Broadcast(transaction string) error {
 	if e.apiURL == "" {
 		return fmt.Errorf("attempted to call Broadcast with no apiURL")
 	}
@@ -71,7 +71,7 @@ func (e ElectrsConnection) Broadcast(transaction string) error {
 }
 
 // VbyteFeeFor25Blocks retrieves the 25-block estimate fee per vbyte on the bitcoin network.
-func (e ElectrsConnection) VbyteFeeFor25Blocks() (int32, error) {
+func (e electrsConnection) VbyteFeeFor25Blocks() (int32, error) {
 	if e.apiURL == "" {
 		return 0, fmt.Errorf("attempted to call VbyteFee with no apiURL")
 	}
@@ -106,7 +106,7 @@ func (e ElectrsConnection) VbyteFeeFor25Blocks() (int32, error) {
 
 // IsAddressUnused returns true if and only if the supplied bitcoin address has
 // no recorded transactions.
-func (e ElectrsConnection) IsAddressUnused(btcAddress string) (bool, error) {
+func (e electrsConnection) IsAddressUnused(btcAddress string) (bool, error) {
 	if e.apiURL == "" {
 		return false, fmt.Errorf("attempted to call IsAddressUnused with no apiURL")
 	}
--- a/pkg/chain/bitcoin/electrs.go
+++ b/pkg/chain/bitcoin/electrs.go
@@ -38,10 +38,7 @@ func (e *ElectrsConnection) setClient(client httpClient) {
 // Broadcast broadcasts a transaction the configured bitcoin network.
 func (e ElectrsConnection) Broadcast(transaction string) error {
 	if e.apiURL == "" {
-		for i := 0; i < 5; i++ {
-			logger.Warningf("Please broadcast Bitcoin transaction %s", transaction)
-		}
-		return nil
+		return fmt.Errorf("attempted to call Broadcast with no apiURL")
 	}
 	resp, err := e.client.Post(fmt.Sprintf("%s/tx", e.apiURL), "text/plain", strings.NewReader(transaction))
 	if err != nil {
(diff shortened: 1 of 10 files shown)


pkg/chain/bitcoin/electrs.go
```diff
@@ -89,3 +89,34 @@ func (e ElectrsConnection) VbyteFee() (int32, error) {
 	logger.Info("retrieved a vbyte fee of [%d]", fee)
 	return int32(fee), nil
 }
+
+func (e ElectrsConnection) IsAddressUnused(btcAddress string) (bool, error) {
+	if e.apiURL == "" {
+		return true, nil
+	}
```
nkuba, line 96: This starts getting hairy. Returning `true` here may give us a false-positive result. 
Maybe we shouldn't have this `e.apiURL == ""` path included in all `electrs.go` functions and let it return an error if the connection details are not configured. In such case, we would need to decide in the code calling electrs function if it has the required details to call electrs or not.

Thoughts?

pkg/chain/bitcoin/electrs_test.go
```diff
@@ -90,3 +90,117 @@ func TestElectrsConnection_vbyteFee(t *testing.T) {
...
+					}
+					return &http.Response{
+						StatusCode: testData.responseCode,
+						Body:       ioutil.NopCloser(bytes.NewReader([]byte(testData.response))),
+					}, nil
+				},
+				mockPost: func(url string, contentType string, reader io.Reader) (*http.Response, error) {
+					return nil, nil
+				},
+			})
```
nkuba, line 190: This seems to be shared across the tests, maybe we could extract it to a separate funcion?

pkg/extensions/tbtc/recovery/derive_address.go
```diff
@@ -131,21 +132,22 @@ func ResolveAddress(
...
 	// derive the address.
 	_, err := btcutil.DecodeAddress(beneficiaryAddress, chainParams)
 	if err != nil {
-		addressIndex, err := storage.GetNextIndex(beneficiaryAddress)
+		addressIndex, err := storage.GetNextIndex(beneficiaryAddress, handle)
 		if err != nil {
-			return "", 0, err
+			return "", err
 		}
 		derivedAddress, err := deriveAddress(beneficiaryAddress, addressIndex)
```
nkuba, line 146: This is already being done inside `storage.GetNextIndex`, right?  Could we modify `GetNextIndex` function to `GetNextAddress` and make it return the address instead of the index?

pkg/extensions/tbtc/recovery/storage.go
```diff
@@ -147,11 +149,28 @@ func (dis *DerivationIndexStorage) GetNextIndex(extendedPublicKey string) (uint3
...
-	return uint32(index + 1), nil
+	nextIndex := uint32(lastIndex + 1)
+	numberOfTries := uint32(100)
+	for i := uint32(0); i < numberOfTries; i++ {
+		index := nextIndex + i
+		derivedAddress, err := deriveAddress(strings.TrimSpace(extendedPublicKey), index)
+		if err != nil {
+			return 0, err
+		}
+		ok, err := handle.IsAddressUnused(derivedAddress)
```
nkuba, line 164: We're not handling this `err`. Is it intentional?

pkg/extensions/tbtc/recovery/storage.go
```diff
@@ -147,11 +149,28 @@ func (dis *DerivationIndexStorage) GetNextIndex(extendedPublicKey string) (uint3
...
 	}
 
-	index, err := dis.read(extendedPublicKey)
+	lastIndex, err := dis.read(extendedPublicKey)
 	if err != nil {
 		return 0, err
 	}
-	return uint32(index + 1), nil
+	nextIndex := uint32(lastIndex + 1)
+	numberOfTries := uint32(100)
```
nkuba, line 157: Thoughts around having this property configurable?

pkg/extensions/tbtc/recovery/storage.go
```diff
@@ -147,11 +149,28 @@ func (dis *DerivationIndexStorage) GetNextIndex(extendedPublicKey string) (uint3
 		return 0, err
 	}
 
-	index, err := dis.read(extendedPublicKey)
+	lastIndex, err := dis.read(extendedPublicKey)
 	if err != nil {
 		return 0, err
 	}
-	return uint32(index + 1), nil
+	nextIndex := uint32(lastIndex + 1)
```
nkuba, line 156: Maybe `startIndex`? `nextIndex` suggests to me that we will be increasing it iteratively in the loop.

```suggestion
	startIndex := uint32(lastIndex + 1)
```

pkg/extensions/tbtc/recovery/storage.go
```diff
@@ -147,11 +149,28 @@ func (dis *DerivationIndexStorage) GetNextIndex(extendedPublicKey string) (uint3
...
+		if err != nil {
+			return 0, err
+		}
+		ok, err := handle.IsAddressUnused(derivedAddress)
+		err = dis.save(extendedPublicKey, index)
+		if err != nil {
+			return 0, err
+		}
+		if ok {
+			return index, nil
```
nkuba, line 170: In relation to one of my previous comments, could we return a derived address here?

nkuba (changes requested):

pkg/chain/bitcoin/electrs.go
beaushinkle: the `e.apiURL == ""` is essentially a check to see if the user has explicitly turned off the electrs connection, which is we also have
```
func (e ElectrsConnection) Broadcast(transaction string) error {
	if e.apiURL == "" {
		for i := 0; i < 5; i++ {
			logger.Warningf("Please broadcast Bitcoin transaction %s", transaction)
		}
		return nil
	}
```
and 
```
// VbyteFee retrieves the 25-block estimate fee per vbyte on the bitcoin network.
func (e ElectrsConnection) VbyteFee() (int32, error) {
	if e.apiURL == "" {
		return 0, nil
	}
```

If there is no electrs connection, broadcast the transaction to the warning logs, and return 0 as the vbyte fee.

If there is no electrs connection, then ultimately the client has no way to check whether or not a btc address is unused - they can only rely on the next generated address from storage.

I'm down to re-write how that behavior is expressed! We can either put the defaulting behavior in the electrs code (like how we have it here), or we can move that up a level to the code that calls it. 

Currently in client.go, we have:
```
electrsConnection := bitcoin.NewElectrsConnection(tbtcConfig.Bitcoin.ElectrsURLWithDefault())
...
beneficiaryAddress, err := recovery.ResolveAddress(
  tbtcConfig.Bitcoin.BeneficiaryAddress,
  derivationIndexStorage,
  chainParams,
  electrsConnection,
)
...
vbyteFee, vbyteFeeError := electrsConnection.VbyteFee()
...
broadcastError := electrsConnection.Broadcast(recoveryTransactionHex)
```

We could rewrite that to check the `ElectrsURLWithDefault()` and see if it's `""`, and then make the default decision in the client code, but I'm not sure that's cleaner.

pkg/extensions/tbtc/recovery/storage.go
beaushinkle: I'm down - but I do want to check before I write that up that number of tries is the retry strategy we want to figuring out how long to spend on finding an unused recovery address

pkg/extensions/tbtc/recovery/storage.go
beaushinkle: good catch!

beaushinkle:
Okay, I think I responded to all of the easy review: open questions:

* What behavior do we want for when persistence is really far behind the supplied public key? As in, say that the user gives us an extended key where the first 40 wallet addresses are all dirty. How do we want the program to behave?

* How do we want to draw the separation-of-concerns line with regards to electrs? A user can set `ElectrsURL = ""` in the config explicitly to indicate that they don't want to use an electrs connection (if they don't configure it, we default to `https://blockstream.info/api/`. Should the `ElectrsConnection` be aware of that choice, or should that be higher up, in say, another implementation of the bitcoin handle interface?

pkg/chain/bitcoin/electrs.go
Shadowfiend: I think we should return this status as an error, and match the bool return of any other error state. The expected behavior from the caller is the same: we can't determine whether the address is used or not with this method, use whatever backup approach you want.

I think most likely the correct bool return for an error state is `false`, simply because that's the zero value of `bool`.

Shadowfiend: With that said, I _think_ most of these functions all have the same pattern, and I wonder if we can just abstract that pattern to a single central function. The pattern I'm seeing is:
- If `apiURL` is blank, return error.
- GET a particular address (= one parameter).
- If the status code is non-200, try to extract an error and return it as possible.
- If the status code is 200, JSON-parse.
- Do something with the JSON.

If I'm right about that, it means all but the “do something with the JSON” should be movable to a single function called by each of these, which prevents us from having this weird blank apiURL check in every function.

pkg/chain/bitcoin/handle.go
```diff
@@ -0,0 +1,8 @@
+package bitcoin
+
+// Handle serves as an interface abstraction around bitcoin network queries
+type Handle interface {
+	Broadcast(transaction string) error
+	VbyteFee() (int32, error)
```
Shadowfiend, line 6: We may need to either make this more generic (because it could return several different fees) or make the name more explicit (because it's checking the suggested fee for a particular target confirmation time).

Shadowfiend:
> What behavior do we want for when persistence is really far behind the supplied public key? As in, say that the user gives us an extended key where the first 40 wallet addresses are all dirty. How do we want the program to behave?

IMO it makes sense to continue trying until we find a clean address or we fail to check one for whatever reason.

pkg/chain/bitcoin/electrs.go
beaushinkle: > If apiURL is blank, return error.

None of them currently do this - if the apiURL is blank, they perform the no-electrs behavior. 0 fee for vbyte, send the transactions to the logger for broadcast, return true for validation, all with `nil` for the error.

I can definitely change this across the board though (by returning the nil value and a non-nil error)

beaushinkle:
I pushed up an experimental commit 23919fe32794c91bf5529f46f9b2a8310e6538c3 that tries to break out the no-electrs logic from the electrs code. Let me know if ya'll like that approach better!

pkg/chain/bitcoin/handle.go
beaushinkle: Went with the more specific name. If we ever find that we want fees from blocks other than 25, it's easy to abstract the code at that time

beaushinkle:
anything else ya'll want to see here?

pkg/chain/bitcoin/electrs.go
Shadowfiend: I think changing across the board makes sense. The contract with `bitcoin.Handle` is then “if there is an API response available, I will return it; if not, you will get an error”, and it's the caller's responsibility to default or handle the error some other way.

pkg/chain/bitcoin/electrs.go
```diff
@@ -64,10 +61,10 @@ func (e ElectrsConnection) Broadcast(transaction string) error {
 	return nil
 }
 
-// VbyteFee retrieves the 25-block estimate fee per vbyte on the bitcoin network.
-func (e ElectrsConnection) VbyteFee() (int32, error) {
+// VbyteFeeFor25Blocks retrieves the 25-block estimate fee per vbyte on the bitcoin network.
+func (e ElectrsConnection) VbyteFeeFor25Blocks() (int32, error) {
```
Shadowfiend, line 65: Small note here: `ElectrsConnection` may not need to be exported. If we do a `func Connect(url string) (Handle, error)`, it can return whatever internal handle is appropriate based on the URL.

pkg/chain/bitcoin/handle.go
```diff
@@ -3,6 +3,29 @@ package bitcoin
 // Handle serves as an interface abstraction around bitcoin network queries
 type Handle interface {
 	Broadcast(transaction string) error
-	VbyteFee() (int32, error)
+	VbyteFeeFor25Blocks() (int32, error)
 	IsAddressUnused(btcAddress string) (bool, error)
 }
+
+type OfflineHandle struct{}
```
Shadowfiend, line 10: Looking at this more closely, I'm wondering if it's the right strategy. We can substitute in the `OfflineHandle` if we see the URL is missing, but the caller will still need to handle situations where we have an `ElectrsConnection` _and_ it fails to return correctly from the API. I think these are the same behaviors, in which case having them in one place (the caller) rather than two (the caller + `OfflineHandle`) seems better.

Let me know if I've misunderstood how the pieces are fitting together though.

beaushinkle: Summarizing from an offline conversation:

* We have a couple of choices here. Given that currently, we have a single caller of off of these methods (client.go), and that caller is in charge of any retry/fallback logic, we need to figure out where to draw the boundaries in responsibilities.

As written, the caller (client.go) is responsible for fallback/retry logic, and the bitcoin handle has a low-level single-try mechanism. Instead, we could put the fallback/retry logic in the handle and privatize the single-try stuff.

Both mildly violate YAGNI (you aren't going to need it):

* If we abstract the retry logic into the handler, and move the api calls to a lower-level function, then we're making an assumption that some other caller in the future will need that same retry logic (which we aren't sure of yet).
* If we leave the retry logic in the caller, and preserve the low-level single-try logic, then we're making an assumption that we're going to need a non-retry version of that function call in the future (which we aren't sure of yet).

In the interest of getting code out of the already-very-bloated client.go, I'm going to see how this looks with the retry/fallback logic in the handle

beaushinkle:
Okay, pushed up a pretty spicy commit. Let me know if that's the sort of direction you had in mind

pkg/chain/bitcoin/handle.go
```diff
@@ -1,31 +1,112 @@
...
+	isAddressUnused(btcAddress string) (bool, error)
+}
+
 type Handle interface {
-	Broadcast(transaction string) error
-	VbyteFeeFor25Blocks() (int32, error)
-	IsAddressUnused(btcAddress string) (bool, error)
+	Broadcast(transaction string)
+	VbyteFeeFor25Blocks() int32
+	IsAddressUnused(btcAddress string) bool
```
Shadowfiend, line 27: Some quick thoughts:
- I think we should still return errors rather than logging in this layer. My take on the separation is “let the handle do its very best to get the right answer from the Bitcoin server, let the caller do what it wants if the handle fails after trying”.
- I wonder if we need the `offlineHandle` or even `internalHandle` at all at this juncture. I'm thinking back to my suggestion that the handle's query structure is similar across each of its three calls (each one being an HTTP request and a response or error; this is now extended by an HTTP request _with retries_ and a response or error). It feels like we can implement each `Handle` function as a wrapper that handles the HTTP request stuff, then does whatever the particular function needs to do with the response. That common wrapper can also handle `if apiUrl == "" { return ...error... }`. On the surface, this feels simpler to me now, for what is currently a fairly simple problem to solve. If we later introduce multiple handles for different server types and such that may shift, but right now it's not clear that we need it.

beaushinkle:
we're green! anything else either of ya'll want to see here?

pkg/chain/bitcoin/electrs.go
nkuba: @beaushinkle what we decided to do about this comment? I can see we've defined a `bitcoin.Handle`, can we return it from `NewElectrsConnection` function and make `ElectrsConnection` non-exported?

pkg/chain/bitcoin/electrs.go
```diff
@@ -38,54 +47,98 @@ func (e *ElectrsConnection) setClient(client httpClient) {
...
 	}
-	resp, err := e.client.Get(fmt.Sprintf("%s/fee-estimates", e.apiURL))
+	var vbyteFee int32
+	err := utils.DoWithDefaultRetry(e.timeout, func(ctx context.Context) error {
+		resp, err := e.client.Get(fmt.Sprintf("%s/fee-estimates", e.apiURL))
+		if err != nil {
+			return err
+		}
+		if resp.StatusCode != 200 {
+			return fmt.Errorf("something went wrong with broadcast: [%s]", resp.Status)
```
nkuba, line 85: The error message has to be updated, it's not a broadcast function.

pkg/chain/bitcoin/electrs.go
```diff
@@ -38,54 +47,98 @@ func (e *ElectrsConnection) setClient(client httpClient) {
...
+	if e.apiURL == "" {
+		return false, fmt.Errorf("attempted to call IsAddressUnused with no apiURL")
 	}
+	isAddressUnused := false
+	err := utils.DoWithDefaultRetry(e.timeout, func(ctx context.Context) error {
+		resp, err := e.client.Get(fmt.Sprintf("%s/address/%s/txs", e.apiURL, btcAddress))
+		if err != nil {
+			return err
+		}
+		if resp.StatusCode != 200 {
```
nkuba, line 117: In `Broadcast` and `VbyteFeeFor25Blocks` functions we return an error right after checking `resp.StatusCode != 200`. Here we're trying to resolve the error message. Is there a reason we have two different approaches? Could we align them all?

pkg/chain/bitcoin/handle.go
```diff
@@ -0,0 +1,8 @@
+package bitcoin
+
+// Handle serves as an interface abstraction around bitcoin network queries
+type Handle interface {
+	Broadcast(transaction string) error
```
nkuba, line 5: Side note: Would it be worth returning a transaction ID from the `Broadcast` function?

pkg/utils/wrappers.go
```diff
@@ -20,19 +20,20 @@ func DoWithRetry(
 	ctx, cancel := context.WithTimeout(context.Background(), timeout)
 	defer cancel()
 
+	var err error
 	for {
 		select {
 		case <-ctx.Done():
-			return fmt.Errorf("retry timeout [%v] exceeded", timeout)
+			return fmt.Errorf("retry timeout [%v] exceeded: [%v]", timeout, err)
```
nkuba, line 27: In some cases, the err will be nil and in the other, it will be an error returned on the last retry. I'm not sure about including this error in the timeout error message as it's not strictly the error that caused the failure.
Maybe we should leave logging the intermediate errors to the `doFn` function implementation?

nkuba (changes requested):

pkg/chain/bitcoin/electrs.go
beaushinkle: For broadcast and vbyte fee, the errors are shipped back in the status, so all we need to read is the status. `IsAddressUnused` has a more complicated API because there's more cases it has to handle (like sending an invalid btc address in the url param). The API ships those errors back in the payload along with a non-200 status code.

If you try to hit `localhost:3002/address/banana/tx`, for example, you get a 400 response with a payload of "Invalid Bitcoin address", and that error doesn't show up in the status.

pkg/chain/bitcoin/handle.go
beaushinkle: I think that has the potential to be useful, but since we wouldn't be doing anything with it at the moment, I'd prefer to leave it out until we have a concrete use case

pkg/utils/wrappers.go
beaushinkle: The motivation here was that `utils.DoWithDefaultRetry` gobbles up useful error message, which breaks the `IsAddressUnused` tests in electrs_test.go.

What if I changed the verbiage to "most recent error: [%v]"? 

That way, if it was a timeout on the first attempt, we see: 
```
retry timeout [2m0s] exceeded. most recent error: [nil]
```
and otherwise we see something like
```
retry timeout [2m0s] exceeded. most recent error: [json: cannot unmarshal number into Go value of type []interface {}]
```

nkuba: Ok, let's replace the error message to: `"retry timeout [%v] exceeded; most recent error: [%v]"`

pkg/chain/bitcoin/handle.go
nkuba: We could log the ID on the caller level instead of doing so inside the implementation.

beaushinkle:
> What we would need to do here is

Oh! I didn't know that works! That's super cool :heart:

nkuba (changes requested):

beaushinkle:
Okay - good for another look! Let me know if you're okay with the workaround in electrs_test

nkuba (approved):
Let's ship it!

PR #788, opened 2021-05-12
Configure The Liquidation Recovery Timeout

beaushinkle:
Previously, we were using the signing timeout to serve as the broader liquidation timeout (not intentionally, just a missed detail). This PR adds a new configurable timeout, `Extensions.TBTC.LiquidationRecoveryTimeout`, that governs how long we'll attempt recovery for before giving up.

Tagging @nkuba for review

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -91,6 +91,10 @@ func TestReadConfig(t *testing.T) {
 			readValueFunc: func(c *Config) interface{} { return c.Extensions.TBTC.TBTCSystem },
 			expectedValue: "0xa4888eDD97A5a3A739B4E0807C71817c8a418273",
 		},
+		"Extensions.TBTC.LiquidationRecoveryTimeout": {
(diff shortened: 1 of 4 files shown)


config/config_test.go
```diff
@@ -91,6 +91,10 @@ func TestReadConfig(t *testing.T) {
 			readValueFunc: func(c *Config) interface{} { return c.Extensions.TBTC.TBTCSystem },
 			expectedValue: "0xa4888eDD97A5a3A739B4E0807C71817c8a418273",
 		},
+		"Extensions.TBTC.LiquidationRecoveryTimeout": {
```
beaushinkle, line 94: I was split here between putting it in `TBTC` directly or `TBTC.Bitcoin`

nkuba: I like it better here 💯

nkuba (approved):

PR #789, opened 2021-05-13
Recalculate ECDSA staker rewards: 2021-04-30 -> 2021-05-07

lukasz-zimnoch:
Here we recalculate staker rewards for the interval `2021-04-30 -> 2021-05-07` published as part of https://github.com/keep-network/keep-ecdsa/pull/778

Recalculation of rewards is caused by a corner case that occurred in the aforementioned interval and was not covered by the staker rewards script. Such a case occurred for operators who completed the undelegation period (60 days) but still had their stakes locked, due to active keep memberships. In that case, the staker rewards script considered their `keepStaked` parameter to be zero and recognized the `minimumStakeAtStart` requirement as violated. Those two factors zeroed the amount of the final reward. This pull request improves the script logic. Specifically:
- for `keepStaked` parameter: the script does not just check the active stake of an operator against the operator contract (`BondedKeepECDSAFactory`). In case the active stake value is zero, it also checks the stake locks to find active keeps and executes an active stake check against one of the active keeps.
- for `minimumStakeAtStart` requirement: the script omits that requirement in case an operator is undelegating (similarly to the `poolRequirementFulfilledAtStart` requirement)


staker-rewards/lib/assets-calculator.js
```diff
@@ -124,14 +124,48 @@ export default class AssetsCalculator {
...
+    // whether it still has a locked active stake in one of the keep contracts.
+    // If so, we count that stake as keep staked.
+    if (keepStaked === "0") {
+      const locks = await callWithRetry(
+        this.tokenStaking.methods.getLocks(operator),
+        block
+      )
+
+      for (let i = 0; i < locks.creators.length; i++) {
+        if (timestamp < locks.expirations[i]) {
```
pdyraga, line 147: I don't think we should take into account expirations. Even if the lock expired, a deposit can still be active.

lukasz-zimnoch: I think we have no choice and we have to check expiration. If the lock expires, activeStake for that keep will return 0, no matter the underlying keep is active or not

pdyraga: This will probably lead to another corner case. We do not need to solve it in this PR but let's work on it right after this one is merged. It does make sense from a contract perspective to do not mark stake as locked if the lock expired but it is still possible that some deposits were not redeemed after 6 months and ETH is bonded. I think we should add a check that if there is a lock and that lock expired, we fallback to `TokenStaking.balanceOf`.

pdyraga (approved):

staker-rewards/lib/assets-calculator.js
lukasz-zimnoch: Addressed in the follow-up PR: https://github.com/keep-network/keep-ecdsa/pull/790

PR #794, opened 2021-05-20
Compile contracts before publishing

michalinacienciala:
`NPM` workflow will be used to publish the contracts needed for the
setup of development environment to NPM registry. Before the change we
were not compiling the contracts before publishing, which was a problem,
because this way not all the files needed by the downstream workflows
were being published. As a result of contracts compilation system will
generate `*.json` files needed by the later workflows. We will then
place them in the `artifacts` directory. By the `package.json`
configuration this directory is being picked by the publish step along
with the `contracts/**/*.sol` files.
Apart from this change, couple of small updates were introduced to `NPM`
workflow, mainly update to higher versions of used external actions.

Corresponding changes in `keep-core` project: https://github.com/keep-network/keep-core/pull/2474


michalinacienciala:
TODO: once I get the OK for rest of the proposed changes, I will remove the pieces of code marked with `# TODO`

michalinacienciala:
> TODO: once I get the OK for rest of the proposed changes, I will remove the pieces of code marked with `# TODO`

TODOs removed. Here is the last run before that change, with --dry-run in the publish step: https://github.com/keep-network/keep-ecdsa/actions/runs/870829610

nkuba (approved):

.github/workflows/npm.yml
```diff
@@ -30,49 +34,35 @@ jobs:
...
-
-          # Check resolved `preid`. Currently only `pre` value is supported,
-          # other types of releases are not handled by this job.
-          if [ "$preid" != pre ]; then
-            echo "Unsupported preid. Resolved info:"
-            echo "$name@$version ; preid $preid"
-            exit 1
-          fi
+          npm update \
+            @keep-network/keep-core \
```
nkuba, line 45: We should also update `package.json` file to reflect the `dev` versions ranges.

nkuba (changes requested):

.github/workflows/npm.yml
michalinacienciala: Right, this is going to be delivered in separate PR: https://github.com/keep-network/keep-ecdsa/pull/797.

PR #796, opened 2021-05-26
Make sed invocations cross-platform

michalinacienciala:
Syntax of the `sed -i` command is different for MacOS and for GNU
systems. The command in its old shape was failing when run on Ubuntu
(during execution of workflow with E2E tests in the `local-setup`
repository). Fixed by doing replacement using temporary file, instead of
replacing in the current file.


michalinacienciala:
I've tested the command on ubuntu in this temporarily created workflow: https://github.com/michalinacienciala/make-docs-action/runs/2674406141?check_suite_focus=true.

nkuba (approved):

PR #803, opened 2021-06-01
Log transaction hash as %s instead of %x

nkuba:
The `Hash` type has a dedicated function to convert the value to string.
It is converted to hex with the `0x` prefix, which seems more consistent for ethereum.

In keep-common code generation we generate the loggers for transaction id with `Hash().Hex()` function which produces `0x` prefixed values. (see: https://github.com/keep-network/keep-common/blob/0d4f5d5ec604ac41aa84b9ef8078038d7592fc43/tools/generators/ethlike/contract_non_const_methods.go.tmpl#L72)

With this PR we will be more consistent across the logs.


pkg/chain/gen/celo/contract/BondedECDSAKeep.go
```diff
@@ -238,7 +238,7 @@ func (becdsak *BondedECDSAKeep) DistributeERC20Reward(
 ) (*types.Transaction, error) {
 	becdsakLogger.Debug(
```
beaushinkle, line 239: Was this written before we started using `Debugf`? Want to change it or leave it in?

beaushinkle: oh wait this fine is generated @_@

beaushinkle (approved):

PR #804, opened 2021-06-02
Removed println for signer loading

nkuba:
The println we're removing here is probably a leftover from development.
After the loading is done we already have following loggers:
info: "loaded [%d] keeps from the local storage"
debug: "loaded signer for keep [%s]"


nkuba:
@Shadowfiend could you please confirm that the Println I removed wasn't left intentionally?

Shadowfiend (approved):
Nope, def just a silly me. Looks like the commit that introduced this one left a few printlns in 🙈

PR #805, opened 2021-06-02
Include remote peer ID in firewall error messages

nkuba:
When firewall validation fails for a given peer we want to include the peer ID in the error message. The peer ID is expected to be the ethereum address for the ethereum chain. 
Having this value in the error simplifies problem debugging.


pkg/firewall/firewall_test.go
```diff
@@ -141,39 +126,31 @@ func TestConsultsAuthorizedOperatorsCache(t *testing.T) {
...
 
-	err = policy.validateAuthorization(localChain.PublicKeyToOperatorID(
-		(*ecdsa.PublicKey)(remotePeer2PublicKey),
-	))
-	if err != errNoAuthorization {
-		t.Errorf("expected error about no authorization; has: [%v]", err)
+	if err := policy.validateAuthorization(remotePeer2ID); err.Error() != expectedError {
+		t.Errorf(
+			"unexpected error about no authorization for remote peer 2\n"+
+				"actual:   [%v]\nexpected: [%v]",
```
beaushinkle, line 150: I like having `[]` in the log messages - it lets a human know when something is variable, and it makes it easier to write programs that parse the logs.

I'm not sure that I like having `[]` in the test error classes though. It feels like:
```
unexpected error about no authorization for remote peer 2
actual:   banana
expected: fofana
```
is easier to read than
```
unexpected error about no authorization for remote peer 2
actual:   [banana]
expected: [fofana]
```

That said, this isn't scoped to this PR, so I won't block on it!

beaushinkle: Also, in the liquidation recovery stuff, we always put `expected` before `actual`:

```go
			if err == nil {
				t.Errorf(
					"unexpected error message\nexpected: %s\nactual:   nil",
					testData.err,
				)
```

beaushinkle:
Okay - I'm good here. I left some non-blocking review, so depending on how that goes we can add some backlog cards, but merge in the meantime.

beaushinkle (approved):

PR #809, opened 2021-06-03
More logging for TBTCSystem address configuration validations

nkuba:
We enhance coverage of cases that cause warnings to be logged on TBCSystem address configuration validations.

The address is expected to be configured under `config.Ethereum.ContractAddresses`.

Currently, if the address is configured in
 - both `config.Ethereum.ContractAddresses.TBTCSystem` and `config.Extensions.TBTC.TBTCSystem`
 - both `config.Ethereum.ContractAddresses.TBTCSystem` and `config.SanctionedApplications.Addresses`
we won't log a warning. With this change, we will log a warning always if the address is configured under a deprecated path to let the operator knows about the needed configuration file cleanup.

We added a check for config.Ethereum.ContractAddresses.TBTCSystem and `config.Extensions.TBTC.TBTCSystem` match to panic on mismatch. We already have similiar check for `config.SanctionedApplications`.


cmd/connector_ethereum.go
```diff
@@ -50,35 +50,53 @@ func connectChain(
...
+		}
 	}
 
 	// DEPRECATED: config.Ethereum.ContractAddresses is the correct container
 	// for the TBTCSystem address from now on; read SanctionedApplications and
 	// assume it has a single entry that is TBTCSystem, warn if
 	// SanctionedApplications needs to be used.
 	applicationAddresses := config.SanctionedApplications.AddressesStrings
-	if !exists && len(applicationAddresses) == 1 {
+	if len(applicationAddresses) == 1 {
```
nkuba, line 88: I'm not sure why did we have the difference between this check (`len(applicationAddresses) == 0`) and the next one (L102: `len(applicationAddresses) > 0`). Shoudln't we align them both to be `len(applicationAddresses) > 0`?

Shadowfiend: I think my thinking was that only with a 1-length application address can we have high confidence that the address in question is the TBTCSystem address, and thus warn. Without it, we don't know what they've thrown in here.

That said, I also think I overthought it haha. Aligning makes sense. The only additional sanctioned application we might expect is the original tBTC deploy, but treating that as equally problematic from a config perspective makes sense.

beaushinkle (approved):

beaushinkle:
Love it :heart:

Issue #811, opened 2021-06-03
Liquidation recovery crashed with panic


nkuba:
Clients crashed with a panic ~~on liquidation recovery. The case that I wanted to test is a mismatch between provided `BeneficiaryAddress` and `BitcoinChainName`.~~
EDIT: The problem occurred also when I provided `BitcoinChainName = "regtest"` in the config. So previous cause of the failure is not the real one.

I provided the address generated for `regtest` network and left `BitcoinChainName` empty. In such a case, `BitcoinChainName` resolves to a default which is `mainnet`.

I propose that we implement validation for parameters provided in the config to catch possible missconfigurations.

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x4a151d9]

goroutine 7038 [running]:
github.com/keep-network/keep-ecdsa/pkg/client.monitorKeepTerminatedEvent.func1.1.1(0x5083228, 0xc0008373e0, 0x0, 0x0)
	/Users/jakub/workspace/keep-network/local-setup/keep-ecdsa/pkg/client/client.go:1154 +0x14b9
github.com/keep-network/keep-ecdsa/pkg/utils.DoWithRetry(0x3b9aca00, 0x1bf08eb000, 0xa06f5356a000, 0xc0002bdec8, 0x0, 0x0)
	/Users/jakub/workspace/keep-network/local-setup/keep-ecdsa/pkg/utils/wrappers.go:29 +0x146
github.com/keep-network/keep-ecdsa/pkg/utils.DoWithDefaultRetry(...)
	/Users/jakub/workspace/keep-network/local-setup/keep-ecdsa/pkg/utils/wrappers.go:66
github.com/keep-network/keep-ecdsa/pkg/client.monitorKeepTerminatedEvent.func1.1(0xc00000c378, 0xc000033700, 0x5095638, 0xc00081e180, 0x50930e8, 0xc0003f8380, 0xc00040a390, 0xc000351e60, 0xc000113170, 0x508d5d8, ...)
	/Users/jakub/workspace/keep-network/local-setup/keep-ecdsa/pkg/client/client.go:991 +0x197
created by github.com/keep-network/keep-ecdsa/pkg/client.monitorKeepTerminatedEvent.func1
	/Users/jakub/workspace/keep-network/local-setup/keep-ecdsa/pkg/client/client.go:990 +0x2be
```

### Configuration:

- client 1
```toml
[...]

PR #814, opened 2021-06-04
Fix panic on liquidation recovery transaction broadcast failure

nkuba:
In this PR we removed logging of ElectrsURL in case of electrs function failure.

we don't need to include the Electrs URL in the logged errors. The address is configured by the operator in the config file and can be checked there. If the address is provided with an empty string the electrs api functions will return an error saying that it's empty.

Closes: https://github.com/keep-network/keep-ecdsa/issues/811

### DIFF
--- a/pkg/chain/bitcoin/electrs.go
+++ b/pkg/chain/bitcoin/electrs.go
@@ -31,7 +31,7 @@ type electrsConnection struct {
 	timeout time.Duration
 }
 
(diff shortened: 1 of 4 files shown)


beaushinkle:
We're getting some new linter errors :o

beaushinkle:
This looks good to me - down to merge as soon as the linter is happy

PR #817, opened 2021-06-08
Upgraded Golang to 1.16

nkuba:
We upgraded Golang to the latest version 1.16.

For the docker image, we use alpine 3.12, while the latest available source image is based on 3.13. We use an older alpine version as it comes with Node 12 which is required by the solidity package. Alpine 3.13 comes with Node 14 which doesn't work well for us yet.

We also install `python3`, as `python` is no longer available in the package registry.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.13.8-alpine3.10 AS gobuild
+FROM golang:1.16.5-alpine3.12 AS gobuild
 
 # HOST_CHAIN argument defines the chain implementation which should be used
 # during image build process.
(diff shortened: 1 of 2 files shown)


beaushinkle (approved):

beaushinkle:
Everything looks to be in order, and the code looks spotless.

I had a look through the [go documentation for specifying a go version](https://golang.org/doc/modules/gomod-ref#go-syntax):

> For packages within the module, the compiler rejects use of language features introduced after the version specified by the go directive. For example, if a module has the directive go 1.12, its packages may not use numeric literals like 1_000_000, which were introduced in Go 1.13.

> If an older Go version builds one of the module's packages and encounters a compile error, the error notes that the module was written for a newer Go version. For example, suppose a module has go 1.13 and a package uses the numeric literal 1_000_000. If that package is built with Go 1.12, the compiler notes that the code is written for Go 1.13.

This PR, then, changes how the docker operates (since we use 1.16 there now), and the tests run in the docker. For developers, we'll all have to individually update our go versions (and will receive a error once we start using 1.16 features).

PR #822, opened 2021-06-10
Temp fix for the problem with failing jobs - move cache instead of copying

michalinacienciala:
We've seen a number of workflows failing with `No space left on device`
error (e.g.
https://github.com/keep-network/keep-ecdsa/actions/runs/924123923).
The problem was a combination of a big cache size and significant memory
allocation after building Docker image.
According to https://github.com/moby/buildkit/issues/1850:
"At the moment caches are copied over the existing cache so it keeps
growing".
As a temporary fix (until issue gets fixed by GH), we introduce a step
that moves the cache (similarily to how it's described in
https://github.com/moby/buildkit/issues/1896).

The solution may not keep us from failures forever, as we still may
hit the problems with the full memory if the size of the Docker image
or cache increases (see this FD discussion:
https://www.flowdock.com/app/cardforcoin/tech/threads/Clml1_lI4HYRKrr3N_vcSInDn6A).

### DIFF
--- a/.github/workflows/client-celo.yml
+++ b/.github/workflows/client-celo.yml
@@ -80,7 +80,7 @@ jobs:
             HOST_CHAIN=celo
           load: true # load image to local registry to use it in next steps
           cache-from: type=local,src=/tmp/.buildx-celo-cache
-          cache-to: type=local,dest=/tmp/.buildx-celo-cache
+          cache-to: type=local,dest=/tmp/.buildx-celo-cache-new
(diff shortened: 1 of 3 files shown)


beaushinkle:
Given that this is an exact port over from https://github.com/docker/build-push-action/blob/master/docs/advanced/cache.md#github-cache, I think we're good here!

beaushinkle (approved):

PR #815, opened 2021-06-04
Fail Fast With Bad BTC Configuration

beaushinkle:
If a user has not configured `Extensions.TBTC.Bitcoin` with the required fields (BeneficiaryAddress and BitcoinChainName presently), this PR changes the client behavior in the following ways:

* We warn at startup about the paths that are missing, and suggest possible values:
```
2021-06-04T11:11:26.586-0400    WARN    keep-cmd        invalid Bitcoin configuration: [a bit
coin address or extended public key (*pub) is required; configure one at [Extensions.TBTC.Bitc
oin.BeneficiaryAddress]]
2021-06-04T11:11:26.586-0400    WARN    keep-cmd        invalid Bitcoin configuration: [a valid chain name is required; choose between [mainnet, regtest, simnet, testnet3] and configure it at [Extensions.TBTC.Bitcoin.BitcoinChainName]: [unable to find chaincfg param for name: [banana]]]
```

* We error out and skip the retry loop during termination handling:
```
2021-06-04T10:29:13.005-0400    ERROR    keep-ecdsa    failed to broadcast the bitcoin recovery transaction: [a bitcoin address or extended public key (*pub) is required; configure one at [Extensions.TBTC.Bitcoin.BeneficiaryAddress]]
2021-06-04T10:29:13.005-0400    ERROR    keep-ecdsa    failed to broadcast the bitcoin recovery transaction: [a valid chain name is required; choose between [mainnet, regtest, simnet, testnet3] and configure it at [Extensions.TBTC.Bitcoin.BitcoinChainName]: [unable to find chaincfg param for name: [banana]]]
```

closes #807 

tagging @nkuba for review

PR #826, opened 2021-06-11
Resolve btc address based on btcutil encoding

nkuba:
An address provided as a beneficiary address can match one of many bitcoin address formats. One of which is P2PK where a public key is provided.

Previously we would return the same public key that was supplied by the user, but `btcutil.DecodeAddress` resolves a correct address for this public key, so we want to use the address resolved by the `btcutil`.

Here we updated unit tests to cover the scenario described above but also enhanced the tests a little bit with some test cases taken from: https://github.com/btcsuite/btcutil/blob/e2ba6805a89060647fa22ec61b799acff6a4f6fd/address_test.go

Depends on: https://github.com/keep-network/keep-ecdsa/pull/815


beaushinkle:
Gorgeous. I'll merge when 815 goes in

nkuba:
@beaushinkle it's ready to be merged.

Issue #801, opened 2021-05-31
Staker rewards: take into account bonded ETH for operators who deauthorized the pool


pdyraga:
An authorizer can deauthorize sortition pool. In this case, operator is still required to keep their node online and ETH is still bonded in all active deposits for that particular operator:
https://github.com/keep-network/keep-ecdsa/blob/f1625856b94887ac36c2fa33f5b853b65a715296/solidity/contracts/AbstractBonding.sol#L287-L299

In #667 we updated the staker rewards mechanism to allocate rewards for undelegating operators but only count ETH that is bonded in ETH_total parameter of the equation. It feels that the same approach should be used for operators whose authorizers deauthorized sortition pool. They need to keep the node online and their ETH is still bonded in all active deposits they are backing:

https://github.com/keep-network/keep-ecdsa/blob/835d3c31254d424def0252381b002ed49a427a43/staker-rewards/lib/rewards-calculator.js#L140-L145

I propose we alter the rules to treat sortition pool unauthorized operators the same way as undelegating operators.

PR #828, opened 2021-06-13
Staker rewards: Deauthorized operators

lukasz-zimnoch:
Closes: #801 

An authorizer can deauthorize the sortition pool. In this case, the operator is still required to keep their node online and ETH is still bonded in all active deposits for that particular operator. Because of that, the staker rewards mechanism will allocate rewards for deauthorized operators but only count ETH that is bonded in `ETH_total` parameter of the equation. Also, `poolDeauthorizedInInterval` requirement will not be considered violated if `ETH_bonded > 0`

This PR also improves the way `ETH_total` is calculated for undelegating and deauthorized operators. Instead of subtracting `ETH_withdrawn` from `ETH_bonded`, we take the smaller value from between `ETH_bonded` at interval start and `ETH_bonded` at interval end. Those calculations are more fair for those who did not redeem anything and they have just withdrawn their unbonded ETH that could not be used by the system anyway.


pdyraga:
Let's remember about updating README.

staker-rewards/lib/rewards-calculator.js
```diff
@@ -129,7 +129,10 @@ export default class RewardsCalculator {
       violations.push("poolAuthorizedAtStart")
     }
 
-    if (requirements.poolDeauthorizedInInterval === true) {
+    if (
+      requirements.poolDeauthorizedInInterval === true &&
+      operatorAssets.ethBonded.isEqualTo(new BigNumber(0))
+    ) {
```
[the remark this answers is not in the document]
lukasz-zimnoch: You're right but I think this way we communicate explicitly why there is no reward. If we detected deauthorization and no ETH is bonded, it will be clearly visible in the report.

staker-rewards/test/data/test-blockchain.json
```diff
@@ -241,6 +197,81 @@
...
+            "name": "BondCreated",
+            "blockNumber": 2000,
+            "returnValues": {
+              "referenceID": "0x64f68c710615F2c86c2d224A1101112076FfcB7E",
+              "operator": "0x88Ed51f84c21Ae246c23137D090cdF441009D916",
+              "sortitionPool": "0xA3748633c6786e1842b5cC44fa43db1ecC710501",
+              "amount": "10000000000000000000"
+            }
+          },
           {
```
pdyraga, line 275: Should `UnbondedValueWithdrawn` events below have `blockNumber` set?

staker-rewards/test/data/test-blockchain.json
```diff
@@ -241,6 +197,81 @@
...
+          },
+          {
+            "name": "BondCreated",
+            "blockNumber": 2000,
+            "returnValues": {
+              "referenceID": "0x64f68c710615F2c86c2d224A1101112076FfcB7E",
+              "operator": "0x88Ed51f84c21Ae246c23137D090cdF441009D916",
+              "sortitionPool": "0xA3748633c6786e1842b5cC44fa43db1ecC710501",
+              "amount": "10000000000000000000"
+            }
```
pdyraga, line 273: `0x88Ed51f84c21Ae246c23137D090cdF441009D916` is undelegating so they could not have `BondCreated` at the end block of the interval. If we were considering only undelegating operators, we could take ETH_locked as ETH_bonded at T_end. But we have sortition pool deauthorization as well, and this operation can be reverted by authorizing the pool again, that's why we need the minimum of ETH_bonded at the start block and at the end block. I'd rather have `BondCreated` event emitted for `0xF1De9490Bf7298b5F350cE74332Ad7cf8d5cB181` at block 2000 because this could happen in a real world.

staker-rewards/test/rewards-calculator.test.js
```diff
@@ -241,25 +241,55 @@ describe("rewards calculator", async () => {
     assert.equal(rewards.totalRewards.isEqualTo(new BigNumber(0)), true)
   })
```
pdyraga, line 243: There is a "should set total rewards to zero if pool is not authorized at start" test at line 224 which I think we need to delete.

staker-rewards/test/data/test-blockchain.json
lukasz-zimnoch: Nope, that field is used just for `Bond*` events.

pdyraga: Still confusing in this whole structure but OK, we do not need to fix it now, the change in this PR is completely unrelated.

staker-rewards/test/data/test-blockchain.json
lukasz-zimnoch: This is just for testing purposes to make sure the logic works well. However, I'll make it more "real-world".

staker-rewards/test/data/test-blockchain.json
lukasz-zimnoch: It's because we simulate events in a very simplified way in tests. Treat the events under key `"blockNumber": 2000` as "events which would be returned if I invoke `getPastEvents` with `toBlock: 2000`". Then, `blockNumber` parameter of each event determines the block in which a given event has been emitted. However, the `blockNumber` is not needed everywhere. It's needed for events that are filtered by the block number. Simulating a real mechanism will be an overkill for such simple tests.

staker-rewards/test/rewards-calculator.test.js
lukasz-zimnoch: Nope, we didn't touch the `poolAuthorizedAtStart` requirement in this PR. We touched `poolDeauthorizedInInterval`

staker-rewards/test/data/test-blockchain.json
lukasz-zimnoch: https://github.com/keep-network/keep-ecdsa/pull/828/commits/17a9bc7a429af318e1d95c1f573658dba1eac732

staker-rewards/test/data/test-blockchain.json
pdyraga: Yeah, that makes sense and I fully support it. What I am trying to say is that we could have block numbers on those events just for informational purposes that they happened between 1000 and 2000 block so that reading this JSON is easier when we try to match the event history so that we can evaluate numbers for assertions (this is what I tried to do).

staker-rewards/test/data/test-blockchain.json
pdyraga: Should we have `BondCreated` emitted for `0x88Ed51f84c21Ae246c23137D090cdF441009D916` so that we can assert `min()` was taken into account? Or is it already covered somewhere and I can't find it?

lukasz-zimnoch: It's emitted in line `222`.

staker-rewards/test/rewards-calculator.test.js
pdyraga: Errr... wait. So say that I deauthorized sortition pool in this interval. I am getting rewards for this interval because I still need to run a node and I have my ETH locked. This interval ends and a new one starts. Am I getting rewards in the new interval?

lukasz-zimnoch: Nope,  because you will violate the `poolAuthorizedAtStart` requirement in the next interval. But in this PR we are discussing what happens if someone violates `poolDeauthorizedInInterval` during the current interval.

pdyraga: Will there be another PR for issue #801? From the PR description, I assumed this is the only one closing the issue.

staker-rewards/test/data/test-blockchain.json
pdyraga: Actually, the `BondReleased` at line 258 emitted at block 2000 does the min testing for `0xF1De9490Bf7298b5F350cE74332Ad7cf8d5cB181` with unauthorized pool:
Bond 20 ETH created at 1000
Bond 5 ETH created at 1000
Bond 35 ETH created at 1000
Bond 20 ETH released at 1000
Bond 15 ETH seized at 1000
Bond 5 ETH released at 2000

At block 1000, they have 20+5+35-20-15 = 25
At block 2000, they have 25 - 5 = 20
min(25,20)=20
... and this is what is asserted in the test.

All good.

staker-rewards/test/rewards-calculator.test.js
lukasz-zimnoch: Ok, I was sure we want to block rewards for deauthorized operators in the next intervals. Here is the change: https://github.com/keep-network/keep-ecdsa/pull/828/commits/b4bb8deae13d0c3d49ae6f571970fd956a5bac05

pdyraga: Let's update README as well.

pdyraga:
Reviewed, I am in. Let's update the README and we should be good to go.

staker-rewards/README.adoc
```diff
@@ -33,9 +33,11 @@ For each Operator:
...
   
 * Calculate the total ETH under management as 
-  `ETH_locked = ETH_bonded + ETH_unbonded - ETH_withdrawn`.
-  In case the operator is undelegating or has been de-authorized for the
-  sorition pool, `ETH_locked = min(ETH_bonded_T_start, ETH_bonded_T_end)`.
+  `ETH_locked = ETH_bonded + ETH_unbonded - ETH_withdrawn`. In case the operator
+  is undelegating OR
+  has been de-authorized for the sorition pool in current interval OR
+  is not authorized for the sortition pool at the interval start
+  their `ETH_locked = min(ETH_bonded_T_start, ETH_bonded_T_end)`.
```
pdyraga, line 40: We also need to update requirements below:
- If the operator does not have BondedECDSAKeepFactory contract or tBTC sortition pool authorized at T_start, they are not getting any rewards for interval T.
 - If the operator deauthorized tBTC sortition pool between T_start and T_end, they are not getting any rewards for interval T.

into:

- If the operator does not have BondedECDSAKeepFactory contract authorized at T_start, they are not getting any rewards for interval T
- if the operator does not have tBTC sortition pool authorized at T_start and ETH_bonded = 0 for that operator, they are not getting any rewards for interval T.

lukasz-zimnoch: https://github.com/keep-network/keep-ecdsa/pull/828/commits/0db6c14858f3d2e9bf2be5068b6c84df4be138cf

pdyraga (approved):

PR #833, opened 2021-06-16
Validate Bitcoin addresses received from peer members

nkuba:
We want to validate Bitcoin addresses received from peer members on liquidation recovery to confirm the addresses are valid for the network the client is supporting. Without this validation, the client may accept an address from a misconfigured peer which was desired for a different network.

The validation here differs from `ValidateAddress` defined in `derive_address.go` as here we don't expect to receive *pub keys from peers, as this is the peer's duty to derive a bitcoin address from *pub key before broadcasting it.


pkg/ecdsa/tss/protocol_recovery.go
```diff
@@ -161,24 +161,51 @@ func BroadcastRecoveryAddress(
...
+		)
+	}
+
+	if !decodedAddress.IsForNet(chainParams) {
+		return fmt.Errorf(
+			"address [%s] is not a valid btc address for chain [%s]",
+			btcAddress,
+			chainParams.Name,
+		)
+	}
```
beaushinkle, line 208: There's a lot of overlap logic-wise with `bitcoin.ValidateAddress`. I think we could switch the ordering of `ValidateAddress` around so that it first attempts to derive and then if that fails, to decode via this new function.

That way we don't end up expressing the "is this raw btc address valid for the chain" logic twice

nkuba: Good point!

pkg/chain/bitcoin/derive_address.go
```diff
@@ -122,29 +122,44 @@ func DeriveAddress(extendedPublicKey string, addressIndex uint32) (string, error
 	return finalAddress.EncodeAddress(), nil
 }
 
-// ValidateAddress checks to see if the supplied btc address is valid on the
+// ValidateAddressOrKey checks to see if the supplied btc address is valid on the
```
beaushinkle, line 125: awesome - this is exactly what I was envisioning

beaushinkle (approved):

PR #837, opened 2021-06-18
Pass configured chainParams to DeriveAddress instead of resolving them from *pub key

nkuba:
When we were deriving address from *pub extended keys we used to
determine chain network type based on the key prefix. It causes a
problem for test networks as the same prefixes (`tpub`, `upub`, and
`vpub`) are used for both testnet and regtest. With the previous
solution, we were always resolving a testnet address.

It's not a big deal for `tpub` and `upub` keys as the addresses for
testnet and regtest have the same format. The problematic keys are
`vpub`. For example an extended public key
`vpub5Zx5difzitDBNPjrr9pTno6C44dJFd89naYzhyk9QWHFTpF7pJqnyAnADhbVrFYX7eCK8V2WBBVprxzJrSk15NsYHiB8CvV8h4JnXkU66as`
resolves addresses at `m/44'/0'/0'/0/4` differently:
- for `testnet`: `tb1qjy5r90er70t2cexwpmmkf9hr4glxdx83jhpwfv`
- for `regtest`: `bcrt1qjy5r90er70t2cexwpmmkf9hr4glxdx83s7cr79`

With the change introduced here we will no longer derive the chain
params from the prefix, but expect them to be provided. This should not
be a problem as the chain type is configured by the user in the config TOML
file.

We also introduced a validation for prefixes and relevant networks.

Depends on https://github.com/keep-network/keep-ecdsa/pull/833

### DIFF
--- a/pkg/chain/bitcoin/derive_address.go
+++ b/pkg/chain/bitcoin/derive_address.go
@@ -35,7 +35,11 @@ import (
 // [BIP44]: https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki
 // [BIP49]: https://github.com/bitcoin/bips/blob/master/bip-0049.mediawiki
 // [BIP84]: https://github.com/bitcoin/bips/blob/master/bip-0084.mediawiki
-func DeriveAddress(extendedPublicKey string, addressIndex uint32) (string, error) {
+func DeriveAddress(
+	extendedPublicKey string,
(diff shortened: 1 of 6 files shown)


beaushinkle:
We have some relevant looking test failures, but other than that I'm liking this PR

nkuba:
All green @beaushinkle

PR #838, opened 2021-06-22
Network troubleshooting checklist

lukasz-zimnoch:
Here we introduce a network troubleshooting checklist that should provide guidance for operators dealing with the most common network problems observed during the Keep Network lifetime.


docs/network-troubleshooting.adoc
```diff
@@ -0,0 +1,135 @@
...
+<<./run-keep-ecdsa.adoc#322-parameters, `LibP2P.AnnounceAddresses`>>
+config property. Last but not least, your firewall rules should allow the
+ingress traffic to reach your client.
+
+=== Check your client is able to reach the public internet
+Your client should be able to reach other clients over the public internet.
+Make sure your network routing, NAT and firewall setup allows your client
+to do so. If this requirement is not fulfilled, your client won't be able
+to initiate outgoing connections though it may be able to receive incoming
+ones.
```
Shadowfiend, line 113: For this and public reachability, are these things that would be reflected in the metrics? That is, would it be reflected in the number of connected peers? Or is that only true if the node is both unreachable *and* unable to reach public internet?

docs/network-troubleshooting.adoc
```diff
@@ -0,0 +1,135 @@
...
+
+=== Make sure your client is publicly reachable
+Check if your client uses a publicly available IP or DNS and announces it
+in the network. In case a private address is used, your node won't be able
+to receive incoming connections though it may be able to initiate them on their
+own. Public availability can be checked using the `nmap` tool. Regarding
+announcing, you need to use the
+<<./run-keep-ecdsa.adoc#322-parameters, `LibP2P.AnnounceAddresses`>>
+config property. Last but not least, your firewall rules should allow the
+ingress traffic to reach your client.
```
Shadowfiend, line 106: Keeping in mind our general audience baseline of “reasonably equipped individual operators”, I wonder if we should use “inbound traffic” and “outbound traffic” instead of ingress/egress. Thoughts?

docs/network-troubleshooting.adoc
```diff
@@ -0,0 +1,135 @@
...
+amount drops below the minimum threshold, your client won't be able to
+connect other clients and all existing connections will be dropped.
+
+=== Make sure your client is publicly reachable
+Check if your client uses a publicly available IP or DNS and announces it
+in the network. In case a private address is used, your node won't be able
+to receive incoming connections though it may be able to initiate them on their
+own. Public availability can be checked using the `nmap` tool. Regarding
+announcing, you need to use the
+<<./run-keep-ecdsa.adoc#322-parameters, `LibP2P.AnnounceAddresses`>>
```
Shadowfiend, line 104: The `322` prefix here I believe is unstable in the face of section changes; I think you can use `#Parameters`. If that doesn't work, you can add a `[[parameters]]` section anchor before `=== Parameters` and use `#parameters` to reference it.

docs/network-troubleshooting.adoc
```diff
@@ -0,0 +1,135 @@
...
+The cause of aforementioned problems may lie in either member of the keep.
+To make sure your client performs well, please pass through the following
+checklist.
+
+=== Check your client health
+Check your client is up and doesn't experience frequent restarts. If your clients
+goes down, it breaks the ongoing protocol execution for other keep members.
+This is especially relevant for cloud workloads where things like VMs preemption
+or Kubernetes pod relocation may take place. It's strongly recommended to enable
+<<./run-keep-ecdsa.adoc#9-metrics, metrics>>
```
Shadowfiend, line 73: See my note below on using a more stable reference anchor.

Shadowfiend (commented):
Left a few small remarks; love this doc!

docs/network-troubleshooting.adoc
```diff
@@ -0,0 +1,135 @@
...
+Your client should be able to reach other clients over the public internet.
+Make sure your network routing, NAT and firewall setup allows your client
+to do so. If this requirement is not fulfilled, your client won't be able
+to initiate outgoing connections though it may be able to receive incoming
+ones.
+
+=== Make sure your client address is stable
+Your client should not change it's announced address to often. All changes
+need time to be propagated over the network so other clients may not be able
+to reach your client immediately. It is also important that your client
```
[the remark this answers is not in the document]
lukasz-zimnoch: I'm afraid we can't provide even a ballpark here. The propagation time isn't linear and strongly depends on the current network topology, the number of mutual connections between clients, and how the peer discovery rounds of an individual client overlap with rounds of their peers.

docs/network-troubleshooting.adoc
lukasz-zimnoch: As you said, this is true only if the node is both unreachable and unable to reach the public internet. Both problems occurring separately may not be observable in metrics. If it's unreachable but can reach the public internet, for example, through a NAT gateway, it can still initiate connections on its own so connected peers will be non-zero. Conversely, when it is reachable but can't reach the public internet, it can still receive inbound connections.

docs/network-troubleshooting.adoc
```diff
@@ -0,0 +1,143 @@
...
+temporarily due to restart or network breaks. In that case, clients should
+recover automatically as `keep-ecdsa` clients are resilient and can
+continue their work after reconnecting to the network or coming up after the
+restart. However, sometimes the problem persists and troubleshooting process
+should be run on all clients taking part in the unlucky protocol execution.
+
+== Common problems
+
+Typically, there are several types of network-related problems reported by
+operators. Here is a brief description of them.
```
pdyraga, line 24: It reads to me that usually there is more than one problem behind a particular key generation/signing failure. Maybe we can rephrase it into something like:

> Here are the most common problems reported by operators with a brief description for each one of them.

?

docs/network-troubleshooting.adoc
```diff
@@ -0,0 +1,143 @@
...
+keep members, the following error will be printed in logs:
+
+```
+failed to announce signer presence: [waiting for announcements timed out after: [2m0s]]"
+```
+
+=== Readiness signaling protocol failed
+
+Readiness protocol is performed before both key generation and signing
+processes. It's purpose is to synchronize keep members and allows each member
```
pdyraga, line 52: `s/It's/Its`

docs/network-troubleshooting.adoc
```diff
@@ -0,0 +1,143 @@
...
+to signal their readiness to perform the TSS protocol. Like the announce
```
pdyraga, line 53: I would replace "TSS" with just "key generation/signing". I do not think anyone but core developers know what TSS stands for.

docs/network-troubleshooting.adoc
```diff
@@ -0,0 +1,143 @@
...
+Check if your node has a working connection with their ETH node. Broken
+ETH connection means that your client becomes blind for all chain events
+taking place for keeps it is a member of. For example, it may not notice a
+starting key generation or signing process. ETH connectivity can be
+watched using metrics and a monitoring system
+
+=== Check your client fulfills the minimum stake rule
+Check your client has the minimum amount of KEEP token staked. If the staked
+amount drops below the minimum threshold, your client won't be able to
+connect other clients and all existing connections will be dropped.
```
pdyraga, line 97: I do not think this is true. If the operator has at least one active keep it will be able to connect the rest of the network, even if the operator does not have a minimum stake. I would consider dropping this section.

docs/network-troubleshooting.adoc
```diff
@@ -0,0 +1,143 @@
...
+
+=== Make sure your client is publicly reachable
+Check if your client uses a publicly available IP or DNS and announces it
+in the network. In case a private address is used, your node won't be able
+to receive incoming connections though it may be able to initiate them on their
+own. Public availability can be checked using the `nmap` tool. Regarding
+announcing, you need to use the
+<<./run-keep-ecdsa.adoc#Parameters, `LibP2P.AnnounceAddresses`>>
+config property. Last but not least, your firewall rules should allow the
```
pdyraga, line 106: I do not have to use this property if the server I am running the client on has a public IP and I am not using any port forwarding/NAT magic, correct? I think we need to be more specific here.

docs/network-troubleshooting.adoc
```diff
@@ -0,0 +1,143 @@
...
+
+=== Check your client is able to reach the public internet
+Your client should be able to reach other clients over the public internet.
+Make sure your network routing, NAT and firewall setup allows your client
+to do so. If this requirement is not fulfilled, your client won't be able
+to initiate outgoing connections though it may be able to receive incoming
+ones.
+
+=== Make sure your client address is stable
+Your client should not change it's announced address to often. All changes
```
pdyraga, line 117: s/it's/its sir

docs/network-troubleshooting.adoc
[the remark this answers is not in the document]
pdyraga: I think the ballpark Antonio proposed makes sense. We had cases when operators expected to have the full mesh in 3 minutes after the restart and they kept restarting not seeing it to happen which was causing only bigger issues. We should not be worried giving a number here, it is important and it's just a ballpark after all 🤷

docs/network-troubleshooting.adoc
```diff
@@ -0,0 +1,143 @@
...
+can help monitor network connectivity as they return the number of connected
+peers. A significant fall of connected peers number may indicate network
+issues.
+
+=== Check Ethereum connectivity
+Check if your node has a working connection with their ETH node. Broken
+ETH connection means that your client becomes blind for all chain events
+taking place for keeps it is a member of. For example, it may not notice a
+starting key generation or signing process. ETH connectivity can be
+watched using metrics and a monitoring system
```
pdyraga, line 92: I would also mention  there should be no errors in the logs about the state of websocket connection.

docs/network-troubleshooting.adoc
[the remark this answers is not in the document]
Shadowfiend: I think if we can talk about what we see in the current network topology, it's a good start. We can also leave a note that larger peer numbers and unusual connectivity can lead to longer propagation time.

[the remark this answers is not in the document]
lukasz-zimnoch: Ok, makes sense. Please see: https://github.com/keep-network/keep-ecdsa/pull/838/commits/6fde9621e265a2d849d0812e13fc5aa3139d83a3

docs/network-troubleshooting.adoc
lukasz-zimnoch: Sure. See https://github.com/keep-network/keep-ecdsa/pull/838/commits/8437b4e2934bcc70f8a88d7b09bd1677234535ab

docs/network-troubleshooting.adoc
lukasz-zimnoch: Good catch, see https://github.com/keep-network/keep-ecdsa/pull/838/commits/c37ea1b3298859422a185d36b928ede271cbd4fe

docs/network-troubleshooting.adoc
lukasz-zimnoch: Yeah, you're right. I missed the part with active keeps. Dropped this section: https://github.com/keep-network/keep-ecdsa/pull/838/commits/90ea18588315678aef928286ff40a6b45fad156c

docs/network-troubleshooting.adoc
lukasz-zimnoch: You're right in general though even then there may be cases when this property is needed. I think we should recommend setting it anyway. It doesn't make anything bad if set redundantly. Also, operators don't need to wonder whether they have to set this property. Instead, they should worry only about the public availability of the address passed here.

docs/network-troubleshooting.adoc
lukasz-zimnoch: https://github.com/keep-network/keep-ecdsa/pull/838/commits/af2019698db95bfe16d02286e70c129ce96715fe

docs/network-troubleshooting.adoc
lukasz-zimnoch: Done https://github.com/keep-network/keep-ecdsa/pull/838/commits/248102e9d48a0600e1d80d5cf1c769832bfce224

lukasz-zimnoch:
Everything addressed! Ready for another shot.

docs/network-troubleshooting.adoc
pdyraga: For professional operators this is true but in general, I see two problems. First, for operators not being technically savvy enough, not setting this value is better than setting this value incorrectly. An example could be a public IP that is changing on VM relocation. Secondly, we are not requiring this value to be set in the config:

https://github.com/keep-network/keep-ecdsa/blob/a82224481891b27e26235012e465493c2e2044a7/configs/config.toml.SAMPLE#L69-L70

I would vote for leaving the decision to the operator but recommending setting this value in case of a more complex setup. Something like:
> For more complex deployment setups when the client is not publicly available under the default address resolved by the client, we recommend setting `LibP2P.AnnouneAddresses` property in the config file.

docs/network-troubleshooting.adoc
pdyraga: Wen BGP ingress/egress filtering recommendations for infra? 😆

docs/network-troubleshooting.adoc
```diff
@@ -0,0 +1,143 @@
...
+
+=== Check your client is able to reach the public internet
+Your client should be able to reach other clients over the public internet.
+Make sure your network routing, NAT and firewall setup allows your client
+to do so. If this requirement is not fulfilled, your client won't be able
+to initiate outgoing connections though it may be able to receive incoming
+ones.
+
+=== Make sure your client address is stable
+Your client should not change its announced address to often. All changes
```
pdyraga, line 113: s/to often/too often

pdyraga:
Two comments on my side. Other than that, I am good to :shipit:.

pdyraga (approved):

PR #841, opened 2021-06-24
Don't Give Up: continue with liquidation recovery when the electrs connection fails

### LINKED ISSUE #840: Liquidation Recovery: Protocol aborts for *pub and missing bitcoin connection
When *pub key is configured and the connection to bitcoin is missing, the liquidation recovery protocol aborts with an error instead of continuing the execution.

**Expected behavior**
> If the GET request fails, the index should be assumed to be valid, as if the Bitcoin node connection were not available.

**Current behavior**
The client logs an error and aborts the execution:

> 2021-06-23T17:02:18.687+0200	ERROR	keep-ecdsa	failed to resolve a btc address for keep: [0x905205c1157aBE87a29C31D1E9168D1429c2Ef77] address: [vpub5Zx5difzitDBNPjrr9pTno6C44dJFd89naYzhyk9QWHFTpF7pJqnyAnADhbVrFYX7eCK8V2WBBVprxzJrSk15NsYHiB8CvV8h4JnXkU66as] err: [attempted to call IsAddressUnused with no apiURL]

beaushinkle:
Currently, when a user has a non-working electrs connection configured and an extended public key as their beneficiary address, we are unable to complete liquidation recovery because we error out trying to verify that a derived address is unused.

This PR changes that behavior to log the failure as an error, and then continue on with processing using the index supplied by local persistence.

Tagging @nkuba for review

closes #840

### DIFF
--- a/pkg/chain/bitcoin/electrs.go
+++ b/pkg/chain/bitcoin/electrs.go
@@ -130,10 +130,12 @@ func (e electrsConnection) VbyteFeeFor25Blocks() (int32, error) {
 }
 
 // IsAddressUnused returns true if and only if the supplied bitcoin address has
-// no recorded transactions.
+// no recorded transactions. NOTE: IsAddressUnused will return true rather than
+// false in the case that it encounters an error. This lets processing continue
+// in the case where there is not a working electrs connection.
 func (e electrsConnection) IsAddressUnused(btcAddress string) (bool, error) {
 	if e.apiURL == "" {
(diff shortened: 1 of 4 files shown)


beaushinkle:
@nkuba anything else you want to see here?

nkuba (approved):
I like simplicity of this fix 😄  The initial problem was fixed 🎉  Tests discovered another issue which is solved in https://github.com/keep-network/keep-ecdsa/pull/850

PR #843, opened 2021-06-24
Run npm install at Package Root

beaushinkle:
Without this, we don't install or pull down the prettier dependencies, and any attempt to commit code won't pass pre-commit hooks.

Tagging @nkuba for review


scripts/install.sh
```diff
@@ -63,6 +63,11 @@ printf "${LOG_START}Starting installation...${LOG_END}"
 cd $KEEP_ECDSA_SOL_PATH
```
nkuba, line 64:
```suggestion
```

nkuba (changes requested):

nkuba (approved):

PR #847, opened 2021-06-28
Extract liquidation recovery handling to a separate function and unit test it

nkuba:
In this PR we extract logic for handling liquidation recovery to a separate function so we are able to cover it with unit tests.

The tests are focused on testing the logic with different states of bitcoin connection. It's expected that if bitcoin connection fails the process moves on and completes based on the local data.

For the time being the signers are generated with TSS protocol but it would be awesome to load them from test data. Unfortunatelly I was not able to get it working in the limited time frame so leaving it as is.

The tests are failing due to the https://github.com/keep-network/keep-ecdsa/issues/840.

Depends on: https://github.com/keep-network/keep-ecdsa/issues/840

### DIFF
--- a/internal/testhelper/registry.go
+++ b/internal/testhelper/registry.go
@@ -7,6 +7,7 @@ import (
 	"github.com/keep-network/keep-ecdsa/pkg/ecdsa/tss"
 )
 
+// PersistenceHandleMock is a mock of persistence handle used in tests.
 type PersistenceHandleMock struct {
 	PersistedGroups  []*TestFileInfo
 	Snapshots        []*TestFileInfo
@@ -15,6 +16,7 @@ type PersistenceHandleMock struct {
 	outputErrorsChan chan error
 }
 
+// NewPersistenceHandleMock creates a mocked persistence handle.
 func NewPersistenceHandleMock(outputDataChanSize int) *PersistenceHandleMock {
 	return &PersistenceHandleMock{
 		outputDataChan:   make(chan persistence.DataDescriptor, outputDataChanSize),
@@ -22,12 +24,14 @@ func NewPersistenceHandleMock(outputDataChanSize int) *PersistenceHandleMock {
 	}
 }
 
+// TestFileInfo holds test data stored in persistence handle.
 type TestFileInfo struct {
 	Data      []byte
 	Directory string
 	Name      string
 }
 
+// Save stores data in persistence handle.
 func (phm *PersistenceHandleMock) Save(data []byte, directory string, name string) error {
 	phm.PersistedGroups = append(
 		phm.PersistedGroups,
@@ -37,6 +41,7 @@ func (phm *PersistenceHandleMock) Save(data []byte, directory string, name strin
 	return nil
 }
 
+// Snapshot creates a snapshot of data in persistence handle.
 func (phm *PersistenceHandleMock) Snapshot(data []byte, directory string, name string) error {
 	phm.Snapshots = append(
 		phm.Snapshots,
@@ -46,6 +51,7 @@ func (phm *PersistenceHandleMock) Snapshot(data []byte, directory string, name s
 	return nil
 }
 
+// MockSigner registers a mock of a signer for membership and keep.
 func (phm *PersistenceHandleMock) MockSigner(membershipIndex int, keepID string, signer *tss.ThresholdSigner) error {
 	signerBytes, err := signer.Marshal()
 	if err != nil {
@@ -61,6 +67,7 @@ func (phm *PersistenceHandleMock) MockSigner(membershipIndex int, keepID string,
 	return nil
 }
 
+// ReadAll reads all data stored in persistence handle.
 func (phm *PersistenceHandleMock) ReadAll() (<-chan persistence.DataDescriptor, <-chan error) {
 	close(phm.outputDataChan)
 	close(phm.outputErrorsChan)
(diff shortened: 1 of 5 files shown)


internal/testhelper/registry.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+}
+
+type TestFileInfo struct {
+	Data      []byte
+	Directory string
+	Name      string
+}
+
+func (phm *PersistenceHandleMock) Save(data []byte, directory string, name string) error {
```
beaushinkle, line 31: I think the linter will yell at you unless you give all of these public structs / functions comments

pkg/client/liquidation_recovery.go
```diff
@@ -0,0 +1,184 @@
...
+			"failed to retrieve a vbyte fee estimate, [%v]",
+			vbyteFeeError,
+		)
+		// Since the electrs connection is optional, we don't return the error
+	}
+	if vbyteFee == 0 {
+		vbyteFee = tbtcConfig.Bitcoin.MaxFeePerVByte
+	}
+	if vbyteFee == 0 {
+		vbyteFee = 75
```
beaushinkle, line 100: What do you think about pulling this 75 out into a constant and then writing a comment to explain the double `if vbyteFee == 0` check?

pkg/client/liquidation_recovery_test.go
```diff
@@ -0,0 +1,506 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	tbtcHandle, err = localChain.TBTCApplicationHandle()
+	if err != nil {
+		panic(err)
+	}
+	tbtcHandle.(*chainLocal.TBTCLocalChain).CreateDeposit(depositAddress.String(), keepMembersAddresses)
+
```
beaushinkle, line 79: extra whitespace

pkg/client/liquidation_recovery_test.go
```diff
@@ -0,0 +1,506 @@
...
+	if err := log.SetLogLevel("*", "DEBUG"); err != nil {
+		t.Fatal(err)
+	}
+
+	bitcoinAddresses := []string{
+		"1MjCqoLqMZ6Ru64TTtP16XnpSdiE8Kpgcx",
+		"bc1q46uejlhm9vkswfcqs9plvujzzmqjvtfda3mra6",
+		"398r9poPaoKJ7vHkaVzNVsXBGRB3mFMXEK",
+	}
+	bitcoinExtenderPublicKeys := []string{
```
beaushinkle, line 92: s/Extender/Extended

pkg/client/liquidation_recovery_test.go
```diff
@@ -0,0 +1,506 @@
...
+		// - resolved maxFeePerVByte
+		// - logged 5x warn on broadcast failure
+		// - cover more failures
+	}
+
+	for testName, testData := range testCases {
+		t.Run(testName, func(t *testing.T) {
+			bitcoinHandle := testData.configureBitcoinHandle()
+
+			ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
beaushinkle, line 183: I think 5 secs might be pushing it for slower machines or when the tests are running under load

pkg/client/liquidation_recovery_test.go
```diff
@@ -0,0 +1,506 @@
...
+
+				networkProviders[memberID.String()] = networkProvider
+
+				providersInitializedWg.Done()
+				providersInitializedWg.Wait()
+
+				// FIXME: Load signers from local test data storage instead of
+				// generating them with tss.GenerateThresholdSigner.
+				//
+				// signer, err := newTestSigner(groupMemberIDs, index)
```
beaushinkle, line 364: I think that if we don't have a working implementation at the moment, I'd rather not commit an unused function or commented out function call. I think the FIXME is good though

pkg/client/liquidation_recovery.go
nkuba: Yep, makes perfect sense! The two ifs confused me when I got back to this code after a while.

pkg/client/liquidation_recovery_test.go
nkuba: It took 2 secs on my machine, but yeah 5 seconds may be risky. I'll bump it up to 10 secs maybe.

internal/testhelper/registry.go
nkuba: 😱 great point!

pkg/client/liquidation_recovery.go
```diff
@@ -0,0 +1,199 @@
...
+			"failed to communicate recovery details for keep [%s]: [%v]",
+			keep.ID(),
+			err,
+		)
+		return err
+	}
+
+	signer, err := keepsRegistry.GetSigner(keep.ID())
+	if err != nil {
+		// If there are no signer for loaded keep that something is clearly
```
beaushinkle, line 117:
```suggestion
		// If there are no signer for loaded keep then something is clearly
```

beaushinkle:
Okay - I'm good here. I highlighted a spelling mistake, but we can definitely fix that in a follow-up PR

pkg/client/liquidation_recovery.go
nkuba: https://github.com/keep-network/keep-ecdsa/pull/854

Issue #857, opened 2021-07-05
Liquidation Recovery: Compare suggested fee vs max configured fee


nkuba:
According to [RFC-2](https://github.com/keep-network/keep-ecdsa/blob/master/docs/rfc/rfc-2.adoc#dealing-with-the-bitcoin-fee) fee suggested by Bitcoin handle should be compared with max fee configured by the operator:

> There is one exception to the rule that the 25-block suggested fee should be used in the presence of a Bitcoin connection. If this suggested fee would result in a fee consuming more than 5% of the UTXO value that is being split, then the lesser of the 25-block suggested fee and MaxFeePerVByte configured fee should be used. Because the true fee computation requires that the full transaction be assembled, the transaction fee should be estimated by assuming the final transaction will be 175 vBytes, which should be very close to accurate given the regular structure of liquidation refund transactions.
If the final fee is computed as being >5% of UTXO value, the client should log a WARN message so that the operator can be aware of this and revise their max fee per vByte if desired.

PR #859, opened 2021-07-07
Liquidation recovery:  Compare estimated transaction fee with 5% utxo value 

nkuba:
According to [RFC-2](https://github.com/keep-network/keep-ecdsa/blob/master/docs/rfc/rfc-2.adoc#dealing-with-the-bitcoin-fee) fee suggested by Bitcoin handle should be compared with max fee configured by the operator:

> There is one exception to the rule that the 25-block suggested fee should be used in the presence of a Bitcoin connection. If this suggested fee would result in a fee consuming more than 5% of the UTXO value that is being split, then the lesser of the 25-block suggested fee and MaxFeePerVByte configured fee should be used. Because the true fee computation requires that the full transaction be assembled, the transaction fee should be estimated by assuming the final transaction will be 175 vBytes, which should be very close to accurate given the regular structure of liquidation refund transactions.
If the final fee is computed as being >5% of UTXO value, the client should log a WARN message so that the operator can be aware of this and revise their max fee per vByte if desired.

This PR has couple more improvements that are described in the commits.

Closes: https://github.com/keep-network/keep-ecdsa/issues/857

### DIFF
--- a/pkg/chain/local/tbtc.go
+++ b/pkg/chain/local/tbtc.go
@@ -3,7 +3,6 @@ package local
 import (
 	"bytes"
 	"context"
-	"encoding/binary"
 	"encoding/hex"
 	"fmt"
 	"math/big"
@@ -17,9 +16,8 @@ import (
 )
 
 const (
-	defaultUTXOValue            = 10000000
 	defaultInitialRedemptionFee = 10
-	defaultUtxoValueHex         = "8096980000000000"
+	defaultUtxoValueHex         = "8096980000000000" // 10000000
 	defaultFundedAt             = "1615172517"
 	previousTransactionHashHex  = "c27c3bfa8293ac6b303b9f7455ae23b7c24b8814915a6511976027064efc4d51"
 	previousTransactionIndex    = 1
@@ -190,9 +188,9 @@ func (tlc *TBTCLocalChain) CreateDeposit(
 	tlc.deposits[depositAddress] = &localDeposit{
 		keepAddress: keepAddress.Hex(),
 		state:       chain.AwaitingSignerSetup,
-		utxoValue:   big.NewInt(defaultUTXOValue),
+		utxoValue:   big.NewInt(int64(chain.UtxoValueBytesToUint32(utxoValueBytes))),
 		fundingInfo: &chain.FundingInfo{
-			UtxoValueBytes:  utxoValueBytes, // 0x0065cd1d00000000
+			UtxoValueBytes:  utxoValueBytes,
 			FundedAt:        fundedAt,
 			TransactionHash: previousTransactionHashHex,
 			OutputIndex:     previousTransactionIndex,
@@ -773,5 +771,5 @@ func (tlc *TBTCLocalChain) Logger() *ChainLogger {
 }
 
 func fromLittleEndianBytes(bytes [8]byte) *big.Int {
-	return new(big.Int).SetUint64(binary.LittleEndian.Uint64(bytes[:]))
+	return new(big.Int).SetUint64(uint64(chain.UtxoValueBytesToUint32(bytes)))
 }
--- a/pkg/chain/tbtc_test.go
+++ b/pkg/chain/tbtc_test.go
@@ -73,3 +73,47 @@ func TestParseUtxoOutpoint(t *testing.T) {
 		})
 	}
 }
+
+func TestUtxoValueBytesToUint32(t *testing.T) {
+	testData := map[string]struct {
+		utxoValueBytes [8]uint8
+		expectedValue  uint32
+	}{
+		"0": {
+			[8]uint8{0, 0, 0, 0}, // 0x00000000
+			uint32(0),
+		},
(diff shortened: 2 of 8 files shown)


pkg/chain/local/tbtc.go
```diff
@@ -17,9 +16,8 @@ import (
 )
 
 const (
-	defaultUTXOValue            = 10000000
 	defaultInitialRedemptionFee = 10
-	defaultUtxoValueHex         = "8096980000000000"
+	defaultUtxoValueHex         = "8096980000000000" // 10000000
```
beaushinkle, line 20: :100:

pkg/chain/tbtc_test.go
```diff
@@ -73,3 +73,47 @@ func TestParseUtxoOutpoint(t *testing.T) {
...
+	}
+	for testName, testData := range testData {
+		t.Run(testName, func(t *testing.T) {
+			actualValue := UtxoValueBytesToUint32(testData.utxoValueBytes)
+			if actualValue != testData.expectedValue {
+				t.Errorf("unexpected result\nexpected: %d\nactual:   %d", testData.expectedValue, actualValue)
+			}
+		})
+	}
+}
```
beaushinkle, line 119: love it

pkg/client/liquidation_recovery.go
```diff
@@ -176,17 +186,59 @@ func handleLiquidationRecovery(
...
+		if estimatedTransactionFee > fivePercentPreviousOutput {
+			if tbtcConfig.Bitcoin.MaxFeePerVByte > 0 {
+				vbyteFee = min(vbyteFee, tbtcConfig.Bitcoin.MaxFeePerVByte)
+			}
+
+			// If the final fee is computed as being >5% of UTXO value, the client
+			// should log a WARN message so that the operator can be aware of this
+			// and revise their max fee per vByte if desired.
+			logger.Warnf(
+				"estimated transaction fee [%d] is greater than 5%% of the UTXO value [%d]; "+
```
beaushinkle, line 235:
```suggestion
				"estimated transaction fee [%d] is greater than 5% of the UTXO value [%d]; "+
```

pkg/client/liquidation_recovery.go
```diff
@@ -176,17 +186,59 @@ func handleLiquidationRecovery(
...
+		// Since the electrs connection is optional, we don't return the error.
+	}
+	if vbyteFee > 0 {
+		// Fee computation requires that the full transaction be assembled. The
+		// transaction fee should be estimated by assuming the final transaction
+		// will be 175 vBytes, which should be very close to accurate given the
+		// regular structure of liquidation refund transactions.
+		estimatedTransactionFee := vbyteFee * estimatedTransactionSizeVByte
+
+		fivePercentPreviousOutput := previousOutputValue / 20 // 5% of UTXO
```
beaushinkle, line 219: Do we lose precision here because everything is integers? Does that actually matter? 🤔

pkg/client/liquidation_recovery.go
```diff
@@ -197,3 +249,10 @@ func resolveVbyteFee(bitcoinHandle bitcoin.Handle, tbtcConfig *tbtc.Config) int3
 
 	return vbyteFee
 }
+
+func min(a, b int32) int32 {
```
beaushinkle, line 253: it's pretty wild to me that this isn't in the standard lib

pkg/client/liquidation_recovery.go
nkuba: We need to use a double percent sign (`%%`) for literal value in the format function. 

See https://golang.org/pkg/fmt/#hdr-Printing:
> %%	a literal percent sign; consumes no value

beaushinkle: !

pkg/client/liquidation_recovery.go
nkuba: We do lose precision, but I believe it doesn't matter to us that much here.

We expect `previousOutputValue` to be relatively big values. The lowest TBTC deposit lot size is 0.01 BTC, so `previousOutputValue` will be `1000000` (0.01 BTC -> satoshis).

We will use the calculated value in `estimatedTransactionFee > fivePercentPreviousOutput` comparison.
`estimatedTransactionFee` will be min. `175` (`estimatedTransactionFee = vbyteFee * estimatedTransactionSizeVByte`)

The lowest possible scenario where precision would matter is `estimatedTransactionFee` to be `175` or `176`. Which corresponds to `previousOutputValue` of `175*20 = 3500` and `176*20=3520`.

Also, the calculation result is rounded down so it works in the operator's favor in `estimatedTransactionFee > fivePercentPreviousOutput` check.

Please let me know if I made some mistakes in my assumptions and we should be more precision caucious.

beaushinkle: Okay! Double-checked that math there and it all looks good. Thanks for proving that out for me!

PR #877, opened 2021-08-02
Remove contract artifact folder before copying

tomaszslabon:
This PR adds a small change - remove the folder with contract artifact before copying.
Also the `artifacts/` directory is added to `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,7 @@ node_modules/
 # Generated files
 *.abi
 abi/
+artifacts/
 
 # Build output
 build/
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -22,6 +22,7 @@ node_modules/
 # Generated files
 *.abi
 abi/
+artifacts/
```
nkuba, line 25: I'm not sure if `Generated files` is a good section for this directory, it feels more like `Build output`, or even `Deployment output`.

nkuba (approved):
Left a non-blocking comment.

PR #878, opened 2021-08-03
Update locked sha3 version to 1.2.6

nkuba:
In this PR we updated version of `sha3` in `package-lock.json` file to `1.2.6`, as the previous version was erroring node-gyp in npm install on Node 14.

### DIFF
--- a/solidity/package-lock.json
+++ b/solidity/package-lock.json
@@ -34096,9 +34096,9 @@
       }
     },
     "sha3": {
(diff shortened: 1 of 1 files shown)


michalinacienciala (approved):
I've tested it locally, looking good! I created a PR with accompanying changes in the `local-setup`: https://github.com/keep-network/local-setup/pull/105.

PR #888, opened 2021-08-13
Deposit is unfunded, no need to panic

nkuba:
Liquidation recovery is invoked when a keep is terminated. It gets Bitcoin funding details from a tBTC Deposit associated with the keep and assembles a Bitcoin recovery transaction. 

If keep termination happens before the deposit got funded (signer setup failure notification) there are no funds to recover.
In such a case, the client panicked and died on the utxo outpoint details parsing.

In this PR we fix the problem described above by checking if the deposit is funded. If not we abort the liquidation recovery process so it doesn't end up in a retry loop (dde85f63ce1d92fe8851d1a87bb93c7eef41154e).

We also introduced a check to utxo outpoint length just in case, so the client don't panic but returns an error (6a97d7627a0e70a81cd02c2adca3adb17224ae51).

There are also some fixes and improvements to the code and unit tests.

Depends on https://github.com/keep-network/keep-ecdsa/pull/887


pkg/chain/celo/tbtc.go
```diff
@@ -452,7 +452,24 @@ func (ta *tbtcApplication) FundingInfo(
...
 	}
 
-	transactionHash, outputIndex := chain.ParseUtxoOutpoint(fundingInfo.UtxoOutpoint)
+	logger.Debugf(
+		"deposit [%s] funding info: %+v",
+		depositAddress,
+		fundingInfo,
+	)
+
+	if fundingInfo.FundedAt == nil || fundingInfo.FundedAt.Cmp(big.NewInt(0)) <= 0 {
```
beaushinkle, line 461: when would `FundedAt` be negative?

pkg/chain/local/tbtc.go
```diff
@@ -180,20 +180,11 @@ func (tlc *TBTCLocalChain) CreateDeposit(
...
 	tlc.deposits[depositAddress] = &localDeposit{
 		keepAddress: keepAddress.Hex(),
 		state:       chain.AwaitingSignerSetup,
-		utxoValue:   big.NewInt(int64(chain.UtxoValueBytesToUint32(utxoValueBytes))),
 		fundingInfo: &chain.FundingInfo{
-			UtxoValueBytes:  utxoValueBytes,
-			FundedAt:        fundedAt,
-			TransactionHash: previousTransactionHashHex,
-			OutputIndex:     previousTransactionIndex,
+			FundedAt: big.NewInt(0),
```
beaushinkle, line 187: ahh, i see - `FundedAt` is probably not going to be <= 0 in normal cases, but it can definitely be set like that in tests/mocks

beaushinkle:
Did a line-by-line. Lookin' good to me!

pkg/chain/celo/tbtc.go
nkuba: > when would FundedAt be <= 0?

We assume a deposit is funded when `FundedAt > 0`. Here we need to check the opposite, so I used `<= 0` just in case we get a negative value from the chain.

> edit: also, does fundingInfo.FundedAt <= big.NewInt(0) work here?

Nope, we need to use `Cmp` function to compare `big.Int`s.

pkg/chain/local/tbtc.go
nkuba: Yes, we expect the lowest to be `0` but you never know...

nkuba:
@beaushinkle #887 was merged already so this PR is ready for review/merge.

beaushinkle (approved):

PR #892, opened 2021-08-24
Bitcoin config cleanup

nkuba:
In b0e7a25 we decided to move the bitcoin specific config to `bitcoin.Config, but looks like we didn't remove the old code. Here we clean it up.

We also resolve the default Bitcoin endpoint to use based on the `BitcoinChainName` property. For `testnet3` we use Blockstream's  testnet endpoint, and in other cases we use Blockstream's mainnet endpoint.


beaushinkle:
> but looks like we didn't remove the old code

whoops!

beaushinkle (approved):

PR #893, opened 2021-08-24
Configure workflow for docs publication

nkuba:
We want to publish documentation in the similiar manner we do for
keep-core and tbtc projects.


.github/workflows/docs.yml
```diff
@@ -0,0 +1,107 @@
...
+  push:
+    branches:
+      - main
+    paths:
+      - "docs/**"
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  docs-detect-changes:
```
lukasz-zimnoch, line 13: Maybe just `detect-changes`? We are already in the `Docs` workflow. This is also consistent with names we have in other repos.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,107 @@
...
+        if: github.event_name == 'pull_request'
+        id: filter
+        with:
+          filters: |
+            path-filter:
+              - './docs/**'
+
+  # TODO: Add generation of Solidity documentation
+
+  docs-html:
```
lukasz-zimnoch, line 30: Maybe `generate-and-publish-html`?

.github/workflows/docs.yml
```diff
@@ -0,0 +1,107 @@
...
+        name: Upload asciidocs preview
+        uses: thesis/gcp-storage-bucket-action@v3.1.0
+        with:
+          service-key: ${{ secrets.DOCS_KEEP_NETWORK_UPLOADER_SERVICE_KEY_JSON }}
+          project: cfc-production
+          bucket-name: docs.keep.network
+          bucket-path: ecdsa/${{ github.head_ref }}
+          build-folder: ${{ steps.html.outputs.asciidoctor-artifacts }}/docs
+
+  docs-pdf:
```
lukasz-zimnoch, line 69: Maybe `generate-and-publish-pdf`?

.github/workflows/docs.yml
```diff
@@ -0,0 +1,107 @@
...
+            path-filter:
+              - './docs/**'
+
+  # TODO: Add generation of Solidity documentation
+
+  docs-html:
+    runs-on: ubuntu-latest
+    needs: [docs-detect-changes]
+    if: | # `always()` returns true, even if job should be cancelled
+      always()
```
lukasz-zimnoch, line 34: What's the exact purpose of `always()` here?

.github/workflows/docs.yml
nkuba: I agree it seems redundant to prefix the job name, but we do it to be able to distinguish them in GitHub. There are places where only the job name is used without the workflow context which may cause confusion on duplicates (e.g. in branch protection rules).
We do the same for `client-*` and `contracts-*` jobs in other workflows.

.github/workflows/docs.yml
nkuba: It was copied from `keep-core`. We use it in `keep-core` to continue docs publication no matter of the result of previous step `docs-solidity`. We don't do solidity docs generation here yet so I removed this `always()`. 92f224b

.github/workflows/docs.yml
nkuba: 64bde9a

.github/workflows/docs.yml
nkuba: 64bde9a

PR #905, opened 2021-09-01
Balance monitoring retries

nkuba:
We want to use a retry mechanism for single executions of the balance check.
The retries will be performed during the period of retry timeout, logging a warning on each error. When the timeout is hit an error will be logged and retries stopped. The next balance check will be triggered at the next tick.

Depends on https://github.com/keep-network/keep-common/pull/82


pdyraga (commented):
I am good to go but please update `keep-common` dependency first to point to `main` and not a feature branch given that https://github.com/keep-network/keep-common/pull/82 has been merged.

Are there plans to port this behavior to keep-core as well?

nkuba:
> I am good to go but please update `keep-common` dependency first to point to `main` and not a feature branch given that [keep-network/keep-common#82](https://github.com/keep-network/keep-common/pull/82) has been merged.

✅ 4139244

> Are there plans to port this behavior to keep-core as well?

✅ https://github.com/keep-network/keep-core/pull/2592

pdyraga (approved):

PR #906, opened 2021-09-02
Adding new LP Rewards contract for Saddle TBTCv2 pool

dimpar:
New Saddle LP Token along with a new Saddle TBTCv2 pool is already on mainnet. We want to support rewards for the new Saddle pool and this contract inherits from the previous LP Saddle Rewards and from LPRewards contract.
Saddle LP token: https://etherscan.io/address/0x122Eca07139EB368245A29FB702c9ff11E9693B7
Saddle TBTCv2 Meta Pool: https://saddle.exchange/#/pools/tbtc/deposit


pdyraga:
@dimpar Could you please also update deployment scripts? Can be in this PR or in a separate one.

dimpar:
> @dimpar Could you please also update deployment scripts? Can be in this PR or in a separate one.

Do you mean this guy https://github.com/keep-network/keep-migrations/pull/20 or something else?

solidity/contracts/LPRewards.sol
```diff
@@ -260,3 +260,13 @@ contract LPRewardsTBTCSaddle is LPRewards {
         gated = _gated;
     }
 }
+
+/// @title KEEP rewards for the tBTC V2 Saddle liquidity pool.
+/// @dev LP token: https://etherscan.io/address/0x122Eca07139EB368245A29FB702c9ff11E9693B7
+///      Pool: https://saddle.exchange/#/pools/tbtc/deposit
+contract LPRewardsTBTCv2Saddle is LPRewards, LPRewardsTBTCSaddle {
```
pdyraga, line 267: Does it have to extend `LPRewards` given that `LPRewardsTBTCSaddle` already extends it?

dimpar: Right, we don't need to extend it twice c8d520d

pdyraga (approved):

PR #911, opened 2021-09-06
Adding cron for daily tests and builds on Mercodee

dimpar:
The purpose is to set a cron job that would run tests and builds on daily basis so it can be displayed on Meercode dashboard.

### DIFF
--- a/.github/workflows/client.yml
+++ b/.github/workflows/client.yml
@@ -46,7 +46,8 @@ jobs:
   client-build-test-publish-ethereum:
     needs: client-detect-changes
     if: |
-      github.event_name != 'pull_request'
+      github.event_name == 'push'
+        || github.event_name == 'schedule'
         || needs.client-detect-changes.outputs.path-filter == 'true'
         || (github.event_name == 'workflow_dispatch' 
         && github.event.inputs.environment != 'alfajores')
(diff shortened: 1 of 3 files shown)


.github/workflows/client.yml
```diff
@@ -44,7 +46,7 @@ jobs:
   client-build-test-publish-ethereum:
     needs: client-detect-changes
     if: |
-      github.event_name == 'push'
+      github.event_name != 'pull_request'
```
michalinacienciala, line 49: We don't want to execute the job for example for `workflow_dispatch` with `alfajores` input. With the proposed config, the job would execute for such workflow.

My suggestion:
```suggestion
      github.event_name == 'push'
        || github.event_name == 'schedule'
```

.github/workflows/client.yml
```diff
@@ -147,7 +149,7 @@ jobs:
   client-build-test-publish-celo:
     needs: client-detect-changes
     if: |
-      github.event_name == 'push'
+      github.event_name != 'pull_request'
```
michalinacienciala, line 152: My suggestion:
```suggestion
      github.event_name == 'push'
        || github.event_name == 'schedule'
```

michalinacienciala (changes requested):

.github/workflows/client.yml
dimpar: agree, good catch c534da4a963b8b428d8936a9ff7e9ca0c54f1983

.github/workflows/client.yml
dimpar: c534da4a963b8b428d8936a9ff7e9ca0c54f1983

michalinacienciala (approved):

PR #943, opened 2022-01-04
Using fixed version of sortition-pools package

michalinacienciala:
The `keep-ecdsa` contracts depend on some `sortition-pools` contracts
which are no longer present in the latest npm packages (this is due to
transision to v2). Until we stop supporting `keep-ecdsa` repo
& contracts, we need to configure there the dependncy to the latest
`sortition-pools` package which contains the required contracts. This
is package `@keep-network/sortition-pools@1.2.0-dev.1`.

### DIFF
--- a/solidity/package.json
+++ b/solidity/package.json
@@ -29,7 +29,7 @@
   "homepage": "https://github.com/keep-network/keep-ecdsa",
   "dependencies": {
     "@keep-network/keep-core": ">1.8.0-dev <1.8.0-pre",
-    "@keep-network/sortition-pools": ">1.2.0-dev <1.2.0-pre",
+    "@keep-network/sortition-pools": "^1.2.0-dev.1",
     "@openzeppelin/upgrades": "^2.7.2",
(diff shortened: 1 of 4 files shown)


solidity/package.json
```diff
@@ -29,7 +29,7 @@
   "homepage": "https://github.com/keep-network/keep-ecdsa",
   "dependencies": {
     "@keep-network/keep-core": ">1.8.0-dev <1.8.0-pre",
-    "@keep-network/sortition-pools": ">1.2.0-dev <1.2.0-pre",
+    "@keep-network/sortition-pools": "^1.2.0-dev.1",
```
nkuba, line 32:
```suggestion
    "@keep-network/sortition-pools": "1.2.0-dev.1",
```

nkuba (changes requested):

solidity/package.json
michalinacienciala: Right, see 76fb64d.

nkuba (approved):

PR #949, opened 2022-01-07
Verify `LPRewardsTBTCv2Saddle` on Etherscan

michalinacienciala:
As part of deployment process of `keep-ecdsa` contracts via execution of
`.github/workflows/contracts.yml` workflow we run a script which
verifies deployed contracts on Etherscan
(`solidity/scripts/etherscan-verify.sh`).
We recently added new contracts (`LPRewardsTBTCv2Saddle`,
`LPRewardsTBTCv2SaddleV2`) to the set of `keep-ecdsa` contracts that can
be deployed and we're now adding it to the `etherscan-verify.sh` script.


r-czajkowski (approved):
LGTM! :shipit:

